PR #43, opened 2019-11-02
Added keyboard binding for 0-9, * and # buttons on the controller 

Testsr:
- key is the * button and = key is the # button


Testsr:
Also changed Jaguar Model M bios to be the same type as Jaguar Model K.

hizzlekizzle:
This seems like a good improvement to me.

PR #53, opened 2021-09-14
Possible speedups

JoeMatt:
I'm making a PR for some old optimizations I made for an iOS fork.

This PR is more for the devs to test and cherry-pick, I don't expect this code to be "merge quality", but certainly buildable/testable at least.


twinaphex:
Both Wolfenstein 3D and Doom don't start up for me with this PR on Windows 10 (64bit build).

twinaphex:
I don't understand why audio_batch_cb is commented out too. This means there will be no sound at all in the libretro core.

twinaphex:
That being said, there are some good ideas in here.

JoeMatt:
> That being said, there are some good ideas in here. 

Thanks!

This was being used in Provenance and we have our own video and audio pipelines.

This PR was never intended for mass consumption but rather for me to get it to work well in Prov and back port to upstream.

But as you know, there really isn't a true upstream (my original work wasn't from retroarch's fork but the original .Zip source).

So any things that don't make sense are due to that history. I later back ported to your fork but a lot of hacks and such for my specific purposes still exist, which is why i wanted to state that in the original description.

I felt bad keeping all these changes to myself after seeing other wanting this core to make progress so as much as I myself hate when people do a dump of code "here's my amazing but difficult to read changes, bye", at the same time, I have very little time to focus on this as much as I would love to so I figured it was better to submit in the present state and hopefully someone else more talented and with the time can take some value from this work.

JoeMatt:
> Both Wolfenstein 3D and Doom don't start up for me with this PR on Windows 10 (64bit build).

Some ROMs work better with internal or provided BIOSs.

I think my hacks on the blitter broke the ID Roms though because they were coded better than most and used some hardware hacks for performance so I'm not suprised those are broken. I bet a little debugging and I or someone else could figure out what's wrong there.

This core could use some unit tests to figure out what about the cpu emu is different with my changes but that's probably a stretch at this point.

src/dsp.c
```diff
@@ -2031,7 +2061,7 @@ static void DSP_jr(void)
       }//*/
       dsp_pc += 2;	// For DSP_DIS_* accuracy
       DSPOpcode[pipeline[plPtrExec].opcode]();
-      dsp_opcode_use[pipeline[plPtrExec].opcode]++;
+//      dsp_opcode_use[pipeline[plPtrExec].opcode]++;
```
JoeMatt, line 2064: Maybe this is a cause of breaking @twinaphex 
i'll re-add and repush

twinaphex:
Alright, thanks.

JoeMatt:
> Alright, thanks.

also i'm using pre-processor macros in XCode,
-DINLINE=inline -DUSE_STRUCTS=1

JoeMatt:
> I don't understand why audio_batch_cb is commented out too. This means there will be no sound at all in the libretro core.
> 
> if this is something you ported from Provenance and did not test in RetroArch with the libretro core, that isn't going to do us much good. We'd need a backport of the changes you made then to the core and to ensure things still work.

Pushed update to un-delete this.

JoeMatt:
Re-based and force pushed off of upstream b44e2e29128ecb4d97a23f1e7a71fa2327e17ca3

JoeMatt:
> Both Wolfenstein 3D and Doom don't start up for me with this PR on Windows 10 (64bit build).

FYI, libretro wiki says Wolf 3D already doesn't boot,
https://docs.libretro.com/library/compatibility/jaguar/

Doom needs a hack which I don't have a menu UI for in Prov, not sure if you tested with that in retroarch.

PR #55, opened 2021-10-08
[WIP] Speedup rebase rebase

JoeMatt:
Rebase of #53 as per @twinaphex requests.

There are diff's around controller input fixes that I reverted from #53 that should be looked into in a new PR.

specifically these commits

fb695b1b0e1f104136428834b2b609c2139c7f8c
f4ebb99aa85369f774d41ba764a17a3e45a6889d
681a1f3ab40c60833e886f0f461b3d47d1ee4524
34ca42f3339dc83d1fcf68ed21802db2cd4c1df5

## Original PR
---

I'm making a PR for some old optimizations I made for an iOS fork.

This PR is more for the devs to test and cherry-pick, I don't expect this code to be "merge quality", but certainly buildable/testable at least.

For instance, probably don't want these hacks that were specific to iOS loading,
https://github.com/libretro/virtualjaguar-libretro/pull/53/commits/811e73f8b574ec2b34a69934fa36f66c3202f45b

The blitter really is the slowest part. I wanted to write something with SIMD or better vectorization or improved memcpy or something. That was by far the slowest part when I profiled this pretty heavily a while ago (at least it was excruciatingly slow on ARM without my C struct hacks to help the compiler optimize better.

---
Update:
Forgot to mention I have these macros defined in my build (in XCode) that need to be added to a header or C compile flags

C
```c
#define INLINE=inline 
#define USE_STRUCTS=1
```

flags
```sh
-DINLINE=inline -DUSE_STRUCTS=1 -D__GCCUNIX__  
```

Inline macro may need to be customized based on arch. Mac OS is using Clang/C99/gnu++11 syntax for inline.

I also have `unroll loops` aka `-qunroll=no` but i forget why.

PR #56, opened 2021-10-13
Blitter.c Unroll loop

JoeMatt:
Pulled from #55 
Minor performance optimization by unrolling a small for loop that's called often in `blitter.c`


JoeMatt:
Tested, runs NBA Jam

PR #63, opened 2021-10-20
Fix 39 & 54 numpad controller/keyboard mapping

### LINKED ISSUE #39: [bug] Numpad 7, 8, 9, *, # are unmappable onto the joypads.
Self expantory

### LINKED ISSUE #54: [bug] Numpad 7, 8, 9, *, # input impossible for player 2 and above in RetroArch or P1 controllers.
This is related to this issue, but that one is specifically about joypad mapping:
https://github.com/libretro/virtualjaguar-libretro/issues/39

There's a workaround of using the keyboard number keys for player 1. But there is no such option for players 2 and above.

JoeMatt:


Testsr (changes requested):
There is a loop above this code block, that resets all the buttons. 
I would suggest removing all the code resetting them explicitly.

PR #77, opened 2022-11-17
Fix resolution changes in games

flytrapdev:
Jaguar games can change resolutions during gameplay.

Previously, the emulator only ran at the default 320 x 240 resolution.
This PR fixes this, allowing other resolutions to display properly.

This should fix issues in (at least) the following games, based on my testing :

- Atari Karts
- Club Drive
- Evolution Dino Dudes
- Flashback
- I-War
- Pinball Fantasies
- Trevor McFur in the Crescent Galaxy


JoeMatt:
Merged this into Provenance branch too. Will be in my next release.

PR #115, opened 2026-04-23
Fix CI buildbot failures: MSVC C89, SSE2 32-bit, webOS cross-compile

JoeMatt:
## Summary

Fixes all CI failures from [pipeline #64992](https://git.libretro.com/libretro/virtualjaguar-libretro/-/pipelines/64992):

- **MSVC 2005/2010 (windows-msvc05-i686, windows-msvc10-x64, windows-msvc10-i686)**: `error C2143: syntax error : missing ';' before 'type'` — `extern` declarations were placed mid-block in `libretro.c` save state functions. MSVC's C89 mode requires all declarations at the top of a block. Moved them to function scope.

- **Linux i686 + Windows i686 MinGW**: `undefined reference to '_mm_cvtsi128_si64'` — this SSE2 intrinsic only exists on x86_64 (needs 64-bit general-purpose registers). Replaced with a `_mm_storel_epi64`-based helper that works on both 32-bit and 64-bit x86. Also fixed C89 mid-block `__m128i` declarations in `sse2_zcomp` and `sse2_byte_merge`.

- **webOS ARM (libretro-build-webos-armv7a)**: `fatal error: emmintrin.h: No such file or directory` — the webOS CI template sources an ARM cross-compiler SDK but doesn't set `platform`, so `Makefile.common`'s native `uname -m` fallback detected the x86_64 host and selected SSE2. Added cross-compiler prefix detection (`arm-`, `aarch64-`, `mips`, `powerpc` in `CC`) to skip the host-arch fallback when cross-compiling.

## Test plan

- [x] Local native build passes (macOS arm64, NEON)
- [x] `CC=arm-webos-linux-gnueabi-gcc` → scalar (cross-compile detected)
- [x] `CC=aarch64-linux-gnu-gcc platform=unix` → scalar (cross-compile detected)
- [ ] CI: MSVC 2005/2010 builds
- [ ] CI: Linux i686 build
- [ ] CI: Windows i686 MinGW build
- [ ] CI: webOS ARM build

### DIFF
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -5,6 +5,7 @@ on:
     branches: [ master ]
   pull_request:
     branches: [ master ]
+  workflow_dispatch:
 
 jobs:
   build:
@@ -31,6 +32,13 @@ jobs:
             cc: 'gcc'
             cxx: 'g++'
 
+          - displayTargetName: 'Linux i686'
+            artifact: 'virtualjaguar_libretro.so'
+            os: ubuntu-latest
+            cc: 'gcc -m32'
+            cxx: 'g++ -m32'
+            multilib: true
+
           # macOS
           - displayTargetName: 'macOS arm64 (Clang)'
             artifact: 'virtualjaguar_libretro.dylib'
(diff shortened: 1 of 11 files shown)


JoeMatt:
Addressed both Copilot review comments in a1bffbf:

1. **Unused `#include <string.h>`** — Was removed in the previous push, but re-added now because we switched to `memcpy` for the extract helper (see below).

2. **Alignment UB with `_mm_storel_epi64` cast** — Good catch on the `__m128i*` alignment concern. Replaced with `memcpy(&r, &v, sizeof(r))` which is alignment-safe and compilers optimize to a single register move at `-O2`. This also keeps the `#include <string.h>` justified.

PR #166, opened 2026-05-05
fix(blitter): skip dstd memory read when bkgwren is set (AvP alpha noise)

JoeMatt:
Fixes the red alpha noise on AvP map screen in accurate blitter mode.

## Summary
- Root cause: Midsummer blitter dwrite state unconditionally read destination memory in phrase mode, injecting stale framebuffer pixels into inhibited byte positions via byte_merge
- When bkgwren is set, hardware uses the DSTDATA register value (typically zero) as background, NOT live framebuffer data
- Fix: add && !bkgwren to the phrase-mode dstd read condition (matching the collapsed pattern-fill path that already had this guard)

## Test plan
- [x] AvP: map screen in accurate blitter mode shows clean pixels
- [ ] Battle Sphere: cockpit HUD reticle still renders correctly
- [x] test/test_blitter_mmio passes
- [x] Build passes on GCC + Clang

Generated with Claude Code

### DIFF
--- a/src/tom/blitter.c
+++ b/src/tom/blitter.c
@@ -2977,17 +2977,13 @@ A2ptrldi	:= NAN2 (a2ptrldi, a2update\, a2pldt);*/
                //byte_merge must see the existing dest byte in the low 8 bits of
                //dstd or the unmodified pixel slots in that byte get zeroed
                //(matches WRITE_PIXEL_1/2/4 RMW in the fast blitter).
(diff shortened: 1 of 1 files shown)


JoeMatt:
Copilot review addressed in c9ace2a — updated comment to match the actual logic. The phrase-mode path reads framebuffer regardless of dsten (matching the collapsed path at line 2117), so the !DSTEN reference was misleading.

PR #167, opened 2026-05-05
fix(dsp): spoof DSPGO=1 in HLE mode for Skyhammer/IS2 audio

JoeMatt:
Fixes Skyhammer and Iron Soldier 2 producing loud full-scale clipped audio in HLE (no-BIOS) mode.

## Summary
- Root cause: games read D_CTRL at boot to check if BIOS audio engine is running (DSPGO=1). In HLE mode DSPGO=0, so they load a broken alternate DSP program that produces saturated square waves
- Fix: phantom DSPGO flag set during DSPReset() when no BIOS. 68K reads of D_CTRL return dsp_control | DSPGO until the game writes to D_CTRL itself
- The DSP does NOT execute during this period — purely a readback spoof
- Safe for games with self-contained audio (Doom, Atari Karts) — they write D_CTRL early, clearing phantom immediately
- State load unconditionally clears the flag (boot-time mechanism only)

## Test plan
- [ ] Skyhammer: audio should be clean, not saturated/clipped
- [ ] Iron Soldier 2: no loud square wave output
- [ ] Doom/Atari Karts/Tempest 2000: audio unaffected
- [ ] Build passes on GCC + Clang

Generated with Claude Code

### DIFF
--- a/src/jerry/dsp.c
+++ b/src/jerry/dsp.c
@@ -298,6 +298,23 @@ static uint8_t dsp_ram_8[0x2000];
 
 static uint32_t dspgo_poll_count;
 
+/* HLE "phantom DSPGO" flag.
+ *
+ * In HLE (no-BIOS) mode, the real BIOS normally loads and starts a DSP audio
+ * engine before handing control to the cartridge.  Games like Skyhammer and
+ * Iron Soldier 2 read D_CTRL to check whether DSPGO is set; if they see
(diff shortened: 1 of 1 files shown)


JoeMatt:
Tested in RetroArch: doesn't fix Skyhammer, Wolf3D, or Doom audio. Superseded by the DSP engine pre-load approach in #170 (which also needs rework). Closing.

PR #168, opened 2026-05-06
fix(gpu): simulate bus contention for GPU external memory accesses

JoeMatt:
## Summary

- Add 60-cycle penalty to all GPU LOAD/STORE opcodes targeting external memory (outside local RAM 0xF03000-0xF03FFF), simulating bus arbitration against the 68K, Object Processor, blitter, and DRAM refresh
- Scale 68K cycle budget by 0.68x when GPU is actively running
- Expose as opt-in core option "Bus Contention (GPU Slowdown)" (`virtualjaguar_68k_contention`, disabled by default)

**Root cause:** On real hardware the GPU is priority 9 of 11 on the system bus. Every external memory access must wait for bus grant. Without this, GPU rendering completes in 1-2 VBlanks instead of the 4-8 real hardware requires, making Doom run at 30fps instead of its correct 10-15fps.

**Calibration:** Penalty value (60 cycles) was tuned against Doom's source code timing model (`TICRATE=15`, adaptive `vblsinframe=4-8`), MiSTer FPGA measurements showing VJ gives the GPU ~8.5x too many effective cycles, and JTRM hardware documentation for DRAM/bus timing. The flat penalty compensates for several unmodeled subsystems (OP bus stealing, per-cycle arbitration, instruction prefetch stalls).

| Penalty | Doom FPS | VBlanks/frame |
|---------|----------|---------------|
| 0 (off) | 30.0 | 2 |
| 20 | 27.4 | 2-3 |
| 35 | 22.4 | 2-5 |
| 50 | 16.9 | 2-6 |
| **60** | **13.6** | **3-6** |
| 80 | 11.1 | 3-8 |
| Real HW | 10-15 | 4-8 |

**Regression testing** (all boot and render correctly with contention enabled):
- Alien vs Predator, Atari Karts, Iron Soldier, Super Burnout, Wolfenstein 3D
- Full unit test suite passes, C89 lint clean

## Test plan

- [ ] Enable "Bus Contention (GPU Slowdown)" in core options
- [ ] Verify Doom gameplay runs at ~12-15fps (noticeably slower than 30fps)
- [ ] Verify AvP, Atari Karts, Iron Soldier boot and play correctly
- [ ] Verify games that don't use GPU rendering (blitter-based) are unaffected
- [ ] Verify option defaults to disabled — no change for users who don't enable it

### DIFF
--- a/exports-test.list
+++ b/exports-test.list
@@ -42,3 +42,4 @@ _perf_counters_dump
 _perf_counters_reset
 _perf_counters_register
 _perf_counters_find
+_GetRamPtr
--- a/libretro.c
+++ b/libretro.c
@@ -320,6 +320,17 @@ static void check_variables(void)
          vjs.hardwareTypeNTSC = true;
    }
 
+   var.key = "virtualjaguar_68k_contention";
+   var.value = NULL;
+
+   if (environ_cb(RETRO_ENVIRONMENT_GET_VARIABLE, &var) && var.value)
+   {
+      if (strcmp(var.value, "enabled") == 0)
+         vjs.use68KContention = true;
+      else
+         vjs.use68KContention = false;
+   }
+
    var.key = "virtualjaguar_alt_inputs";
    var.value = NULL;
    if (environ_cb(RETRO_ENVIRONMENT_GET_VARIABLE, &var) && var.value)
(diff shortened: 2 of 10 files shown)


JoeMatt:
Closing — the flat per-opcode penalty approach is too blunt. It fixes Doom's speed but breaks demo boot and makes other games too slow. Will rework with a proper centralized bus arbitration model based on the MiSTer FPGA core's arbiter architecture.

PR #185, opened 2026-05-10
Add BIOS file info in the description

checktext00:
## Summary

Added info about the BIOS file being provided by the core


JoeMatt:
this verbage is wrong, the bios file isn't provided by the core.
An HLE alternative is, but this setting is to explicitly not use that, therefor the user needs to supply the assocated BIOS.

PR #217, opened 2026-07-30
fix(test): gate the BIOS_INTRO band on the boot-stub marker — the FMV "handoff gap" was a classifier defect

JoeMatt:
**There was no handoff gap.** BrainDead 13, Dragon's Lair and Primal Rage hand off and play in real-BIOS mode. The `BIOS_INTRO` stage label was a defect in `cd_boot_matrix.sh`'s classifier, not an emulation defect — which retires item 1 of `docs/cd-known-issues.md` entirely.

## The one-line proof (checkable against the committed table)

Dragon's Lair's **hle and bios rows both report `final_pc=$004C12`**, and BrainDead 13's both report `$004FCA` — the same instruction in the same code — yet only the bios row was labelled `BIOS_INTRO`, because only that path applied the band rule.

And the control pair the whole candidate ranking rested on was never real: Space Ace vs Dragon's Lair wasn't "works vs stuck" — Space Ace's copy of the *same* ReadySoft player is simply linked at `$00B87A`, outside the band.

## Per-title diagnosis (disassembled from live RAM dumps)

| Title | PC | Polls | Writer |
|---|---|---|---|
| Dragon's Lair | `$004C0A` | `$562E` | GPU PIT ISR at `$F0308A`, **30.0 Hz** |
| BrainDead 13 | `$004FCA` | `$72DA` | same ISR, **24.0 Hz** |
| Primal Rage | `$004436`–`$0044EE` | — joypad scan, not a loop | — |

DL and BD13 are both ReadySoft and share one FMV player linked at different addresses — the instruction streams at `$004BF0` and `$004FB0` are byte-for-byte identical. The loop waits for the movie's presentation clock to reach the next frame's timestamp, and **that clock was measured still advancing**: a 48-bit fixed-point accumulator sampled across frames 1200/1800/2400/2900 gives exactly 30.0/s for DL and 24.0/s for BD13. Sampling had simply landed mid-wait for the next tick. Primal Rage was never in a loop at all.

## The defect

`classify_stage()` mapped all of `$004000-$007FFF` to `BIOS_INTRO`. That band is **shared**: the real BIOS does execute there pre-handoff (the auth path is hooked at `$005E40`), but games link code there too — Dragon's Lair's boot exe loads at `$004000`, Myst's at `$005000`, and BD13/Primal Rage load a second-stage player there. The rule's only cited provenance was a single ISO row, and ISO loads are now refused outright as unbootable — so it rested on a known-broken run.

Fix: gate the band on the `[CD-BOOTSTUB] Injected` marker the core logs when the BIOS reaches handoff. Marker absent → `BIOS_INTRO`; present → falls through to `GAME_CODE`. **No `src/` change** — 45 lines in the test script plus docs.

## Gates

1. All three now `GAME_CODE`, with screenshots read: DL plays its attract FMV (79.5% non-black, RMS 1230); BD13 its castle-and-moon intro (68.1%, RMS 2651); Primal Rage its full attract cycle — title → DEMO gameplay with two dinosaurs fighting → back to title (93.3%, RMS 980). PR needed 6000 frames; its zero-motion windows are black scene transitions, the benign pattern already documented for Myst.
2. **Matrix 20/20: three forward, zero backward.** Every at-risk bios row still `GAME_CODE 1/1` (Baldies, Battle Morph, Highlander, Hover Strike, IS2, Space Ace). **All 20 rows are now `GAME_CODE 1/1` in both boot modes.**
3. Wedge probes clean: Battle Morph hle+bios, IS2 bios, Myst hle+bios, Baldies bios. Myst RMS 2548 (bios) / 2698 (hle), in envelope.
4. `make TEST_EXPORTS=1 test` exit 0, both audio tests green.

## Follow-ups filed in the docs

- **Myst is a fourth title in the same band** that the matrix never exercised — `CUE_TITLES` is a hardcoded list omitting it even though the harness discovers it. The band fix already covers Myst; the list should be fixed separately.
- **Build-stamp convergence is structurally impossible** in a chunked sweep: the stamp is computed once per invocation and `CD_MATRIX_OUT` is tracked, so the first row written dirties the tree. `17375c3-dirty` is the honest build id for the code under test.
- The `marker absent → BIOS_INTRO` branch has no live coverage now that all 20 titles reach handoff; it was verified with a throwaway unit check over both branches plus four control bands.

### DIFF
--- a/docs/cd-bizhawk-comparison.md
+++ b/docs/cd-bizhawk-comparison.md
@@ -12,6 +12,34 @@ legitimate).
 
 Nothing in `src/` was changed for this note.
 
+> **RESOLVED 2026-07-29 -- there was no handoff gap.** All three titles do
+> hand off and play in bios mode. The `BIOS_INTRO` labels came from
+> `cd_boot_matrix.sh`'s stage classifier treating all of
+> `$004000`-`$007FFF` as BIOS code; games link code there too. **None of
+> the three ranked candidates in §3 applies** -- do not sweep them for
+> this. What settled it is §3's own prescribed first step ("Do this
+> first"): the per-PC disassembly. Dragon's Lair and BrainDead 13 were
+> sitting in a shared ReadySoft FMV player's wait-for-next-frame loop on a
+> 30 Hz / 24 Hz presentation clock that was **measured still advancing**;
+> Primal Rage was in its joypad scanner. Diagnosis, clock measurements and
+> screenshots: 2026-07-29 re-run notes in `docs/cd-boot-matrix.md`.
+> The §2.3 control pair dissolves too: Space Ace bios (`$00B87A`) vs
+> Dragon's Lair bios (`$004C12`) was "sampled outside the band vs inside
+> it", not "works vs stuck" -- Space Ace runs the same player, linked
+> higher. §4.1 and §4.2 remain genuine bugs on their own merits, but
+> **candidate 2 is separately falsified for these titles**: BrainDead 13
+> and Dragon's Lair emit only `$150A` and never leave double speed, and
+> Primal Rage's `$1501` comes only after its boot read finishes.
+>
+> Correction to this note's §2.2/§4.2 reading of the wire format: the
+> `$15nn` payload is **not** a bit field. The retail BIOS at `$808978`
+> builds a **one-based** speed code -- `and.w #$1,d2 / add.w #$1,d2`
+> (1 = single, 2 = double) -- then `bset #3,d2` for data mode. So `$150A`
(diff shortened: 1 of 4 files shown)


JoeMatt:
Absorbed by #228, which carried the same `$4000-$7FFF` stage-band gate (commit `5a83d3d`, identical change) plus the Myst matrix row and the baldies.cdi correction. #228 merged as `b300b36`, so this branch's code change is already on `develop` — closing rather than merging a duplicate.

The two review comments here were not lost: the stale "Target question" premise in `docs/cd-bizhawk-comparison.md` and the machine-specific log path in `docs/cd-boot-matrix.md` are being fixed in a follow-up against develop.

PR #250, opened 2026-08-01
perf(net): bounded wall-clock wait for netlink replies (Wi-Fi lag fix)

JoeMatt:
## Root cause
Laggy Wi-Fi link play despite <16 ms ping (reported after #248) is **receive-side frame quantization**. `retro_run` executes 16.7 ms of emulated time in ~1-2 ms of wall clock, so a game spinning on ASISTAT for its partner's reply exhausts the frame's emulated budget long before a real network can answer. The reply always slips to a later `retro_run`, and every lockstep tic exchange rounds up to whole video frames regardless of actual RTT. Localhost (~0.1 ms) lands inside the same frame — which is why it always felt snappy while Wi-Fi lagged. TX-side flushing (#244/#248) could never fix this: it's the receive side that has to wait.

**Measured** with the new 68K ping-pong probe (real ASISTAT-spin code on both sides) through a 3 ms/way delay proxy: **445 exchanges/s direct collapses to ~40/s** (one per 1-2 frames).

## Fix
While a TX burst is unanswered on a real transport, the ASISTAT poll path blocks in wall-clock time, pumping the transport (TCP socket or RA netplay `poll_receive`, which does read the wire) until the reply lands in the **same** frame.

The wait budget is **self-tuning** — no user tuning. Reply latency is sampled at the transport boundary (last TX byte → first byte back) into an EWMA; each frame may wait up to 2× that plus margin, clamped to [4, 15] ms. Localhost converges to the floor (sub-ms cost); real networks converge to what they measure. Sampling is decoupled from the wait-arm flag so a timed-out wait still measures the late reply (otherwise the budget could never grow). This mirrors how melonDS handles local-wireless sync — the emulator stalls for the peer automatically; RA netplay's own latency features (input sync/rollback/frame delay) don't apply to netpacket cores.

Core option: **Network Link Latency Hiding** — enabled (default) / disabled. Guard rails: disarms after one full-budget timeout so idle ASISTAT polling never stalls repeatedly; bails on disconnect; budget refills once per `retro_run`; wall-clock only — no emulated state or savestate impact; no-op stub on socketless console targets; `QueryPerformanceCounter` on Windows. `VJ_NETLINK_WAIT_DEBUG=1` logs the adaptive state once a second.

## Results (through the delay proxy, 6 ms RTT, free-running echo)
| case | exchanges/s |
|---|---|
| disabled (old behavior) | ~47 |
| enabled (adaptive) | **93–95** |

Direct localhost: 445 → **1823** exchanges/s (the wait also beats the /16 pump gate).

## Testing
- New `netlink_delay_proxy` (latency injection) + `netlink_latency` (68K-driven exchange-rate probe) + `netlink_latency_test.sh` wired into `make test`, asserting both the quantization control (disabled stays collapsed) and the adaptive rate — stable across runs.
- Full `make TEST_EXPORTS=1 test`: exit 0. `c89-lint` clean.
- Real-network validation pending: iOS ↔ macOS Doom deathmatch over Wi-Fi (reporter's setup).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -95,3 +95,5 @@ test/tools/__pycache__/
 *.gcda
 *.gcno
 coverage.xml
+test/tools/netlink_latency
+test/tools/netlink_delay_proxy
--- a/Makefile
+++ b/Makefile
@@ -808,7 +808,7 @@ test: test/test_cheat test/test_event_queue test/test_jlink test/test_jlink_tcp 
 		test/test_cd_boot test/test_cd_hle_boot test/test_cd_bios_boot test/test_cd_toc_contract test/test_cd_fifo_stream test/test_cd_ssi_stream test/test_cd_second_transfer test/test_cd_hle_idempotent test/test_cd_lost_wakeup test/test_cd_pregap \
 		test/test_audio_dac test/test_blitter \
 		test/tools/test_memory_map test/tools/test_op_gpu_object test/test_uart_core \
-		test/tools/netlink_pair
+		test/tools/netlink_pair test/tools/netlink_latency test/tools/netlink_delay_proxy
 	./test/test_cheat
 	./test/test_event_queue
 	./test/test_jlink
@@ -817,6 +817,7 @@ test: test/test_cheat test/test_event_queue test/test_jlink test/test_jlink_tcp 
 	./test/test_uart_loopback
 	./test/test_uart_core ./$(TARGET)
 	bash test/tools/netlink_pair_test.sh ./$(TARGET)
+	bash test/tools/netlink_latency_test.sh ./$(TARGET)
 	./test/test_blitter_mmio
 	./test/test_blitter_cmd ./$(TARGET)
 	./test/test_pit_clock_rate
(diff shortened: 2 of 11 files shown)


JoeMatt:
Reworked per design feedback: the millisecond enum is gone. The wait budget is now **self-tuning** — reply latency is sampled at the transport boundary into an EWMA and the per-frame budget is 2× that (clamped 4–15 ms). Localhost converges to the floor (sub-ms cost), Wi-Fi converges to what it measures. The option is now just **Network Link Latency Hiding: enabled (default) / disabled**.

This matches how comparable linked-console emulators handle it (melonDS local-wireless sync stalls for the peer automatically, no knob); RA netplay's automatic latency machinery (input sync/rollback/frame delay) never applies to netpacket cores, so the core owns this.

Updated numbers through the 3 ms/way delay proxy (free-running echo, stable across runs): **47/s disabled vs 93–95/s enabled**. `VJ_NETLINK_WAIT_DEBUG=1` logs the adaptive state for diagnostics. Full suite green.

Issue #243, opened 2026-08-01
CD/HLE: Vid Grid (USA) (Rev 1) livelocks re-issuing the same CD_read; BIOS mode boots fine


JoeMatt:
## Summary

`Vid Grid (USA) (Rev 1)` never gets past its first load in **HLE mode**: it issues one `CD_read`, our shim completes it, and the game immediately re-issues the *identical* read — same `D0`/`D1`/MSF/LBA/dest/size — forever. The 68K sits on a single PC (`$0041B4`, `unique_pcs=1`) for the whole run.

**BIOS mode boots the same disc fine** (`final_pc=$01BF9C`, `unique_pcs=68`), so this is our HLE CD shim, not the dump and not the title.

This is *not* a CDI problem — it reproduces identically from CUE/BIN. It was previously masked: the Rev 1 CDI failed to load outright for an unrelated reason (bad rip, see #230), so this never surfaced.

## Reproduction

```bash
DYLD_LIBRARY_PATH=. \
  VJ_TEST_CD_ROOT="test/roms/private/Jaguar CD/BinCue" \
  VJ_TEST_CD_FOCUS="Vid Grid (USA) (Rev 1)" \
  VJ_TEST_CD_EXTS="cue" VJ_TEST_CD_FRAMES=900 \
  ./test/test_cd_hle_boot
```

```
[FAIL]  Vid Grid (USA) (Rev 1).cue : pc_in_ram=1 not_loop=0 not_thrash=0 ram_payload=6352B unique_pcs=1 final_pc=$0041B4
[DIAG]  gpu_pc=$F03068 irq0=0 irq3=0 butchExec=471600 fifoIRQs=0 dsaIRQs=0 fifoReads=0 seeks=0
```

Swap `test_cd_hle_boot` for `test_cd_bios_boot` and it passes.

## Evidence: the same read, over and over

The boot stub extracts and injects cleanly — the header magic is found and verified — then this block repeats verbatim until the frame budget runs out:

```
[CD-HLE] CD_read: D0=$00023924 D1=$41545249 ('ATRI') MSF=02:57:36 LBA=13161 dest=$00FFE0 end=$1113E0 size=$101400
[CD-HLE] CD_read: sync block (8+ matches) ends at LBA 13168 offset 2350 (scanned 8 sectors from seek base 13161)
[CD-HLE] CD_read: streaming 1053696 bytes to $00FFE0-$1113DF from LBA 13168 offset 2350
[...]

PR #256, opened 2026-08-02
fix(cd): walk HLE sync runs across sector boundaries (fixes #243)

JoeMatt:
Fixes #243.

## Root cause

`Vid Grid (USA) (Rev 1)` livelocked in HLE mode, re-issuing a byte-identical `CD_read` forever. The transfer was correct — it just **started 32 bytes too early**.

`CD_read` finds the sentinel sync block, walks forward to the first non-sentinel byte, and streams from there. That walk was per-sector and only continued into the next sector when the run ended *exactly* on a boundary:

```c
while (scanOff >= 2352) { ... }
```

When the run start is not longword-aligned, the last offset a sector can test is 2348..2351 — so the inner match loop exits with `scanOff` **below** 2352 while the run continues in the next sector. The guard never fired and the walk stopped mid-run.

Rev 1 hits it precisely: the walk stopped at LBA 13168 offset 2350 with 8 sentinel longwords still to come, prepending 32 stray `ATRI` bytes:

```
$00FFE0: ATRIATRIATRIATRI ATRIATRIATRIATRI   <- 32 stray bytes
$010000: ATARI APPROVED DATA HEADER ATRI!    <- should have started at dest
```

Its loader validates the payload with a 12-byte compare at `$004208`:

```
$004208:  LEA     $0000FFF4,A0      ; dest+20
$00420E:  LEA     $00004D68,A1      ; 'HEADER ATRI!'
$004214:  MOVEQ   #11,D0
$004216:  CMPM.B  (A0)+,(A1)+
$004218:  BNE.S   $004220           ; -> ADDI.L #1,$4D64 ; BRA $004136 (reload)
$00421A:  DBRA    D0,$004216
```

With the payload shifted, the tag sat at `dest+52`, the compare missed, and it branched back to reload. The failure counter `$4D64` reached 5 (one per completed-then-rejected transfer) while the poll-loop timeout counters `$4D58`/`$4D5C` stayed 0 — the completion handshake was always fine.

The other two Vid Grid revisions ship a larger loader (`$1BD0` vs `$D88`) without that check, which is why only Rev 1 surfaced it.

## Fix

Walk the run as a continuous byte stream: normalize whole sectors as they're crossed, and fetch each candidate longword *across* the boundary when fewer than 4 bytes remain in the current sector.

## Testing

35-disc CUE/BIN corpus:

| | before | after |
|---|---|---|
| HLE | 21 pass / 14 fail | **22 pass / 13 fail** |
| BIOS | 35 pass / 0 fail | 35 pass / 0 fail |

The HLE pass/fail set diff is **exactly one line** — `Vid Grid (USA) (Rev 1)` `FAIL` → `PASS`. No other disc changed state. BIOS mode is untouched (it doesn't use this scan).

- [x] `make TEST_EXPORTS=1 test` — exit 0
- [x] `scripts/c89-lint.sh` clean
- [x] Rev 1 reaches `final_pc=$01BFF4` / `unique_pcs=56`, matching BIOS mode's `$01BF9C` / 68, and proceeds into its game-data loads

### DIFF
--- a/src/cd/jagcd_hle.c
+++ b/src/cd/jagcd_hle.c
@@ -813,32 +813,100 @@ static void HLEHandleCDRead(void)
                continue;  /* stray match — keep searching for a real sync block */
             }
 
-            /* Sync block confirmed.  Scan forward across sector boundaries
-             * to find where the sentinel pattern ends. */
-            scanLBA = scan_base + s;
-            scanOff = j;  /* first non-sentinel byte in current sector */
-
-            /* If the sync block extends to the end of this sector, keep
-             * scanning subsequent sectors. */
-            while (scanOff >= 2352)
+            /* Sync block confirmed.  Walk the run to its end, treating
+             * the disc as a continuous byte stream.
+             *
+             * A sync run can straddle sector boundaries, and when its
+             * start is not longword-aligned the last offset a sector can
+             * test is 2348..2351 — the walk then stops with scanOff
+             * BELOW 2352 while the run continues in the next sector.
+             * The old `while (scanOff >= 2352)` continuation guard only
+             * fired on an exact boundary landing, so those runs ended
+             * early: Vid Grid (USA) (Rev 1) stopped at offset 2350 with
+             * 8 sentinel longwords still to come, prepending 32 stray
+             * 'ATRI' bytes to the payload.  Its loader compares
+             * 'HEADER ATRI!' at dest+20 (`cmpm.b` x12 at $004208) and
+             * re-issues the entire read forever when that misses. */
             {
-               scanLBA++;
-               scanOff = 0;
-               if (!CDIntfReadBlock(scanLBA, sectorBuf))
-                  break;
-               for (i = 0; i + 1 < 2352; i += 2)
+               uint8_t  nextBuf[2352];
+               bool     haveNext = false;
+               bool     readOK   = true;
+
+               scanLBA = scan_base + s;
+               scanOff = j;
+
+               while (readOK)
                {
-                  uint8_t tmp2 = sectorBuf[i];
-                  sectorBuf[i]     = sectorBuf[i + 1];
-                  sectorBuf[i + 1] = tmp2;
-               }
-               /* Advance past continuing sentinel matches */
-               while (scanOff + 3 < 2352 &&
-                      sectorBuf[scanOff]   == pat[0] && sectorBuf[scanOff+1] == pat[1] &&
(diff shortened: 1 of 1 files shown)


JoeMatt:
## Note: the `video_stall` this title logs is benign — checked

Vid Grid (Rev 1) logs `video_stall` twice in a long HLE run, so recording why that isn't a regression from this PR.

Traced with `cd_visual_verify` over 3000 frames (per-second motion timeline + periodic screenshots):

| frames | behaviour |
|---|---|
| 0–599 | mostly black — the CD load. **First `video_stall` (frame 572) lands here.** |
| 600–1139 | logos animating, 24/60 frames with motion |
| 1200–2159 | **FMV playing** — 55.2% non-black, 12–15% inter-frame change |
| 2220+ | title / copyright screen, static by design |
| 2760+ | **Second `video_stall`** — sitting on that static screen |

Screenshots confirm real content, not a frozen buffer: frame 1350 is a decoded FMV frame (pylon against a night sky) and frame 2850 is the fully-rendered Vid Grid title screen with the Geffen / Jaguar / Jasmine Multimedia logos and complete copyright text.

The decisive check was scripting input. Pressing A at frame 2600 changes the screen (non-black 86.4% → 88.5%) and brings up a modal dialog:

```
MEMORY TRACK NOT PRESENT
GAME INFORMATION WILL NOT BE SAVED.
```

So the 68K is running, the OP is compositing, and input is being processed — the title screen is simply static and waiting for a button, which nothing presses headlessly. Audio corroborates: 1,817,156 of 2,400,000 samples non-silent, onset at frame 449, only 2 dropouts.

This is the same known-benign class already documented in CLAUDE.md for Myst (a title that goes CD-idle after finishing a transfer trips the same watchdog). No code change here.

Incidentally: the Memory Track dialog is the Jaguar CD save-cartridge check. We don't emulate Memory Track, so CD titles can't save — a real gap, but pre-existing and unrelated to this PR.

JoeMatt:
Correction to the Memory Track aside above: **it is emulated** — `src/core/memtrack.c` has the full 128 KB device, a save-state flag and libretro `SAVE_RAM` wiring. I was wrong to say it wasn't.

The conclusion still holds, but for a different reason: all three dispatch sites gate on `jaguarMainROMCRC32 == 0xFDF37F47`, i.e. the loaded cartridge being the Memory Track dump itself — which can never be true while a disc is loaded. So the device is present in the code but unreachable for CD content, which is the one configuration where hardware actually uses it.

Filed separately as #258 with the exact gate and a suggested direction. Still unrelated to this PR.

Issue #296, opened 2026-08-04
CD real-BIOS mode: black screen (user report — needs repro details)


JoeMatt:
### What

@jhdgghost reported in #294 that booting with the **real CD BIOS** produces a black screen (HLE BIOS works for them, apart from FMV titles — tracked separately).

This doesn't reproduce headlessly: the CD boot matrix (`test/tools/cd_boot_matrix.sh`) currently has all tracked titles reaching game code in **bios mode**, and real-BIOS boot needs no external files (both retail and developer CD BIOSes are embedded). So this needs reporter details to make progress.

### Needed to reproduce

@jhdgghost — could you provide:

1. **Which game(s)** and image format (CUE/BIN or CDI), ideally with the redump name.
2. **Platform/frontend** (OS, RetroArch version) and core version (`v3.0.0`? nightly?).
3. **Core options**: `Boot BIOS`, `CD BIOS Type`, blitter mode.
4. A **RetroArch log** (Settings → Logging → Core Logging Level: Info, log to file) from a black-screen boot. The core's crash watchdog logs a signature line (`gpu_pc_escape`, `dsp_wedge`, `video_stall`, `cd_seek_wedge`, …) at the moment things go wrong, and the `[CD-BIOS]` lines will show whether an external BIOS file was picked up.
5. Whether any Jaguar CD BIOS file exists in your RetroArch `system` folder (or its `Atari - Jaguar`/`jaguarcd` subdirs). An external file **overrides the embedded BIOS and is selected by filename only** — a wrong or byte-swapped 256 KiB file with a plausible header would boot to black. Renaming it away is a quick A/B test.

### Hypotheses (unverified)

- Mislabelled/corrupt external BIOS file shadowing the embedded one (see point 5).
- A per-title bios-mode failure not covered by the boot matrix.
- FMV-title black window misread as a hang (Myst's intro has a ~6 s all-black pause).

Split out from #294.

JoeMatt:
Ran the "bad external BIOS file" hypotheses headlessly against develop (d1058d8), Hover Strike CUE/BIN, `cd_boot_mode=bios`, scratch system dirs. Summary of which failure modes a user can actually reach:

| System-dir contents | Loader behavior | Outcome |
|---|---|---|
| empty | embedded retail BIOS | boots to game code (baseline) |
[...]

PR #299, opened 2026-08-04
fix(cd): validate external CD BIOS content, accept developer BIOS dumps

JoeMatt:
## Summary

Fixes both external CD BIOS loader defects confirmed empirically in #296.

The loader validated a candidate file on one signal: the 32-bit run address at offset `$404` had to land in `[$800000,$840000]`. That has two failure modes:

1. **Genuine developer CD BIOS dumps were always rejected.** The developer BIOS (crc32 `55A0669C`) carries `$FFFFFFFF` at `$404`, so the run-addr test can never pass for it. A user dropping a real dev BIOS in `system/` got the misleading `CD BIOS not found` warning and a silent fall back to the embedded image — their file could never be used.
2. **A 256 KiB file of arbitrary content with a plausible long at `$404` was accepted** and staged as the CD BIOS, producing the user-reported cube-animation-then-permanent-black-screen (watchdog logs `dsp_pc_escape` / `gpu_pc_escape` at `$8xxxxxxx`).

Validation is now content-based, in order:

- **(a)** crc32 matches a known dump (`687068D5` retail, `55A0669C` developer) → accept, logging which revision was recognized.
- **(b)** unknown crc32 but plausible run address → accept as an unrecognized revision, behind a loud warning naming the file as prime suspect if boot black-screens immediately after. Genuine revisions/regions we don't have CRCs for still load.
- **(c)** neither → reject, keep searching, fall back to embedded.

The final "no BIOS" warning now distinguishes *nothing was there* from *N candidate file(s) rejected*, so a bad file no longer looks like a missing one.

Reuses `crc32_calcCheckSum()` from `src/core/crc32.c` — no new checksum code. Filenames still drive search order only; the log now reports the revision a file actually contains rather than what its name claims.

## Testing

`bash scripts/c89-lint.sh libretro.c` → passed. Clean build. `make TEST_EXPORTS=1 test` → **exit 0**, 19 result blocks, 0 failed.

Headless matrix (Hover Strike: Unconquered Lands, `virtualjaguar_cd_boot_mode=bios`, 1200 frames, scratch system dirs):

| System-dir contents | Result |
|---|---|
| **developer BIOS file** | **now accepted** — `using external ... (recognized developer CD BIOS, crc32=55A0669C)`; identical per-window timeline and audio RMS to the embedded dev image (control run) |
| retail BIOS file | still accepted — `(recognized retail CD BIOS, crc32=687068D5)`, boots to game code (71.8% non-black, RMS 1544) |
| 256 KiB junk w/ plausible `$404` | accepted only behind the new `unrecognized CD BIOS revision (crc32=8CE070BE ...)` warning; the expected `dsp_pc_escape`/`video_stall` cascade follows, now traceable to the named file |
| 2 bad files (bad `$404` + truncated) | `no usable CD BIOS in ...: 2 candidate file(s) rejected (size or header); using embedded BIOS` → embedded fallback boots |
| empty | unchanged — `CD BIOS not found` → embedded retail → boots |

Note on the dev-BIOS case: both the external file and the embedded dev image render through frame ~479 then blank on this title. That blackout is **pre-existing dev-BIOS behavior, identical on both paths** and outside this fix's scope — the point proven here is that the external file is now loaded at all.

## Notes / follow-ups

- Tier (b) still *accepts* junk by design, so #296's black screen becomes **diagnosable, not prevented**. Hard-rejecting unknown CRCs would break genuine alternate dumps.
- `bios_boot()` sets `jaguarRunAddress = GET32(jagMemSpace, 0x800404)`, which is `$FFFFFFFF` for the developer BIOS. It appears inert on this path (boot ROM is forced on) and behaves identically for the embedded dev image, so it predates this change — worth a separate look if dev-BIOS titles are pursued.

Closes #296

### DIFF
--- a/libretro.c
+++ b/libretro.c
@@ -18,6 +18,7 @@ int64_t rfread(void* buffer, size_t elem_size, size_t elem_count, RFILE* stream)
 
 #include "cheat.h"
 #include "crash_detect.h"
+#include "crc32.h"
 #include "bus_arbiter.h"
 #include "file.h"
 #include "jagbios.h"
@@ -1134,13 +1135,22 @@ static bool has_extension(const char *path, const char *ext)
    return strcasecmp(dot + 1, ext) == 0;
 }
 
+/* CRC32s of known-good 256 KB CD BIOS dumps.  The developer BIOS has
+ * $FFFFFFFF at the $404 run-address slot, so it can only be recognized
+ * by checksum -- a header check alone always rejects it. */
+#define CD_BIOS_CRC_RETAIL 0x687068D5u
+#define CD_BIOS_CRC_DEV    0x55A0669Cu
+
 /* Try to load a 256 KB CD BIOS image from the given path.
- * Returns true on success and sets cd_bios_loaded_externally. */
-static bool try_load_cd_bios_file(const char *path)
+ * Returns true on success and sets cd_bios_loaded_externally.
+ * A file that exists but fails validation bumps *rejected so the
+ * caller's final warning can distinguish "no file" from "bad file". */
+static bool try_load_cd_bios_file(const char *path, int *rejected)
 {
    RFILE   *f;
    int64_t  size;
    uint32_t run_addr;
+   uint32_t crc;
 
    f = rfopen(path, "rb");
    if (!f)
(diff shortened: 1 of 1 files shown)


JoeMatt:
@copilot review

PR #308, opened 2026-08-05
test: make silently-skipping checks visible; ten real-BIOS assertions were inert

JoeMatt:
## Why

The Skyhammer clipping check had been **silently SKIPping** while the suite reported exit 0. Skyhammer is the repo's canonical "loud broken audio" sentinel — CLAUDE.md names it as the title that must keep failing until the audio is genuinely fixed. That sentinel has been inert.

That's the same masked-failure class as PR #170 (a "fix" took Iron Soldier 2 from clipping to silent; the clipping test passed because silence has 0% saturation). **A skip that reads as a pass is that failure wearing a different hat.** So this audits the whole suite for it.

## The big find — ten real-BIOS assertions were inert

`test_bios_config` and `test_boot_config` hardcoded `test/roms/private` as the system directory, but the corpus keeps BIOS dumps in `ROMS/`. **Ten real-BIOS assertions had been reporting SKIP with the full corpus present.** They now run and pass — verified in the log: the BIOS dir resolves to `test/roms/private/ROMS`, the external CD BIOS is found and CRC-recognised (`crc32=687068D5`), and `real_bios_init_succeeds` / `real_bios_boot_rom_space_accessible` execute.

`test_audio_pipeline` was also permanently vacuous — it took an *optional* positional ROM and was invoked bare, so its onset check and BIOS-vs-HLE comparison never ran, with or without ROMs. It now runs real comparisons (`Onset similar: HLE=13 BIOS=1`, `Volume comparable: ratio=0.96`).

**Residual skips with the corpus present: 13 → 1.**

## Skyhammer: not a spelling problem

There is no Skyhammer ROM in the corpus, under any spelling — two exhaustive searches across all 47 directories found nothing as cart or CD image. So it can't be fixed by correcting a filename; it's reported as a named skip instead.

Lookup is now robust via `scripts/find-rom.sh` (case-insensitive patterns in order, filtered to loadable media, ranked by path depth). Two further bugs surfaced while building it: a bare `Atari Karts*` matched `7zips/*.7z` — an archive the harness cannot load, i.e. **a check that appeared to run but tested nothing** — and `Iron Soldier*` matched a duplicate buried in `ROMS/Atari Jaguar Rom Collection/ROMS/`.

## Skip visibility

`scripts/test-skip.sh` prints a named summary of everything that did not run. The ledger is truncated at the **start** of each run — a stale row reading as a fresh skip is exactly the phantom-row bug `cd_boot_matrix.sh` already had.

Non-fatal by default so CI (which has no private ROMs) stays green; `VJ_REQUIRE_ROMS=1` escalates.

Also removed the `[ -f ]` guards on `yarc.j64` and `jagniccc.j64`: both are committed in-tree, so absence means a broken checkout and must fail — the rationale `test_state_compat` already states.

## Verified independently before merge

- Corpus present: `make TEST_EXPORTS=1 test` **exit 0**, exactly **1 SKIP line** suite-wide, audio pair by name (`Clipping check: Iron Soldier 2`, `Presence check: Iron Soldier 1` — that banner is new; `--quiet` was suppressing it, so the suite previously emitted a bare `PASS: audio is present...` naming no title)
- Corpus absent: **7 skips reported by name**, suite still exit 0 (CI parity)
- `VJ_REQUIRE_ROMS=1` with corpus absent: **exit 2**
- Symlink restored afterwards, target verified intact
- c89-lint passes on all three `.c` files; the `&&` chains were checked against five stub-injection cases so a real BIOS-test failure cannot be swallowed

## Two things left open deliberately

1. **CLAUDE.md contradicts the code on Skyhammer.** The doc says it must keep *failing* clipping; the Makefile comment and the MMULT-fix notes say it was resolved and should assert clean. With no ROM, neither is testable. Wired as assert-clean (identical to IS2), and no XFAIL machinery was built on the strength of a stale doc. **Either CLAUDE.md gets corrected or the ROM gets added** — otherwise the next person re-litigates this.
2. **`netlink_latency_test` is pre-existing flaky** (failed 2 of 5 full runs, passed 3/3 in isolation) and runs early, so when it flakes the suite aborts *before* the skip summary prints. Untouched here — but don't read "no summary" as "no skips."


JoeMatt:
Updated: the maintainer pointed out Skyhammer **is** in the corpus, as `ROMS/Skyhammer (World).j64`. The "not present under any spelling" note came from searching a worktree that had no `test/roms/private` symlink — the search was against an empty path. The new case-insensitive lookup finds it correctly.

That means the CLAUDE.md-vs-Makefile disagreement is now settled **on evidence** rather than left open:

```
=== Clipping check: Skyhammer ===
  ROM:    test/roms/private/ROMS/Skyhammer (World).j64
  First audio at frame: 171
  Saturated (|s|>=32760):  0 (0.000%)
  Longest saturation run: 0 samples
  Window RMS (combined): 3079.8
  PASS: no clipping signature in window
```

Skyhammer **no longer clips** — the MMULT secondary-bank fix resolved it — and it is **not silent** either (RMS 3079.8, audio from frame 171), so it isn't the masked-silence failure mode. Asserting clean is correct on merit, not by default. CLAUDE.md's "Skyhammer should still fail clipping until it's actually fixed" was stale *and* inert regardless, since the Makefile looked for `Skyhammer_(1999).jag`, which matches nothing.

Both the Makefile comment and CLAUDE.md are corrected in `425de93`, and no expected-fail wrapper was added.

The suite now reports:

```
```

Also confirming the second caveat from the PR body, since it showed up during verification: `netlink_latency_test` failed one run at `disabled=48.6 enabled=66.5` — a **1.46× ratio against a 1.5× threshold**, while three agents were loading the machine — and because it runs early, the suite aborted **before** the skip summary printed. Re-ran on an idle machine: `PASS (disabled=48.6 enabled=87.5)`. Pre-existing and untouched here, but it is a genuine flake that will produce spurious CI failures, and "no skip summary" means "the suite aborted early", not "nothing skipped".

Issue #321, opened 2026-08-05
[perf] Prune the unused 68K duplicate table + dead handlers from the UAE core (~>300KB)


JoeMatt:
### What

Prune the unused half of the UAE 68K core: duplicate handler tables, dead handler bodies, and the linker hygiene to keep them gone. Recommendation from a RetroArch dev (the UAE ifdef structure exists precisely so integrators bake in only what their system uses); verified against this tree before filing.

### Verified facts (2026-08-05, develop)

- **Two complete 68000 handler tables ship; one is dispatched.** `src/m68000/cpustbl.c` defines `op_smalltbl_4` (fast, **3,162 entries**) and `op_smalltbl_5` ("slow but compatible", **1,581 entries**). `m68kinterface.c:3823` binds `op_smalltbl_5_ff` unconditionally. Table 4 and every handler reachable only from it — inside the 1.8 MB machine-generated `cpuemu.c` (compiles to **1.34 MB** of object) — is linked and never called. The 68010/020/040 tables are already stubbed out (externs commented at `m68kinterface.c:3213-3216`).
- **FPU/MMU: nothing to rip.** No `fpp`, no FPU handler files; 4 vestigial identifier references. This fork is already a fixed-point 68000.
- **Floats: zero** `float`/`double` in `src/m68000/*.c` — core-level determinism already holds.

So the actionable scope is the duplicate table, not a kitchen-sink teardown. Plausible win is **well above the suggested 300 KB** given the object sizes.

### Suggested order (cheapest, least-risk first)

1. **Linker GC, zero source changes**: `-ffunction-sections -fdata-sections` + `-Wl,--gc-sections` (`-dead_strip` on Darwin) in the Makefile. Because `op_smalltbl_4` is a const array only *declared* extern and never referenced by live code, the table and its exclusive handlers should GC automatically. Measure per-platform binary delta; watch the TEST_EXPORTS wide-ABI link (`exports-test.list` keeps test symbols alive) and any platform linkers in the 16-target release matrix that lack gc-sections.
2. **Physical pruning** if step 1 leaves residue or is unavailable on some targets: delete the `op_smalltbl_4` table + its extern, then scripted dead-code elimination of unreferenced `CPUFUNC` bodies in `cpuemu.c`. Note `cpuemu.c` is committed pre-generated (the UAE gencpu generator is not in-tree) — pruning must be scripted-mechanical against the file, never hand-edited, and the script committed under `scripts/` for reproducibility.
[...]

PR #326, opened 2026-08-06
perf(build): let the linker garbage-collect the dead 68K handler table

JoeMatt:
Closes #321.

`src/m68000/cpustbl.c` ships two complete 68000 handler tables — `op_smalltbl_4_ff` (3162 entries, "fast") and `op_smalltbl_5_ff` (1581, "slow but compatible"). `m68kinterface.c` binds table 5 unconditionally, so table 4, plus every machine-generated `cpuemu.c` handler body reachable only through it, has been linked into every shipped binary and never called.

Turning on linker dead-code elimination drops it. **No source change** — this is entirely a Makefile change.

## Results

macOS arm64:

| | before | after | delta |
|---|---|---|---|
| production ABI | 2,106,528 | 1,627,648 | **−478,880 (−22.7%)** |
| test ABI | 2,114,016 | 1,658,272 | −455,744 (−21.6%) |
| `__text` | 803,868 | 515,736 | −288,132 (−35.8%) |
| `__DATA_CONST.__const` | 68,336 | 42,992 | −25,344 (= the table) |

`op_smalltbl_4_ff` is gone from the linked image; `op_smalltbl_5_ff` stays.

## Why `GC_STYLE` is opt-in per platform

The release matrix spans 16 toolchains while PR CI covers four, so a globally-applied flag that one exotic linker rejects fails at tag time, not in review. `GC_STYLE` is set only where the platform's own `SHARED` line already passes GNU-ld-only options (`--version-script` / `--no-undefined`, which proves GNU ld or lld), or where the linker is Apple ld64.

- `gnu` — `-ffunction-sections -fdata-sections` + `-Wl,--gc-sections`
- `macho` — `-Wl,-dead_strip` only; ld64 is atom-based, so the section flags are no-ops and are deliberately not added
- unset — byte-for-byte unchanged: every `STATIC_LINKING=1` target (vita, switch/libnx, ctr, ps3, psl1ght, psp1, emscripten — those emit a `.a`/`.bc` and never link), every MSVC/Xbox target, Solaris ld, theos_ios

Android is unaffected either way; ndk-build uses `jni/Android.mk`.

## Second commit: classic_armv7_a7

That block was passing `-Wl,--gc-sections` in **CFLAGS**, but the link rule is `$(LD) $(LINKOUT)$@ $(OBJECTS) $(LDFLAGS)` — CFLAGS never reaches it. The target has been paying the `-ffunction-sections`/`-fdata-sections` fragmentation cost and getting none of the GC benefit for as long as the flags have been there.

The first commit deliberately left it alone because the target has no CI and no local toolchain. It turns out to be buildable: a Debian bookworm container with `gcc-arm-linux-gnueabihf` cross-links it. Measured there (gcc 12 — a **proxy** for the modmyclassic toolchain, not that toolchain):

| variant | size | delta |
|---|---|---|
| flag in CFLAGS (as-is) | 1,512,628 | — |
| flag deleted entirely | 1,512,628 | **0 — provably inert** |
| flag moved to LDFLAGS | 1,512,548 | −80 bytes (0.005%) |

All three link exit 0 with a byte-identical exported dynsym list.

**The macOS 22.7% does not transfer here.** That block compiles with `-flto=4 -fwhole-program`, so `cpustbl.o` is a slim LTO object and GCC's IPA — not the linker — performs the elimination: `op_smalltbl_4_ff` is already absent from the baseline `.so`. The move is worth landing because a flag that reads as enabled should be enabled, not for the 80 bytes.

Moved to LDFLAGS by hand rather than via `GC_STYLE := gnu`, which would duplicate the `-ffunction-sections -fdata-sections` the block already carries.

## Testing

- 56,103-case opcode sweep passes
- `make TEST_EXPORTS=1 test` exits 0, "Skipped checks: none", both audio clipping and audio presence green
- Exported symbol sets identical for **both** ABIs (46 prod / 422 test)
- Bit-identical 600-frame boot A/B — video hash, audio hash, sample counts, non-black pixel counts, geometry — on yarc, Iron Soldier, AvP, Doom and Tempest 2000
- classic_armv7_a7 cross-links clean; macOS still gets `-dead_strip`, `platform=unix` still gets `--gc-sections`

## Caveat

`-fwhole-program` on a `-shared` build in the classic_armv7_a7 block is a sketchy combination. Exported dynsyms came back identical to baseline, so it is not biting today — pre-existing and untouched here, flagging it only so it is on the record.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -103,9 +103,11 @@ ifeq ($(platform), unix)
 	TARGET := $(TARGET_NAME)_libretro.so
 	fpic := -fPIC
 	ifneq ($(findstring SunOS,$(shell uname -a)),)
+		# Solaris ld: no --gc-sections, so GC_STYLE stays unset.
 		SHARED := -shared -z defs -z gnu-version-script-compat
 	else
 		SHARED := -shared -Wl,--no-undefined -Wl,--version-script=$(LINK_SCRIPT)
+		GC_STYLE := gnu
 	endif
 
 # Classic Platforms ####################
@@ -145,6 +147,7 @@ else ifeq ($(platform), osx)
 	TARGET := $(TARGET_NAME)_libretro.dylib
 	fpic := -fPIC
 	SHARED := -dynamiclib $(MACHO_EXPORTS_FLAGS)
+	GC_STYLE := macho
 	ifeq ($(arch),ppc)
 		FLAGS += -DMSB_FIRST
 		OLD_GCC = 1
@@ -172,6 +175,7 @@ else ifneq (,$(findstring ios,$(platform)))
 	TARGET := $(TARGET_NAME)_libretro_ios.dylib
 	fpic := -fPIC
 	SHARED := -dynamiclib $(MACHO_EXPORTS_FLAGS)
+	GC_STYLE := macho
 	MINVERSION :=
 	ifeq ($(IOSSDK),)
 		IOSSDK := $(shell xcodebuild -version -sdk iphoneos Path)
@@ -196,6 +200,7 @@ else ifeq ($(platform), tvos-arm64)
 	TARGET := $(TARGET_NAME)_libretro_tvos.dylib
 	fpic := -fPIC
 	SHARED := -dynamiclib $(MACHO_EXPORTS_FLAGS)
+	GC_STYLE := macho
 	ifeq ($(IOSSDK),)
 		IOSSDK := $(shell xcodebuild -version -sdk appletvos Path)
 	endif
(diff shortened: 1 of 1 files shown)


JoeMatt:
Rebased onto current develop (`d0c9438`). **Heads-up: before the rebase this branch would have reverted #325** — it was based on pre-#325 develop, so its diff deleted `test/test_cd_synth_subq.c` and 172 lines of `cdrom.c` (the VLM Q-subcode work). Now Makefile-only, as intended.

Re-verified on the new base, since linker GC interacts with everything that links:
- prod dylib **1,644,368 bytes**; `nm -pa` shows only `_op_smalltbl_5_ff` — the dead table is still collected
- **opcode sweep 56,103/56,103**
- suite exit 0, "Skipped checks: none", and the new `test_cd_synth_subq` 6/6 (i.e. GC didn't strip anything the VLM path needs)

PR #327, opened 2026-08-06
chore(info): savestate_features 2 -> 3, with the test that backs it

JoeMatt:
Spotted while updating the [libretro/docs page](https://github.com/libretro/docs/pull/1173): `dist/info/` claimed `savestate_features = "2"` ("serialized") while `retro_load_game` reports **zero** serialization quirks and its comment claims run-ahead. Zero quirks is the level-3 ("deterministic") assertion, so the two disagreed. This makes them agree — but verifies the claim first rather than trusting the comment.

## The test

`test/tools/test_runahead_determinism.c` reproduces RetroArch's run-ahead loop directly: warm up, serialize, run a window while digesting every frame's video and audio, `retro_unserialize` back, replay the same window with the same (empty) input — then roll back and replay a **third** time.

That third pass is the load-bearing one. Pass 1 is the only pass reached *without* an intervening `retro_unserialize`, so anything the load path fails to restore shows up as a pass-1-vs-pass-2 difference and nowhere else. Run-ahead rolls back on every frame, so it lives entirely in the post-rollback regime that pass 2 vs pass 3 measures.

## Results

Identical on Iron Soldier, Skyhammer and Kasumi Ninja (240 warmup / 120 window). `yarc.j64` passes all four.

```
PASS  video_replay_identical     120/120 frames bit-identical
FAIL  audio_replay_identical     1 frame differs
PASS  state_reconverges          2490368 bytes, byte-identical
PASS  repeated_rollback_agrees   pass 2 == pass 3, video and audio
```

The single failure is the **first frame after the first rollback**, off by ~0.05% RMS (Iron Soldier: 2079.86 vs 2080.96) with an identical sample count. Frames 1..119 match bit for bit, and successive rollbacks agree completely including that frame.

So the level-3 claim holds for everything run-ahead actually depends on, and there is a real but bounded gap: some audio state is re-derived rather than restored by the load path.

## What is already ruled out

The I2S ring buffer, despite being absent from `DACStateSave`. `DACPrepareFrame` reseeds `i2sWritePos`/`i2sWriteCount` to 2 every frame, and `DSPSampleCallback` clamps its read index to `i2sWriteCount` — so no sample from a previous frame can ever be read. The culprit is elsewhere in the DSP/DAC path.

## Wiring

The test gets its own `make runahead-determinism` target and is deliberately **not** in `make test` while that assertion fails — a permanently-red suite entry is worse than no entry. Wire it in once the gap is closed.

```bash
make runahead-determinism RUNAHEAD_ROM="path/to/game.j64"
```

No core source is touched by this PR — only the `.info` value, the Makefile target, and the new test.


JoeMatt:
## Root-caused and fixed — all four assertions now pass

The instrumented diff pinned it immediately. Dumping every input the frame's audio is derived from at `DACPrepareFrame`, then diffing a natural run against a post-rollback replay of the same frame:

```
  ltxd       pass1=FF57   pass2=FBFB
  rtxd       pass1=0173   pass2=F8A3
```

Every DSP field was identical — pc, control, flags, accumulator, modulo, remainder, bank, Z/N/C, pipeline pointers, scoreboard, r0/r1/r30/r31. Only **LTXD/RTXD** differed.

### The bug

`ltxd`/`rtxd` point into `jagMemSpace` at `$F1A148`/`$F1A14C` (`vjag_memory.c:177`). `retro_serialize` saves three buffers: `jaguarMainRAM` (the low 2 MB of `jagMemSpace`), `tomRam8`, and `jerry_ram_8` — and **`jerry_ram_8` is a separate array declared in `jerry.c:187`, not the `$F10000` window of `jagMemSpace`**.

So the entire DAC register file (LTXD, RTXD, SCLK, SMODE, LRXD, RRXD, SSTAT) was in *no* saved region. It survived a load only by accident, as whatever the previous run left behind.

`DACPrepareFrame` seeds the resampler's interpolation endpoints from LTXD/RTXD every frame, so the first frame after a rollback started from the wrong endpoints and converged once real DSP writes landed — exactly the observed one-frame, ~0.05% RMS, audio-only signature.

### Fix

Serialize them in `DACStateSave`/`DACStateLoad` behind `STATE_VERSION_DAC_REGISTERS` (v8). Older layouts consume nothing and keep current behaviour. SCLK/SMODE feed the resample ratio, so the loader re-derives it via `DACUpdateSCLKRate` once they hold restored values.

### Verification

All four assertions green on Iron Soldier, Skyhammer, Kasumi Ninja and yarc:

```
PASS  video_replay_identical     120/120 frames bit-identical
PASS  audio_replay_identical     120 audio batches compared
PASS  state_reconverges          2490368 bytes, byte-identical
PASS  repeated_rollback_agrees   pass 2 == pass 3
```

The test now joins `make test` — yarc unconditionally (in-tree, never skips) plus Iron Soldier when the private corpus is present.

`test_state_compat`'s deliberate DAC-block-size tripwire fired at 37 → 51 bytes, exactly as it was written to. Its expected size and layout comment are updated; the v1/v2/v3 fixture offsets are unaffected since the fields are appended, not interleaved.

Audio pair re-verified per CLAUDE.md, since this touches `dac.c`: **Iron Soldier 1 presence RMS 1175.7** (develop baseline ~1175, envelope 200–25000) and **Skyhammer clipping 0 saturation runs**. Full suite exits 0 with `Skipped checks: none`.

### What this closes

`savestate_features = 3` is now backed by a green test rather than by an argument about which property run-ahead depends on. Worth noting the bug was never run-ahead-specific — **every ordinary save-state load** restored stale DAC registers; it just produced one frame of slightly-wrong audio, which nobody would notice by ear. It took replaying the same frame twice to make it visible.

Four candidates that looked guilty and were not (the D_FLAGS retire delay was implemented, measured, and reverted) are recorded in the test header so nobody re-spends the time.

PR #385, opened 2026-08-10
fix(cart): validate JST_ROM entry vector before trusting it as PC

JoeMatt:
First piece of the HLE-vs-BIOS boot-parity class found by the silent-audio sweep (4 titles work under Real BIOS but fail under HLE).

## The fix

`JaguarLoadFileInternal` unconditionally read the 68K entry vector from cart+$404. Rayman Demo doesn't follow the header+vector convention — cart+$400 onward is real 68K code, and cart+$404 decodes to $3BE800F0 → masked $E800F0, just past the boot-ROM window. In HLE that garbage became the initial PC. Now: a candidate vector outside every executable band (RAM mirror / cart / boot ROM) is rejected and the entry falls back to cart base $800000 (verified correct for this ROM — $802000 would land on unformatted $FFFF flash and fault instruction one). Real-BIOS mode never consults this path.

**Blast radius measured, not assumed:** corpus scan of 68 JST_ROM-sized images — only Rayman Demo (CRC-identical across all 3 copies, so not a bad dump) and one already-dead PD image fall outside the valid range; 10-title spot check all healthy; full suite exit 0 zero-skip including both mandatory audio tests; c89-lint clean.

**Honest status:** Rayman Demo now executes ~11 frames of its real code, then hits a second, unrelated deterministic fault (PC parks at $802BE6) — still broken for the user, but now debuggable past boot. The matrix row will shift `? (pc_escape)` → `GAME_CODE (black)` on next regen; that's reclassification, not a working game.

## The rest of the class (documented, deliberately not "fixed" here)

- **White Men Can't Jump** (silent in HLE): the DSP never starts in HLE — by design; `DSPReadLong/Word` auto-ack sound commands exactly when `!DSP_RUNNING && !useJaguarBIOS` so titles proceed silently instead of hanging. Naively setting DSPGO would break that gating; the real fix is a persistent HLE DSP audio engine, a previously-attempted-and-reverted design (`docs/emulation-bug-hunt-todos.md`).
- **Flip Out**: runs ~300 frames then exception-loops at $000404 — characterized, not isolated.
- **Ruiner Pinball**: DSP runs fine; the divergence is *video* (HLE 0% non-black vs BIOS 21.5% despite the game writing OLP $7B80). Also: `emulation-bug-hunt-todos.md`'s claim that Ruiner behaves identically in both modes is stale — freshly contradicted at b48f854.

### DIFF
--- a/src/core/file.c
+++ b/src/core/file.c
@@ -294,13 +294,20 @@ static bool JaguarLoadFileInternal(uint8_t *buffer, size_t bufsize)
        * cart base -- the other common convention, and where this demo's
        * real startup code (VMODE/VI register writes) actually lives.
        * Note: $DFFF00-$DFFFFF is the CDROM overlay, not cart ROM; the
-       * cart window ends at $DFFEFF. */
+       * cart window ends at $DFFEFF.
+       *
+       * An odd candidate is rejected for the same reason: the 68000 cannot
+       * fetch an instruction from an odd PC, and JaguarExecute() bails out
+       * immediately whenever (m68kPC & 1), so an in-range but misaligned
+       * vector (say $800001 out of a raw byte stream) would leave the CPU
+       * parked instead of running the game.  Those take the fallback too. */
       {
-        uint32_t candidate = GET32(jagMemSpace, 0x800404);
-        uint32_t masked = candidate & 0x00FFFFFF;
-        bool validVector = (masked < 0x200000)
-              || (masked >= 0x800000 && masked <= 0xDFFEFF)
-              || (masked >= 0xE00000 && masked <= 0xE1FFFF);
+         uint32_t candidate = GET32(jagMemSpace, 0x800404);
+         uint32_t masked = candidate & 0x00FFFFFF;
+         bool validVector = ((masked & 1) == 0)
+               && ((masked < 0x200000)
+                  || (masked >= 0x800000 && masked <= 0xDFFEFF)
+                  || (masked >= 0xE00000 && masked <= 0xE1FFFF));
 
          if (validVector)
             jaguarRunAddress = candidate;
--- a/src/core/file.c
+++ b/src/core/file.c
@@ -269,8 +269,47 @@ static bool JaguarLoadFileInternal(uint8_t *buffer, size_t bufsize)
       jaguarCartInserted = true;
       memcpy(jagMemSpace + 0x800000, buffer, flatSize);
       JGDLoadROM(buffer, jaguarROMSize);
-      // Checking something...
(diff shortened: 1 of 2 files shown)


JoeMatt:
@copilot review

JoeMatt:
Both remaining Copilot nits addressed in f69563f.

**1. Odd candidates were accepted.** Correct catch — a 68000 cannot fetch an instruction from an odd PC, and `JaguarExecute()` returns immediately on `(m68kPC & 1)` (`src/core/jaguar.c:346`), so an in-range but misaligned vector like `$800001` passed the range check and left the CPU parked. The predicate now requires even alignment before the band test:

```c
bool validVector = ((masked & 1) == 0)
      && ((masked < 0x200000)
         || (masked >= 0x800000 && masked <= 0xDFFEFF)
         || (masked >= 0xE00000 && masked <= 0xE1FFFF));
```

Odd candidates take the same `$800000` cart-base fallback. The comment block was extended with the rationale so the prose and the predicate stay in step.

**2. Fallback branch had no test coverage.** Added Test 20 (`Cart Entry Vector Validation`) to `test/test_hle_bios.c`, following the existing `test_memcon1_width_bits(dummy_rom)` pattern — reuses the 128K dummy cart, rewrites `dummy_rom[$404..$407]`, reloads, and asserts on both `jaguarRunAddress` and the reset PC vector at `RAM[4]`:

| cart+$404 | case | expected entry |
|---|---|---|
| `$3BE800F0` | out of range (the measured Rayman Demo value) | `$800000` |
| `$00802001` | in the cart window but odd | `$800000` |
| `$00802000` | valid, even, in range | `$00802000` (taken verbatim) |

The third row is the positive control — the new conjunct must not reject good vectors — and it also restores the shared `dummy_rom` buffer for the teardown load.

**Verified it is a real guard, not a tautology:** with the alignment conjunct reverted and everything else unchanged, the odd case fails both of its asserts:

```
=== Test 20: Cart Entry Vector Validation ===
  PASS: cart+$404 = $3BE800F0 -> jaguarRunAddress $00800000
  FAIL: cart+$404 = $00802001 -> jaguarRunAddress $00802001 (expected $00800000)
  FAIL: Reset PC vector RAM[4] = $00802001 (expected $00800000)
=== Results: 223 passed, 2 failed ===
```

With the fix in place, `test_hle_bios` is 225 passed / 0 failed, and the full suite is exit 0 with **zero skipped checks** (run against the private ROM corpus, so both mandatory audio tests actually executed rather than silently skipping):

```
```

`bash scripts/c89-lint.sh src/core/file.c test/test_hle_bios.c` → `C89 lint passed`. Clean `make` build, no new warnings.

PR #119, opened 2026-04-23
Fix emulation accuracy: resolution, 68K instructions, events, RAM init, logging

JoeMatt:
## Summary
- **TOM resolution pipeline**: Use HDB1/HDE/VDB/VDE registers with pwidth for dynamic resolution instead of hardcoded constants — fixes black screen / cyan bar on many ROMs (e.g. 240p test suite)
- **68K instruction emulation**: Emulate MULL/DIVL (68020+) via IllegalOpcode trap; fix BSR.L absolute address for Atari aln linker — needed by modern homebrew
- **Event system**: Fix GetTimeToNextEvent() reading uninitialized slot[0] time without validity check, causing periodic execution bursts and visible stuttering; zero eventTime on init
- **RAM init**: Skip randomization over loaded executable region for ABS/COFF files loaded at $4000+
- **JERRY timers**: Fix PIT2 using PIT1's prescaler/divider; disable JERRY trace spam (was on every 44kHz interrupt)
- **Libretro logging**: Add log.h framework (LOG_DBG/INF/WRN/ERR) using retro_log_printf_t so verbosity is toggleable in RetroArch UI; convert fprintf(stderr) calls in DAC, GPU, JERRY, TOM
- **GPU**: Add trace debug guard macro (disabled by default), GPUDumpState() diagnostic function

## Test plan
- [x] Builds clean on macOS (Clang) and Linux (GCC)
- [x] 240p test suite boots and renders correctly (was black screen on master)
- [x] Scrolling bars test runs without periodic stuttering (event system fix)
- [x] Regression test against Tempest 2000 and other known-working titles
- [ ] CI passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #391, opened 2026-08-10
fix(dac): kill the 60 Hz audio crackle on cartridge titles

JoeMatt:
## The bug (user report: "sound is atrociously crackly", Atari Karts, Android nightly)

Every frame, the audio stream stepped discontinuously — a **60 Hz impulse train** heard as constant crackle on cart titles. Atari Karts in-race measured a mean seam step of ~2076 against an interior sample-to-sample mean of ~80 (26x). Super Burnout was the extreme case: its output carried **no waveform at all** — interior delta exactly 0.0, a pure per-frame staircase.

Neither existing audio test can see this class: it is not saturation (`test_audio_clipping` passed Karts at 0.000% saturated) and not silence (`test_audio_presence` saw normal RMS). That is how it shipped for months — it predates the CD merge (the structure came in with #119).

## Three root causes, one per commit concern

1. **Per-write I2S capture** (`DACWriteWord`): the I2S port is a holding register — hardware transmits one pair per word strobe; writes in between never reach the DAC (JTRM, docs/jtrm-jerry.md "Audio Pipeline"). We captured on every RTXD write: Karts writes 417 pairs/frame against 346 word strobes (SMODE=\$0005, SCLK=19 → 20774 Hz), so the ring grew ~20% faster than the resampler consumed and `DACPrepareFrame` discarded **17% of every frame's audio**, yanking the phase forward once per frame. Now `DACWordStrobe()` latches from `JERRYI2SCallback()` — capture and consumption rates are equal by construction. This is the master-mode twin of the slave-mode "60 Hz chop" already documented as fixed in `DACUpdateSCLKRate()`.

2. **Fabricated 800th sample** (`SoundCallback`): an NTSC field is 524 halflines = 16651.56 µs = **799.27** periods of the 48 kHz sample clock, so a fixed 800-sample batch is unsatisfiable — the last sample was fabricated from a raw register read. Batches are now variable-length (799/800); the frontend's dynamic rate control absorbs the remainder, which is its job.

3. **Per-frame resampler reset** (`DACPrepareFrame`): the ring and read phase were reset and re-anchored at the newest register value every frame. Both cursors are now monotonic and persist across frames, with a gross-drift resync for SCLK reprogramming / savestate loads / DSP silence.

## Savestate v9

The ring contents join the state — the replay-determinism test caught that the cursors alone reference data a load cannot reconstruct. `STATE_SIZE` grows 0x260000 → 0x280000; `retro_unserialize` floors its size check at the v8 layout so **every pre-v9 state still loads** (ring zeroed, loaded cursor pair kept — mutually consistent in old states).

## New regression test

`test_audio_boundary` — fails when audio-batch seams are statistically discontinuous (seam mean > 4x interior mean + absolute floor). Wired into `make test` with the find-rom/skip-ledger pattern. Negative control: **10.05x FAIL** on the pre-fix core, **0.98x PASS** fixed.

## Verification

- Full `make TEST_EXPORTS=1 test`: exit 0, zero FAILs, "Skipped checks: none" (private ROM corpus symlinked)
- 12-title corpus sweep, scripted into gameplay, batch-aware seam metric: every title lands at **1.0x** (baseline: Karts 8.9x, Burnout ∞); RMS unchanged everywhere — no silencing-regression risk
- CD slave mode (Primal Rage, CDDA through BUTCH→I2S): seam ratio 0.99x, RMS 2422 identical, video windows byte-identical
- Skyhammer / IS1 / IS2 clipping+presence sentinels: all still pass (Skyhammer RMS 3083 vs documented 3080 baseline)
- Before/after listening clips (5 titles incl. the CD case) pending human sign-off — headless tests cannot tell music from structured noise

## Measured seam step (in-game, 3300 frames)

| title | before | after |
|---|---|---|
| Atari Karts | 697.6 (8.9x) | 93.4 (1.0x) |
| Super Burnout | 466.0 (staircase, no waveform) | 92.9 (0.9x) |
| Cybermorph | 326.2 | 279.4 (1.0x) |
| Wolfenstein 3D | 268.6 | 198.3 (1.0x) |
| 8 more titles | — | all 1.0x, RMS unchanged |

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -824,7 +824,7 @@ clean:
 		test/test_dsp_ops test/test_dsp_unit test/test_hle_bios \
 		test/test_subsystem_init test/test_subsystem_timeline \
 		test/test_irq_cascade test/test_boot_patterns test/test_audio_pipeline \
-		test/test_audio_clipping test/test_audio_presence test/test_audio_boundary test/test_pit_clock_rate \
+		test/test_audio_clipping test/test_audio_presence test/test_audio_boundary test/test_audio_rate test/test_pit_clock_rate \
 		test/test_blitter_mmio test/test_blitter_cmd test/test_eeprom_lifecycle \
 		test/test_eeprom_read_race \
 		test/test_tom_visible_window test/test_framebuffer_integrity \
@@ -970,6 +970,17 @@ test: test/test_dram_timing test/test_cheat test/test_event_queue test/test_jlin
 		./test/test_audio_boundary ./$(TARGET) "$$rom" --label "Atari Karts (boundary)" --quiet; \
 	else \
 		bash scripts/test-skip.sh record "Atari Karts (audio boundary)" "no ROM matching 'Atari Karts*' in the private corpus"; \
+	fi
+	@# Delivered-vs-advertised sample rate.  A steady deficit drains the
+	@# frontend's audio buffer until it underruns -- a pop every few
+	@# seconds in every title.  Orthogonal to the boundary check: the
+	@# pre-2026-08 core passed this and failed that; the first fix
+	@# inverted it.  Both must hold.
+	@rom=$$(bash scripts/find-rom.sh 'Atari Karts (1995).jag' 'Atari Karts*.jag' 'Atari Karts*.j64'); \
+	if [ -n "$$rom" ]; then \
+		./test/test_audio_rate ./$(TARGET) "$$rom" --label "Atari Karts (rate)" --quiet; \
+	else \
+		bash scripts/test-skip.sh record "Atari Karts (audio rate)" "no ROM matching 'Atari Karts*' in the private corpus"; \
 	fi
 	@# Formerly known-broken titles (DSP-synth saturation class): fixed by
 	@# the MMULT secondary-bank fix (JTRM: the vector operand is always
@@ -1338,6 +1349,10 @@ test/test_audio_presence: test/test_audio_presence.c
 test/test_audio_boundary: test/test_audio_boundary.c
 	$(CC) -O2 -Wall -std=c99 $(INCFLAGS) \
 		-o $@ test/test_audio_boundary.c -ldl -lm
+
+test/test_audio_rate: test/test_audio_rate.c
+	$(CC) -O2 -Wall -std=c99 $(INCFLAGS) \
+		-o $@ test/test_audio_rate.c -ldl -lm
 
 test/test_memtrack: test/test_memtrack.c src/core/memtrack.c src/core/memtrack.h
 	$(CC) -O2 -Wall -std=c99 $(INCFLAGS) \
--- a/Makefile
+++ b/Makefile
@@ -824,7 +824,7 @@ clean:
 		test/test_dsp_ops test/test_dsp_unit test/test_hle_bios \
 		test/test_subsystem_init test/test_subsystem_timeline \
 		test/test_irq_cascade test/test_boot_patterns test/test_audio_pipeline \
-		test/test_audio_clipping test/test_audio_presence test/test_pit_clock_rate \
+		test/test_audio_clipping test/test_audio_presence test/test_audio_boundary test/test_pit_clock_rate \
 		test/test_blitter_mmio test/test_blitter_cmd test/test_eeprom_lifecycle \
 		test/test_eeprom_read_race \
 		test/test_tom_visible_window test/test_framebuffer_integrity \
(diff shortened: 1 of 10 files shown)


JoeMatt:
@copilot review

JoeMatt:
## Follow-up: fixed an underrun-pop regression this PR introduced (5d27592)

Testing feedback was **pops in every title**, HLE and BIOS. That was mine, from the variable-length batch change in 2c99748.

Dropping the fabricated 800th sample removed the seam step, but it also dropped the frame's *fractional* remainder: `DACPrepareFrame` tore the sample clock down and restarted it from zero each frame. An NTSC field is 799.27 periods of the 48 kHz clock, so restarting truncated it to a flat **799 — never 800** — and the core delivered **107.9 samples/sec fewer than it advertised**. The frontend's audio buffer drained until it underran: a pop every few seconds, in everything.

Two halves, both needed:

1. **The sample clock free-runs.** `DSPSampleCallback` always reschedules (bounded only by the buffer), and `DACPrepareFrame` no longer removes/restarts it. Sub-period phase carries across the boundary, frames alternate 799/800, average **799.274** — exactly theoretical.
2. **`retro_get_system_av_info` reports the true field rate** (60.0544 NTSC / 49.9201 PAL) instead of a rounded 60/50. `fps` and `sample_rate` must describe the same machine: at a claimed 60 fps, even a perfect 799.274 samples/frame implies 47956 Hz.

Measured drift on Atari Karts, 900 frames: **-107.9 → -0.01 samples/sec**. Iron Soldier: -0.01. Seam stays at 1.00x interior.

### New test: `test_audio_rate`

Asserts samples-per-frame × advertised fps == advertised sample_rate within 5 samples/sec. It is **orthogonal** to `test_audio_boundary`, and the negative controls prove exactly that:

| core | rate check | boundary check |
|---|---|---|
| pre-2026-08 develop | PASS (flat 800) | **FAIL 10.05x** |
| 2c99748 (this PR, before) | **FAIL -60/sec** | PASS 1.00x |
| 5d27592 (now) | PASS -0.01/sec | PASS 1.00x |

Each earlier state traded one defect for the other. Only both together describe working audio — which is why the pops weren't caught before shipping the first two commits.

Full suite green, `Skipped checks: none`.

### Still open, tracked separately

Atari Karts remains gritty for a reason unrelated to this PR: its DSP mixer runs off **JERRY Timer 1 at 25038 Hz** (vector 2) while the serial port transmits at **20774 Hz** (the game writes SCLK=19 itself); the SSI vector is a no-op ack stub. That 1.205 ratio is exactly the 417-writes-vs-346-strobes measured. Both clock models check out against the JTRM (Technical Reference v8 p.85 for the PIT, p.91 for SCLK), so I'm not touching either without hardware ground truth.

JoeMatt:
## CI is green now — and the fix got better in the process (afdeb0d)

The three `regression-*` failures on 5d27592 were **real, not flaky**. I isolated the cause by md5-ing jagniccc's frame 500 across builds: develop and 2c99748 are byte-identical, 5d27592 differs. It was the free-running sample clock — since `JaguarExecuteNew` slices by time-to-next-event, no longer tearing the clock down and re-arming it each frame shifted 68K/GPU/DSP interleaving, and jagniccc's demo ended up two animation frames further along by frame 500. Nothing rendered wrong, but perturbing emulation timing to fix an audio bug is a bad trade.

**afdeb0d solves both problems without touching the event schedule:**

- The tail is padded by **holding** the last resampled pair instead of reading raw LTXD/RTXD. A DAC starved of new data keeps transmitting what it last received. This only works now that the resample cursors persist across the frame boundary (2c99748) — when I tried holding *before* that change, the cursor was re-anchored at the newest sample every frame, so it just moved the step from one frame's tail to the next frame's head. That's why it looked useless at the time and works now.
- The batch stays a fixed `length`, so 800 × 60 fps is exactly the advertised 48 kHz and the rate deficit is gone without renegotiating fps. The honest-fps change is reverted along with the free-running clock — whether to advertise the true 60.0544 / 49.9201 field rate is a **video**-timing question and gets its own PR.

### Verification

| check | result |
|---|---|
| jagniccc frame 500 | **byte-identical to baseline** (md5 6239018c…) — no baseline churn |
| `regression_test.sh` | 10 passed, 0 failed |
| `test_audio_rate` | 800.0000 samples/frame, drift **+0.00**/sec |
| `test_audio_boundary` | Karts 0.49x, Burnout 0.44x, Cybermorph 0.49x, Wolf3D 0.49x, IS1 0.51x |
| `make TEST_EXPORTS=1 test` | exit 0, `Skipped checks: none` |

Seam ratios below 1.0 are expected: a held pair has zero delta, so the seam now reads *smoother* than the surrounding signal.

Both review nits addressed in ea20968 (static-buffer rationale reworded; `v8_header` → `header_is_current_version`) and threads resolved.

### Deliberately out of scope, to follow as separate PRs

- **Advertising the true field rate** (60.0544 NTSC / 49.9201 PAL instead of 60/50) — a video-pacing change, not an audio one.
- **Atari Karts grit** — its DSP mixer runs off JERRY Timer 1 at 25038 Hz while the port transmits at 20774 Hz (the SSI vector is a no-op ack stub). Reported by the maintainer as a regression from ~2026-07-10, currently under bisect.
- **Residual dropouts in some titles** on device.

Issue #401, opened 2026-08-11
Render-bound game loops tick ~2x too fast (Doom, Hover Strike): GPU completes a frame in 1-2 fields where hardware needs 4-8


JoeMatt:
Root-cause characterization for the long-standing "Doom runs too fast" family, and the likely cause of the input-feel reports in #399 (one tap moving 2-3 menu items; Hover Strike firing/moving fast). Established entirely from in-tree measurement plus the open Jaguar Doom source — no hardware or BigPEmu reference needed.

## The mechanism

Jaguar Doom does not tick on VBL. `MiniLoop` ([`d_main.c`](https://github.com/JNechaevsky/jaguar-doom/blob/master/d_main.c)) ticks once per **completed render**:

```c
gamevbls += vblsinframe;
exit = ticker ();                             /* game + menu logic tick */
if (demoplayback || demorecording) vblsinframe = 4;
else vblsinframe = lasttics;                  /* VBLs the last frame took */
...
while (!I_RefreshCompleted ()) ;              /* wait on the renderer */
S_UpdateSounds (); drawer ();
while ( DSPRead(&dspfinished) != 0xdef6 ) ;   /* wait on the DSP */
```

So tick rate — and therefore menu auto-repeat, firing cadence, and movement speed — is a direct function of how fast our emulated renderer completes relative to VBL.

The game's own design range is explicit in that code: `vblsinframe` is clamped to a maximum of 8, the demo hard-codes **4**, and the disabled-but-telling original floor was also 4:

```c
#if 0
else if (vblsinframe < 4) vblsinframe = 4;
#endif
```

i.e. Doom expects **4-8 VBLs per tick (15-7.5 fps)**.

## Measurements (develop @ bae6aff, NTSC, `BENCH_PROFILE=1`)

Machine-level timing is **correct** — this is not a VBL-rate or field-length bug:

| metric | measured | expected |
|---|---|---|
| VBL IRQs per frame | 1.00 | 1 |
| halflines per frame | 524.0 | 524 (VP+1) |
| RISC cycles per frame | 442,701 | 442,780 (26.5909 MHz / 60.0544 Hz) |
| 68K cycles per frame | 221,435 | half of RISC |

But the render-bound loop runs fast:

[...]

Issue #399, opened 2026-08-11
Touchscreen taps trip in-game auto-repeat: one tap reads as 2-3 presses (Doom menu, Hover Strike)


JoeMatt:
Reported on iOS RetroArch: in Doom's start menu a single D-pad tap moves the cursor more than one item, and in Hover Strike gameplay the D-pad makes things fire/move faster than expected. Notably **not** fixed by forcing 60 fps or switching video backends, and only affects **some** games — which rules out frontend frame pacing.

## Measured: the emulation is faithful

Headless measurement on `develop` (`cd_visual_verify` + per-frame extraction of the menu skull's Y position from the captured PPMs), Doom start menu, varying how long `down` is held:

| hold length | menu items moved |
|---|---|
| 1 frame | 1 |
| 2 frames | 1 |
| 4 frames | 2 |
| 8 frames | 2 |

A sustained hold steps at frame intervals `6, 6, 3, 6, 6, 3, …`.

That is an exact match for Jaguar Doom's own menu ticker ([`m_main.c` `M_Ticker`](https://github.com/JNechaevsky/jaguar-doom/blob/master/m_main.c)):

```c
if (! (buttons & (JP_UP|JP_DOWN|JP_LEFT|JP_RIGHT) ) )
    movecount = 0;      /* move immediately on next press */
else
{
    if (cursorpos == level && movecount == 3)
        movecount = 0;  /* fast level select */
    if (movecount == 6)
        movecount = 0;  /* slower everything else */
    if (++movecount == 1)
    { ... cursorpos++ ... }
}
```

So: repeat fires every 6 ticks normally, but every 3 ticks while the cursor sits on `Area` (`cursorpos == level`) — which is exactly why a 4-frame hold already produces two moves, and why a sustained hold gives the 6/6/3 pattern. The core latches input once per `retro_run` (`update_input()`), `JoystickExec()` is an empty stub, and no double-latching exists.

## Actual cause

A finger tap on a touchscreen lasts roughly 100–200 ms = **6–12 emulated frames**, which is past Doom's 6-frame (~100 ms) auto-repeat threshold, so one tap legitimately reads as 2–3 presses. A physical D-pad tap is 1–3 frames and stays under it. "Only some games" follows directly: only titles whose auto-repeat threshold is short enough for a tap to cross it.

## Suggested direction

An **opt-in core option** rather than any change to emulation:

[...]

PR #403, opened 2026-08-11
test(acid): pin the display-gated loop rate that governs Doom's pace

JoeMatt:
Turns the "Doom feels too fast" class into a deterministic, reviewable ROM test — per the approach of writing synthetic ROMs that exercise the same code paths as the commercial title, rather than scripting the cart and reading its private counters.

## What it pins

Jaguar Doom's `I_Update()` ([`jagonly.c`](https://github.com/JNechaevsky/jaguar-doom/blob/master/jagonly.c)) ends every display with:

```c
do { junk = ticcount; } while (junk-lastticcount < 3);
lasttics     = ticcount - lastticcount;
lastticcount = ticcount;
```

`ticcount` is bumped once per VI in the `Frame:` handler ([`init.s`](https://github.com/JNechaevsky/jaguar-doom/blob/master/init.s), `addq.l #1,_ticcount`), with VI programmed to `a_vde|1` — once per field. `MiniLoop` ([`d_main.c`](https://github.com/JNechaevsky/jaguar-doom/blob/master/d_main.c)) then calls the game/menu ticker exactly once per completed display.

So **monster speed, firing cadence and menu auto-repeat are all bounded by this one gate** at `VI_rate / 3` = 20 iterations/second on NTSC. The ROM reproduces that exact path — VI → counter → 3-tick gate — over a fixed 60-field window and asserts 20 ± 1 gate passes.

Window is counted in VC wraps (same reference `vblank_60hz_exact.s` uses), so the assertion is immune to 68K speed and to the `dram_timing` model; it must pass with that option both enabled and disabled.

## Result on develop

**PASS** — exactly 20 gate passes, with `timing_vblank_irqs=61` over 61 frames.

That is a useful negative: our display-gate timing is *correct*, so this path is **not** the cause of the reported fast-pacing behaviour, and #401's investigation is redirected accordingly (the earlier 29.77 tics/s figure came from reading a commercial ROM's counters and does not correspond to MiniLoop iterations — arithmetically it cannot, since the gate caps them at 20/s).

Whole `tests/timing/` category is green with this added (14/14).

## Notes

- `scripts/lint-acid.py`: clean. Added to `BASELINE.txt` as `[PASS`, so a future regression is reported as a regression rather than a new test.
- No core changes.

Related: #401, #399.

### DIFF
--- a/test/acid/BASELINE.txt
+++ b/test/acid/BASELINE.txt
@@ -128,6 +128,7 @@
 [PASS tests/stress/deep_call_chain.jag
 [PASS tests/stress/many_blits.jag
 [PASS tests/stress/rapid_irq_pump.jag
+[PASS tests/timing/doom_update_gate_rate.jag
 [PASS tests/timing/halfline_count_per_frame.jag
 [PASS tests/timing/halfline_period_us.jag
 [PASS tests/timing/hc_advance.jag
--- a/test/acid/tests/timing/doom_update_gate_rate.s
+++ b/test/acid/tests/timing/doom_update_gate_rate.s
@@ -1,0 +1,137 @@
+;
+; tests/timing/doom_update_gate_rate.s - a display-gated game loop must
+; not run faster than one iteration per 3 video fields.
+;
+; This models Jaguar Doom's actual frame gate.  I_Update() (jagonly.c)
+; ends every display with:
+;
+;       do { junk = ticcount; } while (junk-lastticcount < 3);
+;       lasttics     = ticcount - lastticcount;
+;       lastticcount = ticcount;
+;
+; and `ticcount` is bumped once per VI in the Frame: handler (init.s,
+; `addq.l #1,_ticcount`), with VI programmed to a_vde|1 -- once per
+; field.  MiniLoop (d_main.c) then calls the game/menu ticker exactly
+; once per completed display, so the whole title's pace -- monster
+; speed, firing cadence, menu auto-repeat -- is bounded by this gate at
+; VI_rate / 3.  At the NTSC field rate that ceiling is 20 iterations
+; per second, and Doom cannot legally exceed it.
+;
+; Why a ROM and not a measurement on the commercial cart: the game-side
+; number is only reachable by scripting a real ROM into gameplay and
+; reading its private counters, which is neither deterministic nor
+; reviewable.  This reproduces the identical hardware path -- VI ->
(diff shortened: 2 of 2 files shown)


JoeMatt:
Added a second ROM to this branch that isolates the reported tap-repeat bug — a short D-pad tap registering as 2-3 presses (Doom menu jumping entries, Hover Strike accelerating too fast) unless the tap is extremely quick.

Two mechanisms can produce that and they need opposite fixes:
(a) **input over-sampling** — the core manufactures press events, so one continuous press reads as several edges;
(b) **loop rate** — input is clean, but the title's tick loop iterates more often per second than on hardware, so its own auto-repeat threshold is crossed sooner in wall-clock terms.

`input_edge_per_field.s` samples the pad in the VI handler using byte-for-byte the read Jaguar Doom performs (`init.s` `Frame:` — column $81FE, long read, $F0FFFFFC mask, `ror #4`), counts press edges and held fields, and publishes both to RAM. `test/tools/input_edge_test` holds one button for a known number of frames and asserts `edges == 1` and `held == N`.

**Result on develop — input is clean at every tap length:**

| hold (frames) | edges | held |
|---|---|---|
| 1 | 1 | 1 |
| 2 | 1 | 2 |
| 4 | 1 | 4 |
| 6 | 1 | 6 |
| 20 | 1 | 20 |

That rules out (a) definitively: the input path does not multiply presses. The remaining explanation is (b), and Doom's menu is exactly that shape — `M_Ticker` repeats on `movecount == 6`, while `M_Drawer` (unlike the gameplay path) never calls `I_Update()`, so the menu loop carries **no 3-tick gate at all** and free-runs at whatever rate the renderer allows. Combined with the gate ROM already on this branch — which proves the *gameplay* gate is correct at 20/s — the two together localise the fault to render-completion pacing on ungated loops.

Two rig traps this surfaced, both now guarded so a broken harness can't masquerade as a finding:
- `jaguarMainRAM` is a **pointer variable**, not the array — `dlsym` returns the pointer's address and it must be dereferenced. Reading it directly yields the pointer's own bytes and silently reports zeros.
- Synthetic ROMs require `virtualjaguar_bios=disabled`; the real BIOS authenticates the cart and a hand-built ROM never reaches its code.

The tool reports INCONCLUSIVE (exit 2) rather than a verdict whenever the ROM fails to publish or never observes the button.

Issue #411, opened 2026-08-12
AvP renders the full scene every engine cycle while idle — measurements + per-title optimization routes


JoeMatt:
## Summary

Alien vs Predator re-runs its **entire render pipeline at the full engine rate while the player is idle**, producing output the machine already has. Measured with the vjtrace flight recorder: an idle player costs exactly the same machine workload as a moving one (~290 blits/frame, ~42k GPU ops/frame), while the presented image cycles among just **21 distinct frames** over 1,601 idle frames (vs 306 distinct when moving). The redundant re-render is **15–18% of total host frame time** in the idle steady state, and AvP's shipped per-title defaults (2x + true color, #368) roughly double its absolute cost because every redundant blit also does Stage 2 shadow work.

This issue records the measurements and evaluates three per-title-optimization routes (IPS soft patches, per-title hooks, in-core heuristics), ending with a recommendation.

## Method

- Build: `feat/vjtrace` work @ `93c7f26`, `TEST_EXPORTS=1` (VJ_TRACE on), macOS arm64, host clang.
- Scenario driver: `test/fixtures/avp_reach_gameplay.press` reaches sustained Alien first-person gameplay.
  - **idle** run: presses truncated at frame 3000 (player stands still afterwards).
  - **moving** run: full fixture (keep-alive input to ~6300).
- Recorder: `trace_probe` `--field-csv` (per-field BLIT_CMD/GPU GO/OP counts + fb hash) + `--trace-out` ring (BLIT_CMD events carry B_CMD + A1_BASE), plus a throwaway harness tool sampling `gpu_exec_opcode_count` / `dsp_exec_opcode_count` and wall-clock per frame.
- Comparison window: frames 3100–4700 (1,601 frames) of a 4,800-frame run, both runs.

### 1. Engine structure: a 5-field render cycle

AvP's engine runs at ~12 Hz regardless of input. Per 5 fields (idle window, exact):

| frame class | frames | blits | GPU ops | what it is |
|---|---|---|---|---|
| heavy | 1 of 5 (320) | **1,446** | ~205k | full 3D view render + buffer flip |
| light | 1 of 5 (320) | 4 | ~7.5k | HUD touch-up |
| zero  | 3 of 5 (961) | 0 | ~0 | engine idle; 68K/DSP/OP only |

[...]

PR #426, opened 2026-08-13
fix(blitter): rate-limit the blit-memo arena-full notice (#411)

JoeMatt:
`BlitMemoFrame()` logged `[BLITMEMO] shadow arena full; flushing memo` at INFO on **every** arena-exhaustion flush. A title whose blit stream outruns the 400,000-record arena hits that on most frames, so the line reached the host log at frame rate.

Reported on **Alien vs Predator** with `virtualjaguar_blit_memo` enabled by hand. Note AvP is deliberately *not* tagged for the memo in `titledb.c` (the entry has an explicit comment saying it waits for the full corpus sweep plus a device check), so this only affects users who turn the option on themselves.

## Measured

AvP, 1800 frames scripted into gameplay, memo enabled:

| | before | after |
|---|---|---|
| arena-full flushes | 8 | 8 |
| log lines | **8** | **1** |

Sustained play is worse than the fixture — the flush rate rises with scene complexity, which is exactly when the log is least useful and most voluminous.

## Change

Announce the first occurrence (now naming the arena size, so a reader can tell *undersized arena* from *one pathological frame*), then only every 256th, carrying the running total so the frequency stays visible:

```
[BLITMEMO] shadow arena full (400000 records); flushing memo. Further occurrences are logged every 256th.
```

The counter resets when the memo is armed and at teardown, **never in `BlitMemoFlush`** — otherwise the "first" notice would fire on every flush and reintroduce the spam it replaces.

## Verification

- **Inert**: `frame_hash_ab` over 1800 frames with the memo enabled is **byte-identical** between `develop` and this commit. Logging only.
- AvP repro emits **1 line where it emitted 8**.
- `make TEST_EXPORTS=1 test` — **exit 0**.
- `bash scripts/c89-lint.sh src/tom/blit_memo.c` — passes.

## Deliberately not addressed

Whether the arena should be **larger**, or the eviction policy smarter, for AvP-class workloads. Flushing the whole memo several times a run means it rebuilds from scratch and may never pay for its write-hook cost on such titles. That is a real perf question for #411's follow-up, but it needs measurement rather than a bigger constant, and it does not belong in a log-hygiene fix.

### DIFF
--- a/src/tom/blit_memo.c
+++ b/src/tom/blit_memo.c
@@ -163,6 +163,12 @@ static int blitMemoCDNoticeLogged = 0;
 static bm_shadow_rec *bmShArena = NULL;
 static uint32_t bmShUsed = 0;
 static int bmShFull = 0;            /* reclaim at the next frame boundary */
+/* Arena-exhaustion flushes since the memo was armed.  A busy title can
+ * exhaust the arena on most frames, so the notice is rate-limited (see
+ * BlitMemoFrame) -- an unconditional line here reached the log at frame
+ * rate.  Reset only on arm/teardown, never in BlitMemoFlush, so the
+ * "first" notice stays first for the whole session. */
+static uint32_t bmShFullFlushes = 0;
 static bm_shadow_rec bmScratchSh[BM_SH_SCRATCH];  /* verify-mode capture */
 static uint32_t bmScratchShN = 0;
 
@@ -461,6 +467,7 @@ static int bm_alloc_shadow_arena(void)
    }
    bmShUsed = 0;
    bmShFull = 0;
+   bmShFullFlushes = 0;
    return 1;
 }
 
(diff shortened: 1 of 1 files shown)


JoeMatt:
Follow-up pushed (`d84a5e6`): the counter also needs resetting on the **mid-session re-arm** path.

`BlitMemoSetMode(OFF)` frees the pool but deliberately keeps the shadow arena, and `bm_alloc_shadow_arena()` is guarded on `!bmShArena` — so toggling the option off and back on ran neither of the two resets in the first commit. The counter carried over, and the user would get only the terse every-256th line, never the first-occurrence notice that explains the message and names the arena size. Now reset on the transition into an enabled/verify mode, beside the existing `[BLITMEMO] mode=` log (which runs exactly once per re-arm, since `BlitMemoSetMode` early-returns on an unchanged mode).

`make TEST_EXPORTS=1 test` exits 0; c89 lint passes.

Issue #370, opened 2026-08-09
[feat] Per-title enhancement hooks (BigPEmu-style gated patches)


JoeMatt:
Per-title patch/hook surface (#338 track 4b): small, gated, per-game interventions in the spirit of BigPEmu's script hooks — e.g. widescreen enables for titles that tolerate it, targeted overlay fixes.

- Off by default; every hook rides the per-title DB (#368) so it only ever applies to the title it was written for
- Modern-output extras from the epic (frame pacing / VRR friendliness, optional per-title widescreen) land here as the first consumers
- Scope guard: this is data-driven per-title patching, not a general scripting language, unless a real pack of use-cases demands more

Last item of epic #338 — closing this closes the epic and the project is feature-complete.

JoeMatt:
## BigPEmu's scripting system — what we can and cannot borrow

Investigated after the #401 finding that BigPEmu ships a **DOOM script with throttling** (v1.093) rather than modelling Doom's speed. That makes their approach directly relevant to this issue.

### What it actually is

- Scripts are **C, compiled to bytecode and run in a VM** inside the emulator. Files ship as `.c` source plus a compiled `.bigpcvm`.
- The API can set breakpoints, read/write emulated memory, and hook the machine — the same mechanism behind their achievements, TAS, stereoscopic and native-resolution-rendering features.
- Shipped uses include Doom (throttling + music options), Tempest 2000 (patching hardware rotary-controller bugs), and Checkered Flag ("Uncap Framerate").

### Do NOT copy the scripts

The main community collection ([jguffer/BigPEmu_Scripts](https://github.com/jguffer/BigPEmu_Scripts)) has **no license file** — all rights reserved by default. Vendoring any of it into a GPLv3 core is not an option, and neither is machine-translating it into our format, which would be a derivative work. The same caution applies to scripts shipped with the emulator itself.

**What is safely usable is the *fact* of what a title needs** — that Doom wants throttling, that Tempest 2000's rotary handling is buggy on hardware. Facts about game behaviour aren't copyrightable, and we'd be deriving our own fix from our own measurements anyway.

### What their format tells us about scope

Adopting a C-to-bytecode VM would be an enormous amount of machinery for what we currently need. Our per-title DB (#368) already delivers *option presets* keyed by CRC, including romhack aliases — that covers the "apply known-safe settings automatically" half of what BigPEmu's per-game profiles do.

[...]

Issue #408, opened 2026-08-11
Timing accuracy: unified plan — menu 2x root cause, then one correct system with no toggles


JoeMatt:
> **This is the plan of record for epic #319** (milestone [v4.0.0](https://github.com/libretro/virtualjaguar-libretro/milestone/6)). Open sub-issues: #313, #401, #331, #399. Retired in v3.4.0: #406, #392, and #434 — the hardware reference measurement, which landed at **4.0 fields/flip (15.01 flips/s)** for the Doom attract demo against our 2.00.
>
> **2026-08-13 — hardware capture received.** A user supplied a real-hardware Doom capture: the demo runs **much slower on silicon** than either VJ or BigPEmu, which confirms #401 from the source we always lacked, and the menu fast-input issue **exists on hardware but is milder**, making #399 downstream of #401 rather than independent. BigPEmu does not model this either — it ships a per-title *script* to throttle Doom (changelog v1.093), so "BigPEmu gets it right" is not a valid benchmark, and a per-title throttle via the title DB (#368) is a legitimate fallback if the general fix does not land. Counting frames in that capture is #434.

Tracking issue for the timing-accuracy work. Written so a fresh session can pick it up without re-deriving anything. Supersedes the framing in #401.

## End goal (from the maintainer)

One accurate timing system, **no user toggles** — every game paced correctly out of the box, with overclock and upscaling as separate *enhancement* levers that don't change game logic speed. The current option sprawl (`blitter_timing`, `dram_timing`, `gpu_pipeline_timing`, `m68k_clock_scale`, `risc_clock_scale`, env `VJ_DRAM_SCALE`) is a symptom of treating symptoms. Consolidation comes **after** the real mechanisms are known, not before — otherwise we freeze a taxonomy built on a wrong model.

## Targets, in priority order

1. **Doom menu tap-repeat** (#399) — one tap moves 2 items. The reported, reproducible, user-facing bug.
2. **Hover Strike gameplay** — ship moves far too fast under D-pad.
3. **Doom attract demo** — runs ~2x hardware pace.
4. **Battle Morph** — gameplay/enemy speed and framerate shift across the options matrix; unclear what is correct.

---

## ESTABLISHED FACTS — measured, do not re-derive

**Doom menu (the actual bug):**
- Cursor variable: **`$05170C`**. Baseline **2.00 items/tap** at hold=4, **1.25** at hold=2.
[...]

Issue #378, opened 2026-08-10
[feat] Measured overclock presets for the per-title DB (framerate-limited titles)


JoeMatt:
Follow-up to #368/#377: extend the per-title defaults DB with M68K/RISC clock-scale presets for framerate-limited titles (Doom class — and plenty more across cart+CD). Unlike true-color/2x (output-only, safe by construction), overclock changes what the emulated software computes per frame, so every entry needs measurement, not census presence.

## Methodology (two gates per candidate)

1. **Mechanical gate — does framerate actually improve?** Per title × scale grid ({1x, 1.5x, 2x} RISC × {1x, 1.5x} 68K): measure the *internal* frame rate (distinct-framebuffer-hash rate per emulated second, the `cart_boot_probe` motion metric) plus `test/tools/test_frame_timing` wall-speed ratio and anomaly output. A preset qualifies only if distinct-frame rate rises materially (e.g. ≥20%) at full host speed. Audio pitch is already contract-proven at every scale (v3.1.0 suite).
2. **Human gate — does gameplay pacing survive?** For mechanical winners only: play-test in RetroArch — game speed (not just render rate) must feel stock; Doom's known cross-cycle demo desync class (#337 notes) means demos are not valid evidence either way. Only titles passing both gates get DB entries, citing the measurement log.

## Notes
- Clock-scale option text says bug reports are only valid at 1x — DB presets must not change that: the crash/bug-report story needs a "disable per-title defaults" first step in the issue template.
- CD titles need the DB's CD keying first (deferred from #368 v1) — cart titles only until then.
- Candidate list to seed the grid: Doom, Cybermorph, Club Drive, I-War, Iron Soldier, Missile Command 3D, Super Burnout, Val d'Isere.
- ~~Checkered Flag~~ **removed — measured and failed gate 1, and this is the wrong mechanism for it** (see the gate-1 comment below). Presented rate is pinned at ~59/s across the whole grid (0.6% spread) for up to +56% instructions, and the limiter at `$43B0` read 0 (60 Hz) throughout. **A clock scale cannot lift a software cap** — even in a budget-1/2 state (30/20 Hz) the rate is a per-frame decision in the game's own code, not a consequence of render time. Its lever is the decoded byte patch at ROM `$8026EE` via #370 enhancement hooks; tracked there instead.
[...]

Issue #369, opened 2026-08-09
[feat] Texture replacement packs: blit-source dump mode + replacement pipeline


JoeMatt:
Classic HD-pack flow at the blitter (#338 track 2): hash blit source data → replacement lookup at blit time.

Two deliverables, in order:
1. **Dump mode** — write out hashed source tiles so pack authors have something to redraw (standalone value: also a debugging window into what titles actually blit)
2. **Replacement pipeline** — swap hashed sources for pack assets at blit time; >1x replacements ride the Stage 2 shadow surface (#367), 1x recolor packs work standalone

Guardrails per the epic: default off, stock path bit-identical, pace-neutral (replacement lookup must not add bus-model time).

JoeMatt:
## Deliverable 2 tier status + design notes for the remaining tiers

PR #488 lands the replacement pipeline **tier 1: 16bpp source tiles, 1x, straight-copy blits** — proven by `test/tools/test_texreplace.c` (machine bit-identical with a pack active, shadow surface provably carries pack RGB, deterministic) and verified on real hardware content: a solid-magenta pack over Tempest 2000's 395 dumped 16bpp tiles presents in 894/900 frames headless, zero with replacement off.

The architecture that shipped differs from the blit-time-substitution model these tiers were originally scoped against, in a way that changes what tiers 2 and 3 need. Recording it here so the next session doesn't re-derive it.

### Shipped architecture (tier 1) — host-only, presentation-riding

The pipeline never writes emulated state. At launch the source window is hashed (shared `TexDump*` helpers — one implementation of the frozen contract), and the *destination* window is described pre-dispatch (the engines write back pointer registers as they run — reading them post-dispatch describes the blit's END). Post-dispatch, every destination word that now equals its captured source word (**the per-pixel straight-copy witness**) gets pack RGB888 stored into the true-color shadow framebuffer via the new `ShadowFBStoreRGB`, value-tagged. The OP's existing shadow presentation path does the rest. The witness makes wrong models harmless: transparent/shaded/exotic pixels simply never store (it even caught the test battery's own LFU bug — a NOT-copy stored zero pixels).

[...]

PR #447, opened 2026-08-14
feat(titledb): per-title enhancement hooks (#370) - gated byte patches, off by default

JoeMatt:
## What this is

The byte-patch half of the per-title surface (#370, track 4b of epic #338).
`src/core/titledb.c` already carries a CRC32-keyed table of `{key, value}`
core-option presets; this adds the thing a string pair cannot carry — a
**verified write into emulated memory**, applied once, at cartridge load.

**Scope discriminator, enforced rather than described:**

> If an intervention can be expressed as a core-option value it is a
> `TitleDBPair` and it ships today. `hooks[]` exists **only** for writes to
> emulated memory.

And the second rule: a hook is a workaround for a *game* defect. It is never
a substitute for fixing an emulator bug — a proposal whose root cause is our
own inaccuracy is re-filed against the accuracy track (#319 / #408), not
patched per-title. Both rules are written into the table-policy comment and
`docs/enhancement-hooks.md`.

### The record

```c
typedef struct {
   uint8_t        kind;     /* TITLEDB_HOOK_ROM_PATCH; 0 terminates      */
   uint8_t        len;      /* 1..64                                     */
   uint32_t       offset;   /* PAYLOAD-relative; bus addr = $800000+off  */
   const uint8_t *expect;   /* MANDATORY precondition, len bytes         */
   const uint8_t *patch;    /* replacement, len bytes                    */
   const char    *name;     /* stable short id for the log line          */
} TitleDBHook;
```

`hooks[4]` sits beside `pairs[4]`. All 24 existing rows are byte-for-byte
unedited — C89 6.5.7 zero-fills an omitted trailing member of a static
aggregate, i.e. `kind == TITLEDB_HOOK_NONE`.

- **Offsets are payload-relative.** `DetectPrependedHeaderSize()` strips a
  512-byte header before *both* the CRC and the `memcpy` into the cart
  window, so a hex-editor offset taken from a headered dump is silently
  wrong by 512 while the CRC — the table key — is identical for both dumps.
- **`expect[]` is mandatory, and that is the load-bearing decision.** The
  usual "IPS has no source checksum" argument applies, but the decisive one
  is local: **titledb already inherits across CRCs.** The nine Doom EX rows
  deliberately reuse retail Doom's `pairs[]`. That is safe for settings
  (same engine, same rendering) and unsafe for bytes, because offset `$X` in
  a romhack is exactly the region the hack may have rewritten. `expect[]`
  enforces "hooks never inherit by alias" mechanically, instead of by a
  comment someone will eventually copy-paste past. On mismatch: refuse and
  log, never write. **All-or-nothing per entry** — every hook is validated
  first, and one failure means zero bytes are written.
- **One trigger: cartridge ROM at load.** No RAM patches, no per-frame
  enforcement, no watchpoints. Cart ROM is outside the state blob and
  `JaguarReset()` never touches it, so a patch survives `retro_reset()` and
  a serialize/unserialize round trip with **no re-apply, no new state field
  and no version bump** (both asserted, not assumed — see Test C).
- **One gate:** `virtualjaguar_enhancement_hooks`, default `disabled`, read
  **raw** via `environ_cb` so a DB row cannot carry
  `{"virtualjaguar_enhancement_hooks","enabled"}` in its own `pairs[]` and
  defeat off-by-default for its own title. `test_titledb` **fails the build**
  on any row that names the gate key, so the rule is enforced at `make test`
  time and not only by discipline. Latched once at load; `check_variables()`
  only compares-and-logs "takes effect on restart" and never re-latches.
- **Compiled-in C table.** No external file format, no parser, no
  system-dir search, no "which hooks file is this user running?" in bug
  reports. The user-authored path already ships twice — RetroArch soft
  patching (`docs/rom-patches.md`) and the cheat engine.

## The table ships ZERO rows, deliberately

This PR is **the mechanism, not the data**. A clean-room behaviour-research
task for the first candidate address is running separately; it had not
landed, and **no address was invented to fill the gap**. A fully-tested
mechanism with zero rows is a legitimate, releasable state: everything is
gated off, every fence is covered by a CI test, and the first row is pure
data with no code change. `test_titledb`'s hook-invariant loop is currently
vacuous *by design* — it exists so the **first** row added is validated
before it can reach a user.

Nothing in this PR is derived from any third-party patch data. No file,
script or binary from another emulator was downloaded, decompiled, vendored,
transliterated or mechanically converted. The authoring checklist requires
that a row's behaviour be established by **our own** analysis (disassembly,
`vjtrace`, `m68k_pc_histogram`, `gpu_disasm_dump`, `trace_dump`) with the
comment citing which.

## The three hazards, and how each is fenced

These were found by reading the code, not by testing, and each would have
shipped as a silent wrong answer.

### 1. GameDrive content is served from a *different* image

`JaguarLoadFileInternal` copies the first <=6 MB flat into
`jagMemSpace + 0x800000` **and** hands the full image to `JGDLoadROM`, which
`malloc`s a **separate 16 MB copy** (`jgdROM`, `jaggd.c:255-264`). Bank
switching then serves reads out of `jgdROM`, not the flat window — so a hook
that patched only the flat window would be silently defeated the moment the
game switched banks.

**Fence:** refuse when `jgdActive` **or**
`jaguarROMSize > JGD_AUTO_THRESHOLD`, and say why in the log. Both halves
matter: `jgdActive` can be set for a *small* image when the mode is forced
to `JGD_MODE_ENABLED`, so a size check alone is not enough.
`JGDWriteROM8()` is the extension point if GD hooks are ever wanted.

*Spec correction:* the spec asked for a new two-line `JGDIsActive()`
accessor. That is unnecessary — `jgdActive` is already an `extern uint8_t`
in `jaggd.h` (and already covered by the `JGD*` export wildcards), so the
applier reads it directly rather than adding redundant API.

### 2. The cart entry vector is consumed *before* hooks run

`JaguarLoadFileInternal` reads the entry point out of ROM itself
(`GET32(jagMemSpace, 0x800404)`, `file.c:318-333`) into `jaguarRunAddress`,
and `JaguarReset()` latches that into RAM (`jaguar.c:1373`) — **both inside
the boot strategy, i.e. before our call site.** `retro_reset()` then reuses
the cached variable rather than re-reading ROM. A hook patching cart
`$400..$407` would therefore pass `expect[]`, write successfully, and do
nothing — permanently, across resets. This is the one failure mode
`expect[]` structurally cannot catch: a silent no-op.

**Fence:** explicit range check, refused at load with a message naming the
reason. Tested as a real *overlap* test rather than two endpoint
comparisons — patches straddling `$3FE..$401` and `$406..$409` are both
refused, while `$3FC..$3FF` and `$408..$40B` are both **accepted**, proving
the fence is exact and not over-broad.

### 3. `TitleDB*` does not match `TitleHookApplyROM`

`link-test.T:81` and `exports-test.list:78` carry only prefix wildcards.
`TitleDB*` / `_TitleDB*` cover `TitleDBHooks` and `TitleDBSetHooksForTest`
for free, but **not** `TitleHookApplyROM` — the applier lives in its own
translation unit under its own name prefix. Miss this and the end-to-end
test fails at `dlsym`, looking exactly like an applier bug.

**Fence:** real entries (`TitleHook*` / `_TitleHook*`) added to both lists,
with a comment saying why the existing wildcard is not enough.
`scripts/check-export-lists.py` keeps the two in sync.

### A fourth, found while implementing: `TEST_EXPORTS` is link-time only

The spec proposed guarding the test-only hook setter with
`#ifdef TEST_EXPORTS`. That macro **is never defined** — `TEST_EXPORTS=1`
selects a symbol-export script and defines only `-DVJ_TRACE`
(`Makefile:87-94`), so the `#ifdef` would have been false in every build and
the setter would not have existed at all. Adding `-DTEST_EXPORTS` instead
would make `titledb.o`'s *content* mode-dependent, which puts it in the
`VJTRACE_HOOKED_OBJS` delete-on-mode-transition list or it silently survives
as a stale object (the exact trap `Makefile:795-822` documents).

**Resolution:** compile the setter unconditionally and hide it at *link*
time. `link.T` / `exports.list` export only `retro_*`, so it is invisible in
production (verified below), and there is no build-mode hazard. The reason
is written into the header so the next reader does not re-derive it.

## Verification

Every command below was run on this branch. Host: macOS/arm64, Apple clang
via `DEVELOPER_DIR=/Library/Developer/CommandLineTools`.

### C89 lint — all touched `.c` files

```
$ bash scripts/c89-lint.sh libretro.c src/core/titlehook.c src/core/titledb.c
C89 lint passed

$ bash scripts/c89-lint.sh test/tools/test_titlehook.c test/tools/test_titledb.c
C89 lint passed

$ bash scripts/c89-lint.sh          # full default pass (libretro.c + src/**.c)
C89 lint: sse2 pass OK
C89 lint: neon pass OK
C89 lint passed
```

### Clean build

```
$ DEVELOPER_DIR=/Library/Developer/CommandLineTools make -j$(getconf _NPROCESSORS_ONLN)
...
cc -ovirtualjaguar_libretro.dylib ... ./src/core/titledb.o ./src/core/titlehook.o ...
  -Wl,-exported_symbols_list,exports.list
```

No warnings from the new files.

### `make TEST_EXPORTS=1 test` exits 0

```
$ DEVELOPER_DIR=/Library/Developer/CommandLineTools make TEST_EXPORTS=1 test
TEST_EXIT=0
```

The four new gates inside that run:

```
./test/test_titlehook
OK (0 failures)                      # 38 checks

./test/test_titledb
OK (0 failures)                      # + the new hook-table invariants

./test/tools/test_hook_gate ./virtualjaguar_libretro.dylib test/roms/yarc.j64 --case on --quiet
=== Results: 5 passed, 0 failed, 0 skipped ===
  PASS: [gate_on_patch_applied] cart ROM holds patch[] after load
  PASS: [gate_on_apply_logged] [hooks] apply line names the hook
  PASS: [patch_survives_retro_reset] still patched after retro_reset()
  PASS: [patch_survives_state_roundtrip] still patched after serialize/unserialize
  PASS: [apply_returns_hook_count] TitleHookApplyROM() returned 1 and wrote the bytes

./test/tools/test_hook_gate ./virtualjaguar_libretro.dylib test/roms/yarc.j64 --case off --quiet
=== Results: 2 passed, 0 failed, 0 skipped ===
  PASS: [gate_off_rom_untouched] cart ROM still holds the original bytes
  PASS: [gate_off_nothing_logged] no [hooks] line for the canary

./test/tools/test_hook_gate ./virtualjaguar_libretro.dylib test/roms/yarc.j64 --case mismatch --quiet
=== Results: 2 passed, 0 failed, 0 skipped ===
  PASS: [mismatch_rom_untouched] cart ROM unchanged after a failed precondition
  PASS: [mismatch_refusal_logged] [hooks] REFUSED line logged

bash test/tools/hook_identity_ab.sh ./virtualjaguar_libretro.dylib
hook_identity_ab: yarc (1800 frames/arm)
  PASS yarc: baseline is deterministic (base == base2)
  PASS yarc: gate=off is byte-identical to the default arm
  PASS yarc: gate=on is byte-identical to the default arm
hook_identity_ab: jagniccc (1800 frames/arm)
  PASS jagniccc: baseline is deterministic (base == base2)
  PASS jagniccc: gate=off is byte-identical to the default arm
  PASS jagniccc: gate=on is byte-identical to the default arm
hook_identity_ab: OK -- the enhancement-hooks option cannot move the picture
```

**Trap avoided:** the end-to-end test installs its hooks programmatically via
`TitleDBSetHooksForTest()`, **not** as a canary row in the shipped table.
`Makefile` runs `test_pertitle_db --case 5` against `test/roms/yarc.j64` as
the deliberate *non-DB control*, asserting that no CRC match happens — any
yarc row would break it. Case 5 still passes in the run above.

### Gate-off byte-identity vs a pristine `develop` build

The self-comparison in `hook_identity_ab.sh` proves the *option* is inert.
That is **not** the same claim as "byte-identical to develop", so this was
done separately against a core built from a **fresh clone** checked out at
`1aa9c43` — not a `git stash` cycle in this tree, which produces chimera
binaries `VJ_EXPECT_BUILD` cannot catch. One `frame_hash_ab` binary, two
cores, `VJ_EXPECT_BUILD` unset (one build id cannot validate two dylibs),
each dylib identified first:

```
$ strings ./virtualjaguar_libretro.dylib        | grep -E '^v[0-9]+\.'
v3.3.0 1aa9c43-dirty                     <- this branch
$ strings $BASE/virtualjaguar_libretro.dylib    | grep -E '^v[0-9]+\.'
v3.3.0 1aa9c43                           <- pristine clone @ develop

--- gate at its default: PR core vs develop@1aa9c43 ---
IDENTICAL  yarc.j64      rows=1801  md5=f67647cca2c7d7bb3595e10a0f9478a3
IDENTICAL  jagniccc.j64  rows=1801  md5=28b1f7e74eed2f2a8fccefc8636bdbbe
EXIT=0
```

1800 frames per arm, per-frame `frame,w,h,duped,nonblack,hash` CSVs,
`cmp`-identical. Epic #338's stock-path guardrail holds.

### The refusal path actually fires

Real applier output (`test/test_titlehook`), showing that a wrong `expect[]`
refuses and **writes nothing** — and that all-or-nothing is real, because in
the three-hook case the *first* hook does not land either:

```
[hooks] Case1:  applied clean-apply @ $800100 +4
[hooks] Case1:  applied 1/1
[hooks] Case2:  REFUSED bad-precondition @ +$000100 -- expected bytes not found (this is not the dump the hook was written for); no bytes written
[hooks] Case3:  REFUSED h2-bad @ +$000210 -- expected bytes not found (this is not the dump the hook was written for); no bytes written
[hooks] Case3:  all 3 hook(s) in this entry skipped (all-or-nothing)
[hooks] Case3b: applied h1 @ $800200 +1
[hooks] Case3b: applied h2 @ $800210 +1
[hooks] Case3b: applied h3 @ $800220 +1
[hooks] Case3b: applied 3/3
[hooks] Case4:  REFUSED past-end -- +$000FFE..+$001001 is outside the 4096-byte image; no bytes written
[hooks] Case4b: REFUSED wrap -- +$FFFFFFF0..+$FFFFFFF3 is outside the 4096-byte image; no bytes written
[hooks] Case5a: REFUSED len-zero -- len 0 out of range 1..64 (table bug); no bytes written
[hooks] Case5b: REFUSED len-big -- len 65 out of range 1..64 (table bug); no bytes written
[hooks] Case5f: REFUSED bad-kind -- unknown hook kind 99 (table bug); no bytes written
[hooks] Case7a: REFUSED all 1 hook(s) -- Jaguar GameDrive content is served from a separate banked image, so a patch to the flat cart window would be defeated by bank switching; no bytes written
[hooks] Case8a: REFUSED entry-vector -- +$000404..+$000407 overlaps the cart entry vector ($400..$407), which is consumed before hooks run, so the patch would be a permanent silent no-op; no bytes written
[hooks] Case8c: REFUSED straddle-low -- +$0003FE..+$000401 overlaps the cart entry vector ($400..$407), ...
[hooks] Case9:  applied just-below @ $8003FC +4
[hooks] Case9:  applied just-above @ $800408 +4
[hooks] Case9:  applied 2/2
```

  PASS: case2: ZERO bytes written on refusal
  PASS: case3: ZERO bytes written -- hook 1 did not land either
  PASS: case9: the vector itself was not touched

Every refusal case re-`memcmp`s the **whole** image against a pristine copy,
so a partial write cannot masquerade as a refusal. The refusal is logged at
`LOG_WRN` (success at `LOG_INF`), so it reaches a user's RetroArch log
without any env-var dance.

The same refusal path through the **real core** is the `--case mismatch` run
above: `PASS: [mismatch_rom_untouched] cart ROM unchanged after a failed
precondition`.

### Production ABI unchanged

```
$ DEVELOPER_DIR=/Library/Developer/CommandLineTools make -j10       # plain make
$ nm -gU virtualjaguar_libretro.dylib | grep -i hook
$ echo $?
1                        # no matches -- this is the pass
46                       # retro_* only, unchanged
```

And under the wide test ABI they are present, so the export-list edit is
doing real work:

```
$ make TEST_EXPORTS=1 && nm -gU virtualjaguar_libretro.dylib | grep -i title
_TitleDBContentCRC   _TitleDBHooks        _TitleDBOverride     _TitleDBSetCRC
_TitleDBSetContent   _TitleDBSetHooksForTest                   _TitleDBTable
_TitleDBTitleName    _TitleHookApplyROM   _TitleHookApplyToBuffer
_TitleHookGetEnabled _TitleHookSetEnabled
```

## Files

| File | Change |
|---|---|
| `src/core/titledb.h` | `TitleDBHook`, `TITLEDB_HOOK_*`, `hooks[4]`, `TitleDBHooks()`, `TitleDBSetHooksForTest()` |
| `src/core/titledb.c` | hook lookup + test override; authoring checklist in the table-policy comment. **No existing row edited.** |
| `src/core/titlehook.{c,h}` | **new.** `TitleHookApplyToBuffer()` (pure) + `TitleHookApplyROM()` (globals wrapper) + the gate latch |
| `libretro.c` | raw gate read, the single call site, compare-and-log restart notice, static resets on unload/deinit |
| `libretro_core_options.h` | `virtualjaguar_enhancement_hooks`, default `disabled` |
| `link-test.T`, `exports-test.list` | `TitleHook*` / `_TitleHook*` |
| `Makefile`, `Makefile.common` | build + run the four gates; `titlehook.c` in the core sources |
| `test/tools/test_titlehook.c` | **new** host unit test (38 checks) |
| `test/tools/test_hook_gate.c` | **new** end-to-end gate test |
| `test/tools/hook_identity_ab.sh` | **new** stock-path identity gate |
| `test/tools/test_titledb.c` | shipped-table hook invariants |
| `docs/enhancement-hooks.md` | **new** authoring guide |
| `docs/rom-patches.md`, `CLAUDE.md` | how hooks differ from soft patching; layout + option notes |

The applier is split into a pure function and a globals wrapper on purpose:
it lets the host unit test drive every fence against a `malloc`'d buffer
with a handful of stubs, instead of standing up `jagMemSpace` — and it keeps
`titledb.c` dependency-light so `test_titledb` still links with its two-file
line and one stub.

## Issue status — please read before closing the epic

`Closes #370`.

**This does NOT close epic #338.** The spec header and #370's own body both
say "closing this closes the epic"; that is **out of date**. #338 currently
shows 3/6 sub-issues complete, with **#369** (texture replacement packs) and
**#378** (measured overclock presets) still open alongside #370. The epic
should stay open until those land. Stating it here so the next reader does
not re-derive it from the same stale text.

**One sub-claim of #370 is deliberately left unaddressed:** "widescreen
enables for titles that tolerate it". The Jaguar has no viewport-scaling
stage a byte patch can flip — a wider picture needs the *game* to render a
wider viewport. If research ever finds a title with a settable viewport
width in ROM, that specific title becomes a hook; otherwise it is core work,
not a hook. This is written down in `docs/enhancement-hooks.md` rather than
glossed over. Likewise the two "modern-output extras" #370 mentions (frame
pacing / VRR, fast-forward-safe audio) are host-output and core-audio
concerns that write nothing to emulated memory, so they route to core
options and not to this array.

## Follow-ups (not in this PR)

- The first real hook row, as **data only**, once the clean-room research
  supplies a verified address. No code change needed.
- Optional, deferred: a hook-set fingerprint in the savestate header's
  currently-unused `reserved` word, log-only, if cross-session/netplay state
  skew ever generates complaints. With zero rows it would distinguish
  nothing today.


JoeMatt:
Merged `develop` (e6dbb11) — conflicts were in `Makefile` only, both of the "each side adds a test target" kind (this branch adds `test_hook_gate`, develop's #449 added the input-device tools). Union-resolved; `exports-test.list` and `link-test.T` auto-merged and `scripts/check-export-lists.py` passes, so neither list lost an entry.

Local verification on the merge commit: build clean, `make TEST_EXPORTS=1 test` exits 0.

One thing worth flagging about that green: the worktree initially had no `test/roms/private` symlink, so the suite exited 0 having **skipped 15 checks** — including the two that matter most for this PR:

- `Doom (savestate determinism, enhancement path)` — the row that exists specifically because the hi-res frame epoch escaped the state blob in #400;
- `Per-title defaults (AvP apply/disable/override)` — the titledb path this PR extends.

Re-ran both against the real corpus after linking it:

```
test_runahead_determinism (Doom):  4 passed, 0 failed
  video/audio replay identical, state byte-identical after rollback (2621440 bytes)
test_pertitle_db (AvP, cases 1/2/3/4/6):  9 passed, 0 failed
```

Flagging it because "exit 0 with the feature's own tests skipped" is the failure mode this repo keeps re-learning, and a reviewer reading only the CI badge would not see it.

Review status: neither bot is available — Copilot's credits are exhausted and Kimi is out of quota for the cycle (HTTP 403 `access_terminated_error`), which is why this PR has been sitting unreviewed rather than reviewed-and-clean. [#455](https://github.com/libretro/virtualjaguar-libretro/pull/455) makes that state visible on the PR in future. A hand review of the patch path is running now and I'll post its findings here.

JoeMatt:
## Hand review (both bots unavailable)

Reviewed the patch path against the priorities in `.github/copilot-instructions.md`. **No defect found in the feature's code** — one blocking test-claim issue, now fixed in 7098869.

### What was verified clean

- **Bounds are wrap-safe by construction.** `titlehook.c:113` uses `h->offset > romSize || (romSize - h->offset) < h->len` rather than `offset + len > romSize`; `test_titlehook` case4b exercises `offset = 0xFFFFFFF0`.
- **The maximum hooked address lands exactly at the cart window's end.** Accepted `romSize` ≤ `JGD_AUTO_THRESHOLD` (`0x5FFF00`) → highest touched byte `0x800000 + 0x5FFF00 = 0xDFFF00`, which is precisely `cdRAM`'s base. No spill into the CDROM overlay. `file.c:279-281` caps the flat copy at the same threshold, so `romSize` is never larger than the bytes actually loaded — short ROMs fall out of the same check.
- **The CRC cannot self-invalidate or chain into another title's row.** `jaguarMainROMCRC32` is deliberately not recomputed after the write, and `TitleHookApplyROM()` asserts `TitleDBContentCRC() == jaguarMainROMCRC32` *before* writing. Both sides compute the same header-normalized CRC over the same buffer, so the fence is meaningful rather than vacuously true.
- **One latch site, one apply site.** Latch is a raw `environ_cb` read, so a DB row cannot flip its own gate (and `test_titledb` enforces that no row names the key). Gate re-armed off in both `retro_unload_game` and `retro_deinit` — correct for the iOS no-`dlclose` constraint.
- **Nothing added to the state blob**: `[state] v12 payload 2396971 / 2621440 bytes (224469 free)`, no `STATE_VERSION` bump needed.
- **Gate sensitivity proven by sabotage, not by reading**: forcing `--case off` with `virtualjaguar_enhancement_hooks=enabled` turns those assertions **2/2 red** (`cart ROM was patched with the gate at its default!`).

### The finding, and the fix

`hook_identity_ab.sh`'s header claimed the `on` arm "proves the applier is inert on unlisted content rather than merely unreached." It proves the opposite side of exactly that distinction. `TitleHookApplyToBuffer()` returns at its first guard for every ROM the script can run — `hooks == NULL` for unlisted content, `hooks[0].kind == NONE` for every row the shipped table carries — so the applier's body is dead code in every configuration shipping today. **Replace that body with `memset(rom, 0, romSize)` and the script still prints OK on all eight arms.**

Second, the missing-ROM branch left `rc` untouched, so a run that skipped every ROM exited 0 printing OK — while the Makefile comment directly above the call asserted "none of them can skip." True in practice (the ROMs are committed) and enforced by nothing.

Both fixed, and proven in both directions rather than asserted:

```
VJ_HOOK_AB_ROMS=<nonexistent>  -> FAILED -- 0 ROMs compared, rc=1
normal run                     -> OK (2 ROM(s)), rc=0
```

The header, the Makefile comment and `docs/enhancement-hooks.md` now describe what the A/B actually catches — an option that perturbs rendering merely by being registered, which is a real hazard — and point at `test_titlehook`/`test_hook_gate` for the applier itself. No change to the applier, the gate, or any shipped behaviour.

### Noted, not blocking

- `patch_survives_retro_reset` and `patch_survives_state_roundtrip` pass regardless of the applier — they are forward guards on the two invariants the "no savestate story" design rests on, which is legitimate, but `enhancement-hooks.md` cited them as evidence about the applier. Covered by the doc change above.
- The savestate/netplay gate-mismatch hole is already documented in `enhancement-hooks.md:66-70` with its mitigation (a fingerprint in the state header's `reserved` word) deferred. Agreed with the deferral — with zero rows it distinguishes nothing.
- `hook_identity_ab.sh` costs ~3.5 min of `make test` at 1800 frames × 4 arms × 2 ROMs for a comparison that is identical by construction. The `base`-vs-`base2` determinism control is genuinely useful so I left the frame count alone, but it is worth revisiting separately.

Issue #392, opened 2026-08-11
Advertise the true field rate (60.0544 NTSC / 49.9201 PAL) instead of 60/50


JoeMatt:
Split out of #391, where it was reverted because it changed emulation timing.

`retro_get_system_av_info` reports `fps = 60` (NTSC) / `50` (PAL). The emulator actually runs a field of VP+1 halflines — 524 NTSC at 31.777778 us, 626 PAL at 32.0 us — which is **60.0544 Hz** and **49.9201 Hz**. So we under-report by 0.09% NTSC and over-report by 0.16% PAL.

Audio is unaffected now: #391 delivers a fixed 800 samples/frame, and 800 x 60 is exactly the advertised 48 kHz. This is purely a **video-pacing** question — what the frontend uses for vsync matching and dynamic rate control.

Worth deciding deliberately, because two things are entangled:

1. Whether 524 halflines is itself correct. Real NTSC is 262.5 lines/field = 525 halflines = 59.94 Hz. We use VP=523 -> 524, which is one halfline short and the reason we land on 60.0544 rather than 59.94. Fixing the field length and fixing the advertised rate are different changes with different blast radii.
2. Anything that alters the emulated field length will move the screenshot baselines and every timing-sensitive test, so it needs the same A/B treatment #391 got (md5 frame comparison against `test/baselines/`).

Suggested order: settle whether VP+1 should be 525, then make the advertised rate follow from whatever the emulated field actually is — rather than hard-coding either number.

Refs: JTRM Technical Reference v8 (video timing); `src/core/jaguar.c` `HalflineCallback`, `libretro.c` `retro_get_system_av_info`.

JoeMatt:
## Attempted, reverted, and it exposed the real shape of this

I implemented the change as scoped — derive `timing.fps` from the emulated field instead of hard-coding — and `test/test_audio_rate.c` failed it:

```
Advertised: 60.0544 fps, 48000 Hz
Delivered:  800.0000 samples/frame over 900 frames
Implied rate: 48043.56 Hz -> drift +43.56 samples/sec (limit 5.0)
FAIL: the core does not deliver its advertised sample rate.
```

[...]

Issue #330, opened 2026-08-06
[test] test_cd_synth_subq cannot distinguish BCD from binary Q-position encoding


JoeMatt:
Coverage gap found while verifying #325, recorded rather than papered over.

**The gap:** stubbing `SubQBCD()` to return its raw argument (emitting binary instead of BCD) leaves `test_cd_synth_subq` **fully green, 6/6**. Nothing in the suite would catch a core that skipped BCD encoding entirely.

**Why:** the test decodes через `from_bcd()` and compares decoded values, and BCD and binary agree for values below 10. The fixture seeks to `track_data_lba(2)` and collects within a few sectors, so every position field — relative MSF and absolute MSF — is single-digit. The CRC can't help either: it's computed over whatever bytes are present, so it stays self-consistent under either encoding.

**Attempted and rejected:** adding a nibble-validity assertion (both nibbles ≤ 9) to the existing frame-content test. It passes under the sabotage too, because no field ever reaches 10 at that seek target — a check that cannot fail is worse than no check, so it wasn't shipped.

**What would actually close it:** collect a frame at a position where some field is ≥ 10 — e.g. prime deeper into a track so the frames field lands in 10..74 — then assert the raw byte equals `ref_bcd(expected)` directly (not round-tripped through `from_bcd()`). Cleanest as a separate small test so the existing `rel <= 10` and `abs` bounds don't need loosening.

Low severity: the encoding is currently correct, and the invariants that gate the VLM unmute (frame validity, $1D0F residual, position advance, stop→$0000) are all genuinely pinned — stubbing the Q builder turns 5 of 6 red.

JoeMatt:
## Verified closed — the sabotage this issue describes now reddens

Fixed by `7de96cb` (PR #448), which added `raw_bytes_are_bcd_at_double_digit_position`. I re-ran the exact experiment from the issue body rather than taking the commit message's word for it.

Applied the stub the issue names — `SubQBCD()` returning its raw argument, i.e. binary where the Red Book requires BCD — rebuilt, and ran the test:

```
PASS  word_format_valid_bit_and_49_sample_pacing
PASS  frame_is_valid_adr1_position_with_1d0f_residual
PASS  pregap_frame_has_index_zero_and_audio_control
PASS  q_position_advances_one_sector_per_588_samples
PASS  stop_returns_subdata_to_zero
FAIL  raw_bytes_are_bcd_at_double_digit_position:728: ASEC byte $0A,
[...]

Issue #393, opened 2026-08-11
Audio dropouts remain in some titles after the 60 Hz crackle fix (#391)


JoeMatt:
Follow-up to #391, which fixed the 60 Hz seam crackle and the underrun-pop regression. On-device testing (iOS, and Android/Odin originally) reports **dropouts still present in some titles**, while others (Wolfenstein 3D) now sound clean.

## Ruled out

- **Geometry churn is not the cause.** The core only ever calls `RETRO_ENVIRONMENT_SET_GEOMETRY` (libretro.c, in `retro_run` on a resolution change), never `SET_SYSTEM_AV_INFO`. `SET_GEOMETRY` re-allocates the frontend texture but does not re-init the audio driver, so the `SET_GEOMETRY: 652x450 -> 652x480` transitions visible in user logs are not it.
- **Rate drift is not the cause any more.** `test_audio_rate` measures +0.00 samples/sec against the advertised 48 kHz.

## Leading hypothesis (untested)

`sampleBuffer` is `BUFMAX = 2048` uint16 = **1024 stereo pairs** (libretro.c:55). A normal NTSC frame produces 799. If any frame ever advances more emulated time than usual — a long event, a state load, frameskip interaction — `DSPSampleCallback` would run past the buffer and either overrun it or (with a guard) silently truncate. Either presents as a dropout.

Worth doing regardless of whether it is the cause: size the buffer from the worst-case frame rather than a round number, and assert rather than truncate.

## Reproduction notes

- `test/test_audio_boundary` and `test/test_audio_rate` (both added in #391) are green on the corpus, so whatever this is, neither currently catches it — a dropout is a *gap*, not a seam step or a rate deficit. A third detector may be needed (e.g. run-of-zeros or batch-count anomaly per frame).
- The probe used throughout #391 supports `--wav` plus a `.batches` sidecar recording per-batch sample counts, which is the natural place to look for a missing or short batch.

Titles: to be narrowed — reporter has it on device across several games.

PR #448, opened 2026-08-14
fix(timing): advertise the emulated field rate (#392); test(cd): pin Q BCD encoding (#330)

JoeMatt:
Closes #392
Closes #330

Two independent accuracy items, one commit each (disjoint file sets, disjoint
call graphs — either can be reviewed, reverted, or bisected alone).

---

## #392 — advertise the field rate the core actually paces

`retro_get_system_av_info` hard-coded `fps = 60` (NTSC) / `50` (PAL). The core
paces a field of `VP+1` halflines, which is **60.05445 Hz** / **50.08013 Hz** —
so the frontend matched vsync and ran dynamic rate control against a number the
core does not run at.

### The JTRM settles the field length — and the emulated one is already correct

Both quotes verified against the primary source in `docs/atari-jaguar-1999/`
(page numbers are the printed ones):

> **JTRM Revision 8, p.15** — `VP` (Vertical Period, `$F0003E`):
> *"This eleven bit register determines the number of half lines per field. The
> number is one more than the value written into this register. **If the number
> of half lines is odd then the display is interlaced.**"*

> **JTRM Revision 10, p.8** — "Video Timings":
>
> | | PAL | NTSC |
> |---|---|---|
> | Vertical lines (interlaced) | 625 | 525 |
> | **Vertical Lines (non interlaced)** | **624** | **524** |

Titles run non-interlaced, so `VP+1` is the even member — **524 NTSC / 624 PAL**,
which is exactly what `TOMReset` already programs (`VP = 523 / 623`). **So there
is no field-length change to make; this PR is advertise-only, and the screenshot
baselines must not move.**

525 halflines would *select interlace* by the Rev 8 parity rule, so the
issue's "settle whether VP+1 should be 525" is **refuted, not deferred** —
recorded here so it is not reopened. A real non-interlaced Jaguar NTSC field is
60.05 Hz; 59.94 Hz is the 262.5-line *interlaced* rate.

Corroboration from the same manuals: `HP = 844 / 850` gives
`845 / 26.590906 MHz = 31.77778 us` and `851 / 26.593900 MHz = 31.99972 us`,
matching Rev 10's 63.5555 us / 64.0 us line periods (Rev 8 p.14: `HP` is "the
period of half a display line ... one tick longer than the value written").

```
NTSC: 524 x 31.777777777 us = 16651.5556 us -> 60.05445 Hz
PAL:  624 x 32.0          us = 19968.0000 us -> 50.08013 Hz
```

### :warning: Issue #392's stated PAL figure is wrong

The issue says the PAL field is **"626 PAL at 32.0 us"** → **"49.9201 Hz"**, and
that PAL is *over*-reported by 0.16%. Both halves are wrong. The PAL field is
**624** halflines → **50.08013 Hz**, so PAL is *under*-reported, in the same
direction as NTSC. The same wrong `626` was baked into `DACPrepareFrame`'s clamp
fallback in `src/jerry/dac.c` and into its comment block — fixing that is this
PR's only behaviour-bearing audio hunk. A comment correcting the record is
posted on the issue.

### One source of truth

New in `src/core/jaguar.{h,c}`, consumed by `libretro.c` and `src/jerry/dac.c`,
so the advertised number cannot drift from the emulated field:

```c
double   JaguarGetHalflinePeriodUs(void);      /* 31.777777777 / 32.0 */
uint32_t JaguarGetDefaultFieldHalflines(void); /* 524 / 624          */
double   JaguarGetFieldRateHz(void);           /* 60.05445 / 50.08013 */
```

`jaguar.c`'s four inline `(vjs.hardwareTypeNTSC ? 31.777777777 : 32.0)`
ternaries (bus arbiter, blitter tick, both halfline `SetCallbackTime` calls) now
call the accessor — same literal, same `double`, bit-identical; the
byte-identical A/B below is the evidence rather than the assertion.

`JaguarGetFieldRateHz()` deliberately uses the *default* field length, not the
live `VP`: chasing runtime `VP` would mean firing `SET_SYSTEM_AV_INFO`
mid-session, tearing down the frontend's audio pipeline every time a title
reprograms TOM.

**Boundary of the "one source of truth" claim, stated honestly:**
`src/cd/jagcd_hle.c`'s `HLE_NTSC_HALFLINE_US 31.777777` (seven 7s) is left
alone — a 1.1e-8 relative difference, worth 0.004 B/s of a 352800 B/s stream,
and it is an independently-tuned CD-stream constant whose perturbation would
churn the CD boot corpus for no gain.

### :warning: The declared sample rate had to follow — a deviation from the plan

The plan for this PR was fps-only, keeping `sample_rate = 48000`. **That
combination is incoherent and fails a gating test in `make test`.**

`retro_run` hands the frontend a *fixed* `BUFNTSC`/`BUFPAL` batch — 800 / 960
stereo pairs — exactly once per field. So:

| advertised pair | 800 pairs x fps | vs declared | verdict |
|---|---|---|---|
| `(60, 48000)` — before | 48000.00 | 0 | self-consistent **lie** |
| `(60.05445, 48000)` — fps-only | 48043.56 | **+43.56/s** | **incoherent** |
| `(60.05445, 48043.56)` — this PR | 48043.56 | 0 | self-consistent and true |

`test/test_audio_rate.c` asserts `samples-per-frame x fps == sample_rate` within
**5.0 samples/sec**, and its header says so in as many words ("`retro_get_system_
av_info` claimed a rounded 60 fps against a true 60.0544 ... fps x
samples-per-frame must equal sample_rate"). The fps-only variant misses that
bound by 8.7x and would leave a frontend pacing at 60.05445 fps while consuming
at 48000 — the permanent audio-buffer drift that test exists to catch.
**`DRIFT_LIMIT` was not touched.**

`48000` remains the DAC's internal resample target (`DAC_AUDIO_RATE`) and
cancels out of the pitch math (the per-output phase step is
ring-samples-captured / output-pairs), so this stays an advertised-value change.
`SAMPLERATE` in `libretro.c` had exactly one consumer — the line being changed —
and is removed; `info->timing.fps` likewise had exactly one reference in the
whole tree.

**Deferred instead:** making the *batch* fractional (799.27 pairs/field NTSC) so
that `pairs x fps` lands on a round 48000. That is the third coherent end state,
it is a real audio-behaviour change, and it needs a listening test — so it gets
its own PR rather than riding along in a pacing change.

### :warning: One planned change was NOT made, deliberately

The design called for inverting `src/tom/tom.c`'s comment *"Bit 0 in VP is
interlace flag. 0 = interlace, 1 = non-interlaced"* as wrong. **It is not wrong.
It is correct, and inverting it would have introduced a real defect.**
`tomRam8[VP + 1]` is the *low byte of the 16-bit VP register*, not `VP+1`
halflines: VP odd (bit 0 = 1) → `VP+1` even → non-interlaced, exactly what the
comment and the adjacent branch labels say. The comment was extended with the
JTRM derivation instead, so the misreading is not repeated.

### Comment and doc corrections (no behaviour)

* `src/core/jaguar.c` — the `HalflineCallback` header claimed "525 for NTSC, 625
  for PAL"; the perf-counter comment claimed "expected 525 halflines/frame".
* `src/jerry/dac.c` — the `#393` comment block said "961.536 periods for PAL
  VP=625"; a 624-halfline field is **958.464** periods, and the PAL cursor
  therefore drifted the *opposite* way from NTSC (gaining 1.536 x ratio ring
  samples per frame, not falling behind). Same magnitude, corrected sign.
* `test/harness/timing_probe.{c,h}` — `PAL_HALFLINES` was **625**, the
  *interlaced* value, making the PAL M68K/RISC cycle expectations 0.16% high.
  Diagnostic-only (it gates nothing in `make test`), but wrong. NTSC's 524 was
  already right.
* `docs/jtrm-clocks-timing.md` — the vertical-lines row read `524 (262.5 per
  field) | 625 (312.5 per field)`: internally inconsistent (524 halflines is
  262.0 lines, not 262.5) **and** it gave PAL the interlaced value. Its
  `Source:` line pointed at `src/tom/tom.c`, i.e. the thing the JTRM is supposed
  to check. Replaced with the split interlaced/non-interlaced rows, both derived
  field rates, and the Rev 8 / Rev 10 citations.
* `docs/fmv-drift-notes.md` §5.2 and §6 — **RESOLVED** banner, original text left
  in place per that file's house style. Its blocker ("the JTRM PDFs have no text
  layer") applied only to the chunked `03/04 - *.pdf` scans; `Technical
  Reference v8.pdf` and `v10.pdf` in the same directory *do* carry one. 524 is
  confirmed correct, so §5.2's feared 0.19% field-clock error does not exist —
  its own hedge ("a real Jaguar NTSC field may genuinely be 60.05 Hz") was right.

---

## #330 — pin the Q-subcode BCD encoding

Stubbing `SubQBCD()` to return its raw argument left `test_cd_synth_subq`
**6/6 green**: the existing tests decode through `from_bcd()` and collect within
a few sectors of a low LBA, so every position field is single-digit — and BCD
and binary agree below 10. The CRC cannot help; `CDROMBuildSubQ()` computes it
over whatever bytes are present, so it stays self-consistent either way.

One new `TEST()` seeks 50 sectors into track 3 (absolute time `00:10:50`) and
compares **raw bytes against `ref_bcd()` directly**. `from_bcd()` appears
nowhere in it. Existing tests 1-6 are untouched, no bound is loosened, and the
nibble-validity check the issue records as attempted-and-rejected is not
re-proposed.

Two discriminators:

* **ASEC is exactly 10 by construction** → BCD `$10` vs binary `$0A`. Exact.
* **AFRAME and the track-relative FRAME land in [50, 70]**, whose BCD images
  (`$50..$70`) and binary images (`$32..$46`) are **disjoint**.

> The design sketch used offset 20 / slop 24 and argued disjointness from
> `ref_bcd` injectivity: *"a binary `v` matches only if `ref_bcd(v) == v`, i.e.
> `v < 10`"*. **That argument is invalid** — the byte only has to equal
> `ref_bcd(v')` for *some* admissible `v'`, and at 20/24 the two sets overlap at
> `$20..$29` (binary 32..41), so those checks could have passed under the
> sabotage. Offset 50 / slop 20 makes them genuinely disjoint, and a **runtime
> tightness guard** now asserts that disjointness instead of claiming it in a
> comment.

Expected values are derived from the seek target the test computes via
`p_track_info` / `track_data_lba`, never hardcoded, and a precondition guard
fails loudly if the fixture geometry ever stops reaching a double-digit field —
the exact hole #330 documents.

---

## Verification

macOS arm64, `DEVELOPER_DIR=/Library/Developer/CommandLineTools`, all runs
against the committed tree (build id `85a9051`, clean).

### C89 lint — every touched `.c`

```
$ bash scripts/c89-lint.sh src/core/jaguar.c src/jerry/dac.c libretro.c \
      src/tom/tom.c test/harness/timing_probe.c test/test_cd_synth_subq.c
C89 lint passed
```

### Build

```
$ make -j$(getconf _NPROCESSORS_ONLN)
```
Zero warnings, zero errors; `virtualjaguar_libretro.dylib` links against the
shipped slim ABI.

### Negative control — the advertised number actually moved

A probe that `dlopen`s the core, loads `test/roms/yarc.j64` and prints
`retro_get_system_av_info()`, run against **two separately built binaries**: a
detached worktree at `develop` (1aa9c43) and this branch. Never a `git stash`
cycle — that builds chimera binaries from stale objects that `VJ_EXPECT_BUILD`
cannot catch.

```
--- develop (reference) ---
NTSC: fps=60.000000 sample_rate=48000.0000  (pairs/field x fps = 48000.0000)
PAL : fps=50.000000 sample_rate=48000.0000  (pairs/field x fps = 48000.0000)
--- fix/392-true-field-rate ---
NTSC: fps=60.054449 sample_rate=48043.5595  (pairs/field x fps = 48043.5595)
PAL : fps=50.080128 sample_rate=48076.9231  (pairs/field x fps = 48076.9231)
```
`sample_rate` equals `pairs x fps` exactly on both regions after the change,
because both come from the same accessor.

### A/B inertness proof — byte-identical framebuffers

A proof that **nothing moved**, not a rebaseline. Two worktrees, two separately
built production-ABI binaries, the same committed baselines, the same pre-built
miniretro.

`test/regression_test.sh` — develop reference (1aa9c43):
```
   PASS: jagniccc (0 pixels differ)
   PASS: yarc (0 pixels differ)
   PASS: jagniccc determinism (identical across runs)
   PASS: yarc determinism (identical across runs)
   PASS: jagniccc frameskip invariance (skip=0 matches skip=3)
   PASS: yarc frameskip invariance (skip=0 matches skip=3)
   PASS: jagniccc save state round-trip (frame 400 matches)
   PASS: yarc save state round-trip (frame 400 matches)
   PASS: jagniccc rewind (resume from frame 200 matches)
   PASS: yarc rewind (resume from frame 200 matches)
==> Results: 10 passed, 0 failed, 0 new (no baseline)
```

`test/regression_test.sh` — this branch, identical:
```
   PASS: jagniccc (0 pixels differ)
   PASS: yarc (0 pixels differ)
   PASS: jagniccc determinism (identical across runs)
   PASS: yarc determinism (identical across runs)
   PASS: jagniccc frameskip invariance (skip=0 matches skip=3)
   PASS: yarc frameskip invariance (skip=0 matches skip=3)
   PASS: jagniccc save state round-trip (frame 400 matches)
   PASS: yarc save state round-trip (frame 400 matches)
   PASS: jagniccc rewind (resume from frame 200 matches)
   PASS: yarc rewind (resume from frame 200 matches)
==> Results: 10 passed, 0 failed, 0 new (no baseline)
```

Stronger check, every 20th frame of a 600-frame run md5'd on both sides:
```
  jagniccc.j64: 30 dumped frames, md5 sets IDENTICAL
  yarc.j64: 30 dumped frames, md5 sets IDENTICAL
```

**Zero baseline files modified** (`git status test/baselines` is empty).

### Full suite

```
$ make TEST_EXPORTS=1 test
SUITE_EXIT=0
```
The suite's own tail confirms nothing silently sat out:
```
```

Notably, `test_audio_rate` — the gate discussed above — passes with the
*coherent* pair:
```
=== Audio rate check: Atari Karts (rate) ===
  Advertised: 60.0544 fps, 48044 Hz
  Delivered:  800.0000 samples/frame over 900 frames (900 batches)
  Implied rate: 48043.56 Hz -> drift +0.00 samples/sec (limit 5.0)
  PASS: delivered rate matches the advertised rate
```

### Audio — both tests, mandatory (`dac.c` is touched)

Running only clipping masks the silencing-regression class (PR #170 took Iron
Soldier 2 to RMS=521/silent while clipping still passed), so both are run and
both are reported.

```

=== Clipping check: Skyhammer ===
  ROM:    test/roms/private/ROMS/Skyhammer (World).j64
  Frames: 300, Window: [60, 300)
  BIOS:   disabled (HLE)
  First audio at frame: 171
  Samples in window:    384000
  Saturated (|s|>=32760):  0 (0.000%)
  Longest saturation run: 0 samples
  Window RMS (combined): 3083.3
  Hot frames (RMS>20000): 0/129 (0.0%)
  PASS: no clipping signature in window

=== Clipping check: Iron Soldier 2 ===
  ROM:    test/roms/private/ROMS/Iron Soldier 2 (World).j64
  Frames: 300, Window: [60, 300)
  BIOS:   disabled (HLE)
  First audio at frame: 21
  Samples in window:    384000
  Saturated (|s|>=32760):  0 (0.000%)
  Longest saturation run: 0 samples
  Window RMS (combined): 2598.3
  Hot frames (RMS>20000): 0/240 (0.0%)
  PASS: no clipping signature in window
```

```

=== Presence check: Iron Soldier 1 ===
  ROM:    test/roms/private/ROMS/Iron Soldier (1994).jag
  Frames: 300, Window: [60, 300)
  BIOS:   disabled (HLE)
  First audio at frame: 12
  Active window frames: 240 / 240
  Window RMS:           1174.2  [floor 200.0, ceiling 25000.0]
  Longest zero run:     1888 stereo frames (~0% of window)
  PASS: audio is present and within expected envelope
```

Iron Soldier 1 is inside develop's measured envelope, and identically so —
the same probe against the **develop reference binary** in the same session:

```

=== Presence check: Iron Soldier 1 (develop reference) ===
  ROM:    test/roms/private/ROMS/Iron Soldier (1994).jag
  Frames: 300, Window: [60, 300)
  BIOS:   disabled (HLE)
  First audio at frame: 12
  Active window frames: 240 / 240
  Window RMS:           1174.2  [floor 200.0, ceiling 25000.0]
  Longest zero run:     1888 stereo frames (~0% of window)
  PASS: audio is present and within expected envelope
```

`Window RMS 1174.2` on both sides, `First audio at frame 12` on both sides.

**Scope note, stated plainly:** the `626 -> 624` fallback fix is **PAL-only and
clamp-only** — the clamp fires only when a title leaves `VP` outside
[200, 1200], and `TOMReset` installs 623 before the first PAL frame, so the
normal PAL path already reads 624 from the register. The NTSC audio tests do not
exercise it. What they *do* prove is that routing `dac.c`'s halfline period and
field default through the new accessors changed nothing.

### #393 regression — the field-derived DAC cadence

`i2s_lag_probe`, both regions, `--max-resyncs 0` (a nonzero resync count is the
direct signature of breaking the field-derived cadence):

```
harness: core Virtual Jaguar v3.3.0 85a9051 (./virtualjaguar_libretro.dylib)
# window=2400 frames=2400 rom=test/roms/private/ROMS/Atari Karts (1995).jag
  2400  lag=    1.68  resyncs=  0  writes/f=  346  dlag/f=+0.0007
PASS i2s_lag: peak lag 2.16, resyncs 0
harness: core Virtual Jaguar v3.3.0 85a9051 (./virtualjaguar_libretro.dylib)
# window=2400 frames=2400 rom=test/roms/private/ROMS/Atari Karts (1995).jag
  2400  lag=    1.41  resyncs=  0  writes/f=  415  dlag/f=+0.0006
PASS i2s_lag: peak lag 2.16, resyncs 0
```

### #330 sabotage legs

**Leg 1 — clean:**
```
  PASS  disarmed_subdata_reads_return_zero
  PASS  word_format_valid_bit_and_49_sample_pacing
  PASS  frame_is_valid_adr1_position_with_1d0f_residual
  PASS  pregap_frame_has_index_zero_and_audio_control
  PASS  q_position_advances_one_sector_per_588_samples
  PASS  stop_returns_subdata_to_zero
  PASS  raw_bytes_are_bcd_at_double_digit_position
--- BUTCH Q-subcode serial window (synthetic audio disc): 7 passed, 0 failed, 0 skipped ---
```

**Leg 2a — `SubQBCD()` → `return (uint8_t)v;`**, the sabotage #330 documents as
leaving the file 6/6 green:
```
  PASS  disarmed_subdata_reads_return_zero
  PASS  word_format_valid_bit_and_49_sample_pacing
  PASS  frame_is_valid_adr1_position_with_1d0f_residual
  PASS  pregap_frame_has_index_zero_and_audio_control
  PASS  q_position_advances_one_sector_per_588_samples
  PASS  stop_returns_subdata_to_zero
  FAIL  raw_bytes_are_bcd_at_double_digit_position:728: ASEC byte $0A, expected $10 (ref_bcd(10)); $0A is the BINARY encoding -- SubQBCD() is not encoding
--- BUTCH Q-subcode serial window (synthetic audio disc): 6 passed, 1 failed, 0 skipped ---
```
Exactly one failure, and it is the new test, on the ASEC byte.

**Leg 2b — the *different* sabotage: stub the whole Q builder.** The issue's
"turns 5 of 6 red" figure refers to this mutation, not the `SubQBCD` one.
Stubbing `CDROMBuildSubQ()` immediately after the position fetch with
`subQValid` forced true:
```
  PASS  disarmed_subdata_reads_return_zero
  PASS  word_format_valid_bit_and_49_sample_pacing
  FAIL  frame_is_valid_adr1_position_with_1d0f_residual:504: CRC residual over 12 bytes is $0000, the DSP requires $1D0F (CRC must be stored inverted)
  FAIL  pregap_frame_has_index_zero_and_audio_control:556: pregap frame CRC residual $0000 != $1D0F
  FAIL  q_position_advances_one_sector_per_588_samples:602: Q position did not advance (0 -> 0)
  PASS  stop_returns_subdata_to_zero
  FAIL  raw_bytes_are_bcd_at_double_digit_position:714: CRC residual $0000 != $1D0F
--- BUTCH Q-subcode serial window (synthetic audio disc): 3 passed, 4 failed, 0 skipped ---
```
3 pass / 4 fail rather than the issue's 5-of-6 — a different stub point gives a
different count, so this is reported as measured rather than as remembered.

**Leg 3 — reverted, rebuilt, re-run.** The build id printed by the harness
confirms the rebuild actually happened rather than being inferred:
```
core: Virtual Jaguar v3.3.0 85a9051 (./virtualjaguar_libretro.dylib)
  PASS  disarmed_subdata_reads_return_zero
  PASS  word_format_valid_bit_and_49_sample_pacing
  PASS  frame_is_valid_adr1_position_with_1d0f_residual
  PASS  pregap_frame_has_index_zero_and_audio_control
  PASS  q_position_advances_one_sector_per_588_samples
  PASS  stop_returns_subdata_to_zero
  PASS  raw_bytes_are_bcd_at_double_digit_position
--- BUTCH Q-subcode serial window (synthetic audio disc): 7 passed, 0 failed, 0 skipped ---
```

---

## Outstanding — cannot be done headlessly

CLAUDE.md requires in-RetroArch verification for audio changes, and this PR
moves the **advertised audio contract**, so it is not a formality:

- [ ] Load a cartridge title in RetroArch; confirm the log reports
      `fps = 60.054449 / sample_rate = 48043.56` (or the PAL pair
      `50.080128 / 48076.92`).
- [ ] ~2 minutes of play: audio clean, video pacing smooth, and RetroArch's
      audio-stats overlay showing no sustained buffer drift.

The ~0.09% / ~0.16% mismatch is *removed* from the advertised pair by this PR
rather than introduced, so buffer behaviour should improve or stay the same —
but it is frontend-visible and deserves a listen.


JoeMatt:
### Code review

Found 3 issues:

1. The `SoundCallback()` comment states a contract this PR changes: it says the fixed batch keeps the delivered rate "exactly on the advertised 48 kHz (800 x 60 fps)". After this PR the advertised values are 60.05445 fps and 48043.56 Hz, so both numbers are false. This is the sentence `afdeb0d` wrote to justify the fixed-length batch when it reverted the earlier honest-fps attempt, and it sits in a file this PR otherwise re-commented.

https://github.com/libretro/virtualjaguar-libretro/blob/85a9051215adae9da8ac5731e245919ef38d0b8c/src/jerry/dac.c#L379-L381

2. The comment block above `HalflineCallback` now contradicts itself. The PR added "i.e. 60.05445 Hz and 50.08013 Hz -- not 60/50" at line 1227, but left "Each field is created in 1/60s (NTSC) or 1/50s (PAL)" at line 1218 in the same contiguous block.

https://github.com/libretro/virtualjaguar-libretro/blob/85a9051215adae9da8ac5731e245919ef38d0b8c/src/core/jaguar.c#L1217-L1228

3. `docs/frontend-pacing.md` is the canonical statement of the av_info contract this PR changes, and was not updated. The table still advertises `timing.fps` 60/50 and `timing.sample_rate` 48000/48000, and the prose still says both "work out to exactly 48 000 sample-frames per emulated second". `test/test_frontend_pacing.c` cites this doc as its contract.

https://github.com/libretro/virtualjaguar-libretro/blob/85a9051215adae9da8ac5731e245919ef38d0b8c/docs/frontend-pacing.md#L12-L20

Two sub-threshold notes, not filed above but worth a look before merge:

- Deriving `sample_rate` from `fps` (`sample_rate = (BUFNTSC/2) * fps`) makes `test/test_audio_rate.c`'s drift check insensitive to exactly the fps-vs-buffer mismatch that produced the +43.56 samples/sec figure motivating this PR. `test/test_frontend_pacing.c` names that scenario ("moving fps to 59.94 without touching the buffer size") as the class it exists to catch, and uses the same ratio. An assertion computing `fps` from field geometry independently in the test would restore it.
- The PAL `626 -> JaguarGetDefaultFieldHalflines()` clamp is called out in the PR body as the only behaviour-bearing audio hunk, but it is PAL-only while both mandated audio tests use NTSC fixtures. The outstanding RetroArch listen should include a PAL title.

Verified as correct, for the record: the decision not to invert the `tom.c` interlace comment (`tomRam8[VP + 1]` is the register's low byte, VP=523 is odd, so the non-interlace branch is right), and that this does not repeat #391 -- `afdeb0d` reverted that over sample-clock event scheduling, which this PR does not touch.

Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with the thumbs-up. Otherwise, react with thumbs-down.</sub>

Issue #436, opened 2026-08-13
[feat] Rotary controller support (Tempest 2000)


JoeMatt:
Sub-issue of #428. Tempest 2000 shipped with support for a **rotary (spinner) controller** — a continuous-rotation encoder rather than a direction pad.

**Estimate:** M — same shape as the mouse (#429): an encoder synthesised onto the `$F14000` matrix, on one axis instead of two.

## Scope

- Accept a spinner source from libretro (`RETRO_DEVICE_ANALOG` axis or `RETRO_DEVICE_MOUSE` relative X, whichever the frontend convention favours) and advertise the device via `RETRO_ENVIRONMENT_SET_CONTROLLER_INFO`.
- Synthesise the encoder signal onto the joypad matrix that Tempest 2000 polls, at a rate a polling game can track.
- Reuse the device-selection scaffolding established by #429 — do not build a second mechanism.

## Prerequisites

- The pin/bit mapping spike (#433) covers the mouse; the rotary needs its own authoritative mapping before implementation. Same rule: **do not guess bit assignments.**
- Worth knowing before scoping: **BigPEmu ships a script specifically to patch rotary bugs that existed on real hardware.** So "matches hardware" and "plays correctly" may not be the same target here — decide which we are aiming at, and record the reasoning. We are not cloning their script (the community repo carries no license); the fact that the bugs exist is free to act on.

## Guardrail

With no rotary device selected, `$F14000` behaviour must be bit-identical to today.

JoeMatt:
Shipped on develop via https://github.com/libretro/virtualjaguar-libretro/pull/451 (Tempest rotary on both ports).

Closing the ticket. Epic #428 stays open on analog/driving (#437), lightgun research (#438), and per-axis tuning (#439).

Issue #433, opened 2026-08-13
[spike] Source the authoritative ST/Amiga mouse pin -> $F14000 bit mapping (blocks #429)


JoeMatt:
Blocker on #429. Before a line of mouse code is written we need the authoritative mapping from Atari ST / Amiga mouse connector pins to the bits a game reads back from JERRY's `$F14000` muxed matrix. **Guessing the bit assignment is not acceptable** — a plausible-but-wrong quadrature encoding will decode as motion in some games and as garbage in others, and the failure looks like a game bug.

**Estimate:** S — research, not code. Unbounded until a source is in hand; that is why this is `blocked`, not sized larger.

## What we need, precisely

1. Which controller-port pins the ST mouse's X/Y quadrature phases (XA/XB, YA/YB) and buttons land on.
2. How the port-2 adapter presents those to the Jaguar — which **row select** written to `$F14000` makes them readable, and which **column bits** carry them.
3. The same for the **Amiga** wiring, which is *not* compatible (Elansar makes the user pick: `*` = Atari, `#` = Amiga).
4. Whether the PS/2 adapter presents an identical signal to the ST one — if so, one implementation covers both.
5. Whether phase ordering is X-then-Y or interleaved, and any rate ceiling a polling game can keep up with.

## Sources, in order

- **Matthias Domin's Jaguar mouse connector diagram (mdgames.de)** — the reference we want. WebFetch failed with a **TLS error**; try a different client, `curl --tlsv1.2`, the Wayback Machine, or ask in the community.
- Adapter vendor documentation — [AtariAge ST/Amiga adapter](https://store.atariage.com/products/atari-stamiga-mouse-adapter-for-atari-jaguar), [PS/2 variant](https://store.atariage.com/products/ps2-mouse-adapter-for-atari-jaguar), [The Brewing Academy](https://thebrewingacademy.com/products/atari-st-mouse-adapter). Vendors often publish or will share a schematic.
- **Atari ST hardware docs** for the mouse side of the protocol (well documented; the ST mouse predates the Jaguar and its quadrature encoding is public).
- **jag_sim / MiSTer** — check whether either models the adapter. jag_sim netlists have already broken JTRM ties for us (#354) and outrank secondhand sources.
- The Jaguar TRM's JERRY joypad section for what the matrix can physically express.
- AtariAge forum threads, as corroboration only — not as the primary source.

## Done when

[...]

Issue #429, opened 2026-08-13
[feat] Atari ST / Amiga mouse support (JagDoomEX, Elansar, Defender of the Crown)


JoeMatt:
Sub-issue of #428, and the one with live demand: **JagDoomEX** — the Doom romhack currently getting a lot of attention — reportedly supports (or is adding) mouse control, and several existing titles already do.

## Hardware background

Mouse adapters plug into the Jaguar's **port 2** and present an Atari ST (or Amiga) mouse to the console. Commercially available today from [AtariAge](https://store.atariage.com/products/atari-stamiga-mouse-adapter-for-atari-jaguar) (ST/Amiga and a [PS/2 variant](https://store.atariage.com/products/ps2-mouse-adapter-for-atari-jaguar)) and [The Brewing Academy](https://thebrewingacademy.com/products/atari-st-mouse-adapter). ST-compatible trackballs work through the same adapter.

Known supporting titles: Defender of the Crown, Gridrunner (Jeff Minter Classics), Elansar, Star Wars Arcade (ST port). Elansar selects mouse type in-game (`*` = Atari, `#` = Amiga) — **the two wirings are not compatible**, so a type selector is part of the feature, not a nicety.

## How it works, and what we'd need to emulate

An ST mouse is a **quadrature** device: each axis emits two phase-shifted square waves. The adapter routes those onto the Jaguar controller port's direction lines, and the game polls the pad and decodes phase transitions into relative movement; mouse buttons appear as fire buttons.

We already model the substrate: `src/jerry/joystick.c` implements `$F14000` as a muxed active-low matrix (write a row select, read columns). So the work is:

1. Accept `RETRO_DEVICE_MOUSE` on port 2 (relative X/Y + buttons) and advertise it via `RETRO_ENVIRONMENT_SET_CONTROLLER_INFO`.
2. **Synthesise quadrature** from libretro's relative deltas: accumulate movement and emit correctly-ordered phase transitions on the direction bits at a bounded rate, so a polling game decodes smooth motion. The emission rate is the crux — too fast and games miss transitions, too slow and the pointer lags.
3. Support **both ST and Amiga wirings** (selectable option; some games ask which you have).
4. Buttons onto the appropriate pad bits.

## Verification plan

- A homebrew/commercial title with known mouse support as the fixture (Elansar is a good candidate — it has an explicit ST/Amiga selector, so both wirings get exercised).
[...]

Issue #428, opened 2026-08-13
Epic: input device expansion — mouse, rotary, analog/driving, lightgun


JoeMatt:
The core currently models one controller type: the standard Jaguar digital pad (plus optional RetroPad remapping and numpad-to-keyboard). Real Jaguar hardware — and BigPEmu — support considerably more, and several are relevant to homebrew shipping **now**.

## What exists on real hardware

- **Atari ST / Amiga mouse** via a controller-port adapter ([AtariAge store](https://store.atariage.com/products/atari-stamiga-mouse-adapter-for-atari-jaguar), [The Brewing Academy](https://thebrewingacademy.com/products/atari-st-mouse-adapter)), and a **PS/2 mouse** variant ([AtariAge](https://store.atariage.com/products/ps2-mouse-adapter-for-atari-jaguar)). Plugs into **port 2**; ST-compatible trackballs work through the same adapter. Games that support it let you pick mouse type (Elansar: `*` = Atari, `#` = Amiga — the wirings are not compatible). Supported by Defender of the Crown, Gridrunner (Jeff Minter Classics), Elansar, the Star Wars Arcade ST port — and reportedly planned/landed for **JagDoomEX**.
- **Rotary controller** (Tempest 2000). BigPEmu ships a script specifically to *patch rotary bugs that existed on hardware*.
- **Analog / driving controllers**, **ADC-Reg**, and a **head tracker** (Jaguar VR) — all offered as controller types in BigPEmu.

## What BigPEmu exposes, for comparison

Controller types: standard, rotary, analog, driving, ADC-Reg, Head Tracker. Plus touch, gyro, padwheel, and per-axis tuning: X/Y offsets, dead zones, exponent curves.

## Why this is tractable for us

libretro already provides the input plumbing: `RETRO_DEVICE_MOUSE`, `RETRO_DEVICE_LIGHTGUN`, `RETRO_DEVICE_ANALOG` and `RETRO_ENVIRONMENT_SET_CONTROLLER_INFO` are standard, and `libretro.c` already consumes `RETRO_DEVICE_ANALOG` for stick-to-dpad mapping. The work is on the *emulation* side: synthesising what each peripheral puts on JERRY's joypad matrix.

That matrix is already modelled — `src/jerry/joystick.c` implements `$F14000` as a muxed, active-low row/column matrix (write a row select, read the columns back). Every one of these peripherals is a matter of driving the right bits on that matrix in response to a libretro device, not new bus plumbing.

[...]

PR #449, opened 2026-08-14
feat(input): ST/Amiga mouse on port 2 (#429)

JoeMatt:
## What this is

PR 2 of 4 of the input-devices track (epic #428): the **ST/Amiga mouse** (#429).

Implements steps **2-6** of `docs/superpowers/specs/2026-08-14-input-devices-mouse-rotary-design.md` — the quadrature encoder, the device model, the libretro layer, the mouse test, and the savestate chunk. Step 1 (the guardrail) is #446. **Step 7 (the Tempest rotary, #436) is deliberately not here** — it is PR 3. Steps 8-9 (titledb auto-select + docs) are PR 4.

Background reading: `docs/jaguar-mouse-adapter-mapping.md` (spike #433), which is the sourced pin -> `$F14000` bit mapping this is built from. No bit assignment in this PR is guessed.

### Stacking on #446

This branches off `test/429-input-guardrail`, **not** `develop`, so the guardrail is merged and green before anything can perturb `$F14000`. Until #446 merges, the diff shown here **includes A1's commit** (`test/tools/joymatrix_identity.c`, the `Makefile` test wiring, and the first pass at both export lists) — that work is not being re-landed, it is just not in `develop` yet. Once #446 merges this rebases away to my one commit.

---

## The load-bearing fact: the adapter is row-blind

Neither the ST nor the Amiga adapter connects Jaguar connector **pins 1-4** — and those are exactly `J4`..`J7`, the port-2 row-select *outputs* (mapping doc section 7). The adapter cannot see the row select at all. All six port-2 input lines (`J12`, `J13`, `J14`, `J15`, `B2`, `B3`) carry the mouse's state **continuously and identically in rows 0, 1, 2 and 3**.

**A Jaguar pad is a matrix; a mouse adapter is not.** So the mouse is not expressible as `joypadNButtons[]` slots. Setting `joypad1Buttons[BUTTON_U/D/L/R]` would return the mouse in row 0 and **nothing** in rows 1-3, where real hardware returns the same bits — a divergence running the *opposite* way from the one it appears to fix. The device instead drives raw `$F14000` bits 12-15 and `$F14002` bits 2-3 through a **row-independent overlay**, applied after the existing matrix decode and after the `joysticksEnabled` early return.

Three consequences are reproduced rather than "fixed", each with a comment at the site saying so:

| Behaviour | Why it is correct |
|---|---|
| In rows 1-3 the direction bits read as **keypad digits** (`*`/7/4/1, 2/5/8/0, 3/6/9/`#`) | Those are the same four wires. A title doing a full four-row scan while the mouse moves sees phantom digits on hardware. |
| LMB reads as **A and B and C and Option** | `joystick.c` decodes `$F14002` bit 3 per-row; a real adapter holds it low in every row. |
| RMB perturbs the **C1/C2/C3 controller-type** field | `B2` is the type-ID bit in rows 1-3, so a title re-probing controller type while RMB is held misidentifies the device. |

`joysticksEnabled` is deliberately **not** touched (spec section 2.4): the overlay sits strictly after the enable check, which is what makes the guardrail true by construction rather than by measurement.

---

## The emission rule: arm on write, advance on the next read

**Not** "advance one state per offset-0 read." That naive rule has a latent motion-loss bug and it was rejected on purpose.

The binding constraint is the game's decoder, not the device. A **two**-state Gray jump between consecutive polls indexes the "undetermined" entry of Domin's 16-entry table and is **silently discarded** — direction is unrecoverable. A driver that writes the row select once and then reads twice per poll (one read for directions, one for buttons — a very common shape) would, under "one per read", advance the phase twice between decode steps and emit exactly that discarded jump. That reads as **jitter**, which is worse than lag.

So:

- `JoystickWriteWord(offset == 0)` -> `InputDevArm()` sets an armed flag.
- `JoystickReadWord(offset == 0)` -> `InputDevClock()`: if armed, advance **at most one** Gray state per axis and clear the flag.

Extra reads without an intervening write return the same phase, which a decoder reads as "no movement" — harmless — instead of a discarded diagonal. `test/tools/mouse_decode_test` drives all three poll shapes and asserts it:

```
mouse_st: poll shapes (the emission rule)
  ok   shape C (write + TWO reads) decodes as shape A: A=40 C=40, dropped A=0 C=0
  ok   shape B (write once, then read-only) is frozen by design: NET_X=0 (expect 0)
```

Shape B (write the row select once at init, then poll read-only forever) is **frozen by design and asserted as such**. No such driver is documented; the assertion exists as the detector that would justify adding the stall-breaker, so "frozen" is a recorded, tested property rather than a surprise for whoever does the rotary.

---

## The guardrail still matches, exactly

**The headline result.** #446's digests are asserted unchanged, constants untouched:

```
```

The third row is the second assertion spec section 8 deferred until the device layer existed. It re-runs the identical sweep with an ST mouse attached to port 2 and a draining encoder, and asserts the port-1 sub-digest is *still* `0xD0CD02E2`. The pair matters: the **full** digest moves to `0xE0E2711A`, so the mouse is provably **live**; the **port-1** digest does not move, so it is provably **isolated**. A pass on the port-1 row alone could just mean the mouse did nothing.

---

## Savestate: v12

`STATE_VERSION` was 11 (`src/core/state.h`), and v11 **shipped** in v3.3.0, so a bump is correct rather than a shared in-flight version.

A **trailing** 41-byte chunk, appended after the v11 hi-res epoch — the mid-layout `JoystickStateSave` chunk does not change size. Per port: the X and Y quadrature backlog, Q8 carry and Gray phase, plus the two button latches; then one global armed byte.

The phase **is** what the game reads at `$F14000`, and the backlog and carry determine every future read, so all of it is machine-visible. A rollback without it replays different motion — issue #400's exact failure class, where the enhancement path's epoch sat outside the blob.

Deliberately **not** saved: the device type and the sensitivity scale. Both are option-derived and constant for a session; restoring them from a state would let a stale state fight the user's current option.

### STATE_SIZE headroom — measured, not assumed

Spec section 7.3 flagged this as unverified. A one-shot `LOG_INF` in `retro_serialize` now reports it:

```
[state] v12 payload 2396971 / 2621440 bytes (224469 free)
```

**224,469 bytes free (8.6%)**, with the v12 chunk costing 41 of them.

Backward compatibility is asserted, not assumed — `test_state_compat` still passes v1/v2/v3 loads and now reports `serialize_writes_current_version ... version=12` / `future_state_rejected ... version 13`, and `mouse_decode_test` writes a state, patches its header down to 11 and asserts it still loads (the v3.3.0 release layout, i.e. what real users' saves carry).

---

## Wiring cases, and the one thing that stays unresolved

Three cases ship as option values, from a literal table so they cannot drift apart (mapping doc section 4d):

| Option value | Case | X pair (A, B) | Y pair (A, B) |
|---|---|---|---|
| `mouse_st` (ST adapter + ST mouse, **and PS/2**) | 1 | 13, 12 | 14, 15 |
| `mouse_amiga` (Amiga mouse in an ST-wired adapter) | 3 | 13, 15 | 12, 14 |
| `mouse_amiga_adapter` (dedicated Amiga adapter) | 2 | 13, 12 | 15, 14 |

`mouse_amiga` maps to **case 3** because the shipping AtariAge adapter is ST-wired and an in-game "Atari / Amiga" selector exists to compensate — that is the case such a selector picks between.

### UNVERIFIED: the case-2 Y sign

Case 2 lands the Amiga's `V` (Y phase A) on bit 15 and `VQ` (Y phase B) on bit 14, i.e. A/B swapped relative to case 1, which **inverts Y**. Domin nonetheless drives both with one software mode. Three explanations exist, none sourced: ST and Amiga mice differ in encoder orientation and the swap cancels it; the vendor wiki's `(YA)`/`(YB)` labels on the Amiga column are editorial rather than measured; or he simply tolerated an inverted Y.

**No document settles it and no fixture in this repo can** — case-2 hardware is rare and no available title is known to target it. Per the design spec this is implemented **literally per the wiring diagram**, with an `UNVERIFIED` block at the table entry saying exactly that and telling the next reader not to "correct" it without a source. No invert toggle: a knob for an unsourced sign would just let users paper over a wiring bug.

The test asserts the case-2 behaviour with wording that keeps this honest — *"asserts the table as implemented, not a measured hardware fact."*

---

## Verification

All commands run with `DEVELOPER_DIR=/Library/Developer/CommandLineTools`.

### C89 lint

```
$ bash scripts/c89-lint.sh src/jerry/quadrature.c src/jerry/inputdev.c \
      src/jerry/joystick.c src/jerry/jerry.c libretro.c \
      test/test_quadrature.c test/tools/mouse_decode_test.c \
      test/tools/joymatrix_identity.c
C89 lint passed
```

### Build

```
$ make -j$(getconf _NPROCESSORS_ONLN)
make exit=0    (0 warnings, 0 errors)
```

### Full suite

```
$ make TEST_EXPORTS=1 test
suite exit=0
0 FAIL lines
```

Run with the private ROM corpus linked in, so the audio clipping / presence, run-ahead determinism and CD rows all actually executed rather than reporting SKIP (skip ledger empty).

### Production ABI unchanged by a plain make

```
$ make -j$(getconf _NPROCESSORS_ONLN)
      46
$ nm -gU virtualjaguar_libretro.dylib | awk '{print $3}' | grep -v '^_retro_'
(no output)
```

Every exported symbol is `retro_*`. The `InputDev*` / `Quad*` additions to `exports-test.list` and `link-test.T` extend the **test** ABI only; `scripts/check-export-lists.py` reports `export lists in sync (73 symbols)`.

### The mouse test — synthetic deltas in, decoded direction + distance out

`test/tools/mouse_decode_test` implements a quadrature decoder **independently of `src/jerry/quadrature.c`** — it walks the register bits, converts (A,B) levels to a Gray index and takes the modulo-4 difference between consecutive samples, discarding two-state jumps the way Domin's table discards its "undetermined" entries. If the emulated device ever emits a two-state jump, `dropped` goes non-zero and the run fails.

Abridged (the same block runs for all three wiring cases):

```
mouse_st: direction and distance
  ok   +X 40 units over 40 polls -> NET_X=40 NET_Y=0 dropped=0
  ok   -X 40 units over 40 polls -> NET_X=-40 NET_Y=0 dropped=0
  ok   +Y 40 units over 40 polls -> NET_X=0 NET_Y=40 dropped=0
  ok   -Y 40 units over 40 polls -> NET_X=0 NET_Y=-40 dropped=0
  ok   diagonal 40 polls -> NET_X=40 NET_Y=-40 dropped=0
mouse_st: rate ceiling (lag, never overshoot, never a sign flip)
  ok   4 units/poll x 30 polls -> NET_X=30 (expect 30), dropped=0
  ok   -4 units/poll x 30 polls -> NET_X=-30 (expect -30), dropped=0
mouse_st: poll shapes (the emission rule)
  ok   shape C (write + TWO reads) decodes as shape A: A=40 C=40, dropped A=0 C=0
  ok   shape B (write once, then read-only) is frozen by design: NET_X=0 (expect 0)
mouse_st: row-blindness
  ok   $F14000 bits 12-15 read identically in rows 0-3 (value $2)
  ok   LMB asserts $F14002 bit 3 in every row (1111)
  ok   RMB asserts $F14002 bit 2 in every row (1111)
  ok   released buttons leave $F14002 bits 2 and 3 high in every row

[ ... identical blocks for mouse_amiga_adapter and mouse_amiga (case 3) ... ]

wiring cases are genuinely distinct
  ok   case-3 wiring read with a case-1 decoder does not decode cleanly (NET_X=0 NET_Y=0 dropped=0; the matching decoder gives 40/0/0)
  ok   case-2 wiring read with a case-1 decoder inverts Y only (NET_X=0 NET_Y=-40)
savestate round-trip mid-motion
  ok   retro_serialize succeeded mid-motion
  ok   retro_unserialize accepted the v12 state
  ok   post-restore replay matches (NET_X 25 vs 25, NET_Y 19 vs 19, dropped 0 vs 0)
  ok   a v11 state (the v3.3.0 release layout) still loads
state chunk
  ok   InputDevStateSave wrote 41 bytes; inputdev.h advertises INPUTDEV_STATE_SIZE = 41
port scope
  ok   InputDevSetType refuses a mouse on port 1 (vendor-documented port-2 device)
no device selected -> the overlay is inert
  ok   port 2 reports pad and nothing is attached
  ok   no motion or buttons reach $F14000/$F14002 (NET_X=0 NET_Y=0 LMB=0 RMB=0)

  PASS: [mouse_decode] 0 failure(s)
```

The savestate round-trip is a genuine rollback: it drives the encoder to a mid-cycle phase with a non-empty backlog **and** a non-zero Q8 carry (37.5% sensitivity), saves, runs 25 more polls, restores, and replays the identical input — a mismatch is what a missing accumulator would produce.

### Encoder unit test

`test/test_quadrature` compiles `src/jerry/quadrature.c` directly, no core, no `dlopen`:

```
=== quadrature encoder ===
Gray sequence, positive direction (A leads B)
  ok   8 advances walk 0,1,2,3,0,1,2,3 with the table's levels
  ok   backlog fully drained after 8 advances
  ok   an empty backlog does not advance
Gray sequence, negative direction (B leads A)
  ok   8 advances walk 3,2,1,0,3,2,1,0 with the table's levels
  ok   backlog fully drained after 8 advances
one state per advance (the rate policy)
  ok   200 saturated polls each move exactly 1 state (never a jump)
  ok   advances past an empty backlog hold the phase steady
backlog clamp
  ok   positive backlog clamps at +64
  ok   negative backlog clamps at -64
Q8 sensitivity carry
  ok   4 x (1 unit @ 25%) accumulates exactly 1 state
  ok   4 x (-1 unit @ 25%) accumulates exactly -1 state
  ok   3 units @ 200% is 6 states
  ok   1 unit @ 25% emits nothing yet
  ok   ...and the carried remainder completes the state
reset
  ok   QuadReset zeroes backlog, carry and phase
All quadrature checks passed
```

Both tests are wired into `make test`.

---

## What is NOT verified

**Not yet exercised in a real frontend.** Everything above drives `InputDevFeed()` directly. The open question is whether RetroArch returns non-zero for `input_state_cb(1, RETRO_DEVICE_MOUSE, 0, RETRO_DEVICE_ID_MOUSE_X)` when the mouse is selected via the **core option** rather than by the frontend assigning port 2's device — RetroArch has historically bound a mouse to port index 0, and several cores read mouse state from port 0 for that reason. If port index 1 yields zeros, the core-option path is inert and only the Controls-menu path works.

No speculative port-0 fallback has been added: that would be a guess about frontend behaviour, and a wrong guess would let a **port-1 pad drive port 2's lines** — a worse bug than the one it would paper over. This wants one experiment in RetroArch (set the option with port 2 left as JOYPAD, versus setting port 2 to "Atari ST / PS2 Mouse" in Controls) before the feature is called done.

Also still open, and belonging to later steps rather than this one:

- **Which case Elansar's `*`/`#` selector actually means.** Only running Elansar and watching the cursor can answer it (clean axis-aligned motion vs the coupled, near-motionless signature the case-discrimination test above shows). Expect case 3.
- **Whether JagDoomEX actually reads the mouse.** That is the precondition for the titledb row, and the row is PR 4's, not this one's — `auto` resolves to `pad` for every title today, so the mouse is strictly opt-in.

---

## What changed

| File | |
|---|---|
| `src/jerry/quadrature.{c,h}` | **new** — pure 2-bit Gray-code encoder, no emulator headers. Shared with the rotary when #436 lands. |
| `src/jerry/inputdev.{c,h}` | **new** — device model, the three-case wiring table, the overlays, the emission clock, the v12 chunk. |
| `src/jerry/joystick.c` | **+22 / -2** — `InputDevArm()` in `JoystickWriteWord`, the row-gated `InputDevClock()` in `JoystickReadWord`, and the two overlay calls (which replace the two bare `return data;` lines, hence the 2 removed). Behaviour on a pad-only build is unchanged: every entry point short-circuits on `inputdev_attach_mask`. Note `InputDevArm()` lives in `JoystickWriteWord`, which has no `joysticksEnabled` early return — an earlier revision of this row claimed all the additions sat behind one. |
| `src/jerry/jerry.c` | `InputDevInit()` / `InputDevReset()` alongside the existing `Joystick*` calls. |
| `libretro.c` | subclass IDs, `SET_CONTROLLER_INFO`, `retro_set_controller_port_device`, option reads, visibility, the `update_input` feed + pad suppression, `InputDevShutdown()` in `retro_deinit`, the v12 save/load. |
| `libretro_core_options.h` | `virtualjaguar_p2_device`, `virtualjaguar_mouse_sensitivity`. |
| `src/core/state.h` | `STATE_VERSION` 11 -> 12, `STATE_VERSION_INPUT_DEVICES`, stale v9-v11 prose corrected. |
| `test/test_quadrature.c`, `test/tools/mouse_decode_test.c` | **new** tests, wired into `make test`. |
| `test/tools/joymatrix_identity.c` | third and fourth assertions with a mouse attached: port-1 isolation, and `EXPECT_DIGEST_FULL_WITH_MOUSE = 0x5E1858EA` pinning the mouse's own bits (the only assertion that catches an inverted active-low polarity). `EXPECT_DIGEST_FULL` and `EXPECT_DIGEST_PORT1` untouched and re-measured identical. |
| `exports-test.list`, `link-test.T` | `InputDev*` / `Quad*` — test ABI only. |

`InputDevShutdown()` is called from `retro_deinit` and clears every static including the option-derived ones, per the standing iOS rule (no `dlclose`, so statics persist across loads).

### Pad suppression

`update_input()` withholds the retropad from a mouse port entirely — physically there is no pad there, and leaving the host contribution would let a gamepad and a mouse drive the same six lines at once. There are **three** fill paths (the plain bitmask branch, the `enable_alt_inputs` remap branch including its analog-as-button cases, and the `numpad_to_kb` sub-path that writes slots 4-15 straight from `RETRO_DEVICE_KEYBOARD`); the suppression runs after all of them so none can be missed. A comment notes that the rotary will need this to become *selective* (buttons kept, only U/D/L/R withheld) rather than growing another type here.

### For whoever takes #436

`INPUTDEV_ROTARY` is declared in the enum but explicitly **not implemented** — reserved so PR 3 is a pure addition rather than a renumbering. Nothing in this build can select it: no option value maps to it, and `InputDevSetType()` treats it as not-attached. `InputDevOverlayF14002()` already takes the decoded `row0`/`row1` arguments the rotary's C2/C3 controller-type reporting needs (the mouse ignores them), so `joystick.c` will not have to change again.

Refs #429, #428. Stacked on #446.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -60,6 +60,8 @@ test/tools/fmv_seek_probe
 test/tools/sram_test
 test/tools/netlink_pair
 test/tools/netlink_game
+test/tools/joymatrix_identity
+test/tools/mouse_decode_test
 # vjtrace flight-recorder analyzer/smoke tools -- built on demand by
 # test/tools/vjtrace_selftest.sh (and by hand per each tool's own header
 # comment); never tracked, same Mach-O-on-Linux-CI hazard as above.
--- a/Makefile
+++ b/Makefile
@@ -852,7 +852,8 @@ clean:
 		test/tools/test_memory_map test/tools/test_option_visibility test/test_memtrack test/test_nvmbios test/tools/test_dsp_audio_diag \
 		test/tools/test_frame_timing test/tools/test_runahead_determinism test/tools/test_pertitle_db \
 		test/tools/test_wedge_spin test/tools/i2s_lag_probe \
-		test/tools/joymatrix_identity \
+		test/tools/joymatrix_identity test/tools/mouse_decode_test \
+		test/test_quadrature \
 		test/.skipped-checks
 
 # Self-contained unit tests (parser + list management + simulated
@@ -905,7 +906,8 @@ test: test/test_dram_timing test/test_cheat test/test_event_queue test/test_jlin
 		test/test_audio_dac test/test_blitter \
 		test/tools/test_memory_map test/tools/test_op_gpu_object test/tools/test_option_visibility test/test_memtrack test/test_nvmbios test/test_uart_core test/test_netlink_host \
 		test/tools/netlink_pair test/tools/netlink_latency test/tools/netlink_delay_proxy test/tools/test_pertitle_db \
-		test/tools/i2s_lag_probe test/tools/joymatrix_identity
+		test/tools/i2s_lag_probe test/tools/joymatrix_identity \
+		test/test_quadrature test/tools/mouse_decode_test
 	@# Skip ledger: truncate FIRST so a previous run's rows cannot resurface
 	@# as fresh skips (the stale-row failure mode documented for
 	@# cd_boot_matrix.sh).  Every optional check below records into it, and
@@ -1138,6 +1140,14 @@ test: test/test_dram_timing test/test_cheat test/test_event_queue test/test_jlin
 	@# an FNV digest measured on develop.  Committed before any device code
 	@# so later PRs are measured against a number that predates them.
 	./test/tools/joymatrix_identity ./$(TARGET) test/roms/yarc.j64 --quiet
+	@# Quadrature encoder unit test (no core): the Gray sequence in both
+	@# directions, the one-state-per-advance rate policy, the backlog
+	@# clamp and the Q8 carry.
+	./test/test_quadrature
+	@# ST/Amiga mouse end-to-end: synthetic deltas in, decoded direction
+	@# and distance out, all three wiring cases and all three poll shapes,
+	@# plus row-blindness and the v12 savestate round-trip.
+	./test/tools/mouse_decode_test ./$(TARGET) test/roms/yarc.j64 --quiet
 	./test/test_memtrack
 	./test/test_nvmbios
 	@# Option visibility is content-type dependent; the disc half runs only
@@ -1453,6 +1463,16 @@ test/tools/joymatrix_identity: test/tools/joymatrix_identity.c \
 		-o $@ test/tools/joymatrix_identity.c \
 		test/harness/harness.c \
 		$(if $(filter Linux,$(shell uname -s)),-ldl) -lm
+
+test/tools/mouse_decode_test: test/tools/mouse_decode_test.c \
+		test/harness/harness.c test/harness/harness.h
+	$(CC) -O2 -Wall -std=c99 $(INCFLAGS) \
+		-o $@ test/tools/mouse_decode_test.c \
+		test/harness/harness.c \
+		$(if $(filter Linux,$(shell uname -s)),-ldl) -lm
+
+test/test_quadrature: test/test_quadrature.c src/jerry/quadrature.c src/jerry/quadrature.h
+	$(CC) -O2 -Wall $(INCFLAGS) -o $@ test/test_quadrature.c src/jerry/quadrature.c
 
 test/tools/test_option_visibility: test/tools/test_option_visibility.c
 	$(CC) -O2 -Wall -std=c99 $(INCFLAGS) \
--- a/Makefile
+++ b/Makefile
@@ -852,6 +852,7 @@ clean:
 		test/tools/test_memory_map test/tools/test_option_visibility test/test_memtrack test/test_nvmbios test/tools/test_dsp_audio_diag \
 		test/tools/test_frame_timing test/tools/test_runahead_determinism test/tools/test_pertitle_db \
 		test/tools/test_wedge_spin test/tools/i2s_lag_probe \
+		test/tools/joymatrix_identity \
 		test/.skipped-checks
 
 # Self-contained unit tests (parser + list management + simulated
@@ -904,7 +905,7 @@ test: test/test_dram_timing test/test_cheat test/test_event_queue test/test_jlin
 		test/test_audio_dac test/test_blitter \
 		test/tools/test_memory_map test/tools/test_op_gpu_object test/tools/test_option_visibility test/test_memtrack test/test_nvmbios test/test_uart_core test/test_netlink_host \
 		test/tools/netlink_pair test/tools/netlink_latency test/tools/netlink_delay_proxy test/tools/test_pertitle_db \
-		test/tools/i2s_lag_probe
+		test/tools/i2s_lag_probe test/tools/joymatrix_identity
 	@# Skip ledger: truncate FIRST so a previous run's rows cannot resurface
 	@# as fresh skips (the stale-row failure mode documented for
 	@# cd_boot_matrix.sh).  Every optional check below records into it, and
@@ -1132,6 +1133,11 @@ test: test/test_dram_timing test/test_cheat test/test_event_queue test/test_jlin
 	./test/test_cart_format ./$(TARGET)
 	./test/test_audio_dac
 	./test/tools/test_memory_map ./$(TARGET)
+	@# $F14000/$F14002 identity guardrail for the input-devices track
+	@# (#428/#429/#436): sweeps the whole joystick read domain and asserts
+	@# an FNV digest measured on develop.  Committed before any device code
+	@# so later PRs are measured against a number that predates them.
+	./test/tools/joymatrix_identity ./$(TARGET) test/roms/yarc.j64 --quiet
 	./test/test_memtrack
 	./test/test_nvmbios
 	@# Option visibility is content-type dependent; the disc half runs only
@@ -1436,6 +1442,15 @@ test/tools/test_pertitle_db: test/tools/test_pertitle_db.c \
 		test/harness/harness.c test/harness/harness.h
 	$(CC) -O2 -Wall -std=c99 $(INCFLAGS) \
 		-o $@ test/tools/test_pertitle_db.c \
+		test/harness/harness.c \
+		$(if $(filter Linux,$(shell uname -s)),-ldl) -lm
+
+# $F14000 / $F14002 identity guardrail (#428 input-devices track).  Needs
+# the wide test ABI's Joystick* / joypad0Buttons / joypad1Buttons exports.
+test/tools/joymatrix_identity: test/tools/joymatrix_identity.c \
+		test/harness/harness.c test/harness/harness.h
+	$(CC) -O2 -Wall -std=c99 $(INCFLAGS) \
+		-o $@ test/tools/joymatrix_identity.c \
 		test/harness/harness.c \
 		$(if $(filter Linux,$(shell uname -s)),-ldl) -lm
 
--- a/Makefile
+++ b/Makefile
@@ -1458,6 +1458,7 @@ test/tools/test_pertitle_db: test/tools/test_pertitle_db.c \
 # $F14000 / $F14002 identity guardrail (#428 input-devices track).  Needs
 # the wide test ABI's Joystick* / joypad0Buttons / joypad1Buttons exports.
 test/tools/joymatrix_identity: test/tools/joymatrix_identity.c \
+		src/jerry/inputdev.h \
 		test/harness/harness.c test/harness/harness.h
 	$(CC) -O2 -Wall -std=c99 $(INCFLAGS) \
 		-o $@ test/tools/joymatrix_identity.c \
--- a/Makefile.common
+++ b/Makefile.common
@@ -59,6 +59,8 @@ SOURCES_C :=  \
 	$(CORE_DIR)/src/bios/jagcdbios.c \
 	$(CORE_DIR)/src/bios/jagdevcdbios.c \
 	$(CORE_DIR)/src/jerry/joystick.c \
+	$(CORE_DIR)/src/jerry/quadrature.c \
+	$(CORE_DIR)/src/jerry/inputdev.c \
 	$(CORE_DIR)/src/core/settings.c \
 	$(CORE_DIR)/src/core/memtrack.c \
 	$(CORE_DIR)/src/core/jaggd.c \
--- a/exports-test.list
+++ b/exports-test.list
@@ -62,7 +62,9 @@ _jaguar_cd_*
 _ResolveBootConfig
 _BlitterCompare*
 _DAC*
-_JoystickState*
+_Joystick*
+_joypad0Buttons
+_joypad1Buttons
 _MTState*
 _NVMBios*
 _NVM_*
--- a/exports-test.list
+++ b/exports-test.list
@@ -65,6 +65,8 @@ _DAC*
 _Joystick*
 _joypad0Buttons
 _joypad1Buttons
+_InputDev*
+_Quad*
 _MTState*
 _NVMBios*
 _NVM_*
--- a/libretro.c
+++ b/libretro.c
@@ -36,6 +36,7 @@ int64_t rfread(void* buffer, size_t elem_size, size_t elem_count, RFILE* stream)
 #include "jlink_netpacket.h"
 #include "uart.h"
 #include "joystick.h"
+#include "inputdev.h"
 #include "settings.h"
 #include "shadowfb.h"
 #include "blit_memo.h"
@@ -191,6 +192,61 @@ static bool show_cd_options        = true;
 static bool show_cart_bios_option  = true;
 static bool enable_alt_inputs = false;
 static uint8_t *joypad_buttons[2] = { joypad0Buttons, joypad1Buttons };
+
+/* ---- non-pad input devices (#428/#429) ------------------------------
+ *
+ * Subclass IDs for RETRO_ENVIRONMENT_SET_CONTROLLER_INFO.  The mouse is a
+ * RETRO_DEVICE_MOUSE subclass because it is a relative-motion pointer;
+ * the three subclasses are the three adapter/mouse wiring combinations
+ * (docs/jaguar-mouse-adapter-mapping.md section 4d), not three different
+ * devices.  Port 2 only -- see inputdev.h. */
+#define RETRO_DEVICE_JAG_PAD             RETRO_DEVICE_JOYPAD
+#define RETRO_DEVICE_JAG_MOUSE_ST        RETRO_DEVICE_SUBCLASS(RETRO_DEVICE_MOUSE, 0)
+#define RETRO_DEVICE_JAG_MOUSE_AMIGA     RETRO_DEVICE_SUBCLASS(RETRO_DEVICE_MOUSE, 1)
+#define RETRO_DEVICE_JAG_MOUSE_AMIGA_AD  RETRO_DEVICE_SUBCLASS(RETRO_DEVICE_MOUSE, 2)
+
+/* Device the frontend explicitly selected via
+ * retro_set_controller_port_device, or INPUTDEV_PAD when it never did.
+ * An explicit frontend selection outranks the core option (a frontend
+ * that sets a port device is making a deliberate statement); otherwise
+ * the option decides.  Not serialized: option/frontend derived. */
+static InputDevType port_device_frontend[2] = { INPUTDEV_PAD, INPUTDEV_PAD };
+static bool         port_device_forced[2]   = { false, false };
+/* Device actually attached, so the resolution is logged only when it
+ * changes rather than on every check_variables() call. */
+static InputDevType port_device_active[2]   = { INPUTDEV_PAD, INPUTDEV_PAD };
+static bool         show_mouse_options      = true;
+
+static const char *inputdev_type_name(InputDevType t)
+{
+   switch (t)
+   {
+      case INPUTDEV_MOUSE_ST:            return "Atari ST / PS2 mouse";
+      case INPUTDEV_MOUSE_AMIGA_ADAPTER: return "Amiga mouse (Amiga adapter)";
+      case INPUTDEV_MOUSE_AMIGA_ON_ST:   return "Amiga mouse (ST adapter)";
+      default:                           break;
+   }
+   return "standard joypad";
+}
+
+static void apply_port_device(int port, InputDevType type)
+{
+   InputDevSetType(port, type);
+
+   /* Read back: InputDevSetType refuses a mouse on port 1, so the
+    * resolved type is whatever it actually accepted. */
+   type = InputDevGetType(port);
+
+   if (type != port_device_active[port])
+   {
+      port_device_active[port] = type;
+      if (type != INPUTDEV_PAD)
+         LOG_INF("[input] port %d: %s attached\n", port + 1,
+                 inputdev_type_name(type));
+      else
+         LOG_INF("[input] port %d: standard joypad\n", port + 1);
+   }
+}
 
 static int number_keys[12] = {
    RETROK_MINUS,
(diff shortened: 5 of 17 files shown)


JoeMatt:
@kimi review

JoeMatt:
## Review fixes — all findings addressed

Pushed as `443595d`. Wiring tables, `STATE_VERSION` (11→12), `EXPECT_DIGEST_FULL` and `EXPECT_DIGEST_PORT1` all untouched; the two protected digests were re-measured **bit-identical**.

### Blocking

**A — the frontend's port-device call silently detached the mouse.** Confirmed: `retro_set_controller_port_device` forced `INPUTDEV_PAD` on any JOYPAD/NONE assignment, and the comment's "the core option takes over again on the next `check_variables()`" was false — nothing schedules one (`check_variables()` runs from `retro_load_game` and from `retro_run` only behind `GET_VARIABLE_UPDATE`). Option resolution is now factored into `p2_device_from_option()` and re-run **immediately** when the frontend releases its claim. Comment corrected. Noted in the code: called that early, `TitleDBOverride()` returns NULL with no title loaded, so the documented `auto → per-title DB` leg genuinely cannot fire on the early call — it resolves at the `retro_load_game` `check_variables()`.

**A2 — selecting the mouse could leave port 2 with no input.** Rule chosen and commented in `libretro.c`:

> Selecting a mouse must never leave the port with no working input. Mouse state is read **first**; the RetroPad is released only once the mouse has proven live (a non-zero delta or a button). Once live it stays released for the session, or until the device type changes.

So a frontend that never routes mouse state to the port keeps its pad, and a mouse that has moved once owns the port unambiguously.

**B — the third assertion did not assert the thing it claimed.** Fixed, plus two things found while fixing it:

- `--print` could not measure the new constant at all: the `print_only` branch returned *before* the `InputDev*` dlsym block. The mouse sweep now runs ahead of the early return and `--print` reports all three digests.
- The old code folded the mouse result into `ok_port1`; when I split it into its own row, the FAIL row printed but **`main` still returned 0**. That is the skip-as-pass class wearing a different hat. All four assertions now gate the exit status. The sabotage below is what surfaced it.

`0xE0E2711A` from the review was measured before D and E, both of which change which phases the sweep visits. Re-measured after everything landed: **`EXPECT_DIGEST_FULL_WITH_MOUSE = 0x5E1858EA`**.

The `:333` comment is corrected rather than re-fed — re-feeding would put mouse-specific code inside `run_sweep()`, the one function that also produces the two protected digests. It now says the encoder drains over the first 64 qualifying polls of ~131,072 and then holds a fixed phase, which is exactly what makes the polarity class visible.

**C — missing symbols skipped the assertion and exited 0.** Now `harness_shutdown()` then `return 77`. **Reporting as asked: the Makefile does not treat 77 as a skip — and no change is needed.** `make test` invokes the tool bare (`./test/tools/joymatrix_identity ./$(TARGET) …`), so any non-zero status, 77 included, fails the build. That is louder than a skip and satisfies "never a silent pass". The repo's actual skip idiom is the `test/.skipped-checks` ledger, which is for *optional private-corpus* checks; this branch's trigger is an `exports-test.list` / `link-test.T` divergence, i.e. a build bug in this repo, which should never land in that ledger.

### Strong

**D — an idle mouse asserted all four direction lines.** `QuadReset` now parks at `QUAD_REST_PHASE = 2` = `(A=1,B=1)`, the only rest position where an idle mouse is bit-identical to an idle pad under the active-low overlay. The comment states plainly that **no source names a rest phase** — it is a free choice on the indistinguishable-from-idle argument, and notes #436's rotary shares the module and may want to revisit it.

**E — the emission clock was not gated on port 2's row select.** Gated on `offset1 != 0xFF` — any port-2 row, not row 0 — with the `InputDevClock()` call moved below the offset decode. Only the *clock* is gated; the overlays still assert in every row, and the pad-only path is untouched (the call still sits behind `!inputdev_attach_mask`), which the two unchanged protected digests confirm. `mouse_decode_test`'s poll-shape and rate-ceiling tests drive `$817F/$81BF/$81DF/$81EF`, all valid port-2 rows, and all still pass unmodified.

New test so E is not an unverified claim — `test_port1_poll_does_not_clock` in `mouse_decode_test.c`: 200 port-1 row polls (`$81F7`: port 1 row 0, port 2 unselected) must leave `$F14000` bits 12-15 untouched, and the motion queued before them must still decode afterwards. **Both legs fail with E reverted** (see below).

**F** — `inputdev_armed` cleared on the type-change path in `InputDevSetType` (not unconditionally: `apply_port_device` calls it on every `check_variables()`, and an unconditional clear would eat a legitimate arm).
**G** — `InputDevStateLoad` now clamps `backlog` to ±`QUAD_MAX_BACKLOG` and `frac` to ±255 alongside the existing `phase &= 3`.
**H** — `headroom_logged` was a *function-local* static in `retro_serialize`, unreachable from `retro_deinit`; hoisted to file scope and added to the reset block.
**I** — sensitivity percent clamped to 1600 (`InputDevSetScale`'s 4096 ceiling as a percent) **before** the `*256`.
**J** — `src/jerry/inputdev.h` added as a prerequisite of the `joymatrix_identity` rule.
**K** — PR body corrected: the row now reads **+22 / -2**, explains the two removed `return data;` lines, and notes `InputDevArm()` is in `JoystickWriteWord`, which has no `joysticksEnabled` early return.

---

## Verification

**C89 lint, every touched `.c`:**

```
$ bash scripts/c89-lint.sh libretro.c src/jerry/inputdev.c src/jerry/quadrature.c \
      src/jerry/joystick.c test/tools/joymatrix_identity.c \
      test/tools/mouse_decode_test.c test/test_quadrature.c
C89 lint passed
LINT EXIT=0
```

**Build and suite** (make's own status, not a pipeline's):

```
$ DEVELOPER_DIR=/Library/Developer/CommandLineTools make -j8
MAKE EXIT=0

$ DEVELOPER_DIR=/Library/Developer/CommandLineTools make TEST_EXPORTS=1 test
MAKE TEST (full corpus) EXIT=0
```

Run with the private corpus symlinked, so the audio/savestate/per-title checks actually executed — one skip, unrelated (`Netlink latency (frame quantization)`, host too loaded to demonstrate the effect).

**Guardrail, all four assertions:**

```
  PASS: [joymatrix_identity_port1_with_mouse] port-1 bits with an ST mouse on port 2: 0xD0CD02E2 (expected 0xD0CD02E2)
  PASS: [joymatrix_identity_full_with_mouse] full digest with an ST mouse on port 2: 0x5E1858EA (expected 0x5E1858EA); 524288 reads
EXIT=0
```

Digests 1 and 2 are the protected constants, unmoved.

**Sabotage 1 — polarity inversion** (`inputdev.c`, `if (!ax)` → `if (ax)` on all four lines):

```
  PASS: [joymatrix_identity_full] digest 0xC24DDCDE (expected 0xC24DDCDE); ...
  PASS: [joymatrix_identity_port1] port-1 bits digest 0xD0CD02E2 (expected 0xD0CD02E2); ...
  PASS: [joymatrix_identity_port1_with_mouse] port-1 bits with an ST mouse on port 2: 0xD0CD02E2 (expected 0xD0CD02E2)
  FAIL: [joymatrix_identity_full_with_mouse] full digest with an ST mouse on port 2: 0x5C4A4EAA (expected 0x5E1858EA); 524288 reads
SABOTAGE LEG: JOYMATRIX EXIT=1
```

And the review's claim confirmed — every modulo-4 check is blind to it:

```
$ ./test/tools/mouse_decode_test ... ; echo $?
MOUSE_DECODE EXIT=0
```

Reverted:

```
REVERTED LEG: JOYMATRIX EXIT=0
```

**Sabotage 2 — E's gate removed** (`if (offset1 != 0xFF) InputDevClock();` → `InputDevClock();`):

```
emission clock is gated on port 2's row select
  FAIL 200 port-1 row polls do not advance the port-2 encoder ($F14000 bits 12-15 held at $D)
  FAIL the motion queued before them is not lost (NET_X=0 expect 30, dropped=0)
UNGATED: MOUSE_DECODE REAL EXIT=1
```

Both legs red without the gate, both green with it — `NET_X=0` is precisely the lost-motion failure §5's ceiling exists to prevent.

**Savestate:**

```
  ok   retro_serialize succeeded mid-motion
  ok   retro_unserialize accepted the v12 state
  ok   post-restore replay matches (NET_X 25 vs 25, NET_Y 19 vs 19, dropped 0 vs 0)
  ok   a v11 state (the v3.3.0 release layout) still loads
  ok   InputDevStateSave wrote 41 bytes; inputdev.h advertises INPUTDEV_STATE_SIZE = 41

[state] v12 payload 2396971 / 2621440 bytes (224469 free)
```

---

### Still owed

Nothing headless can settle: **A and A2 are frontend-behaviour fixes and want a RetroArch check on a real mouse title.** The headless harnesses drive `JoystickReadWord` directly and never exercise `retro_set_controller_port_device` the way a frontend does, which is exactly the blind spot that let A ship.

JoeMatt:
### Follow-up `0b5ab5f` — finding A is now pinned by a test, not by inspection

A was the finding called out as the reason the feature was never seen working, and I had only fixed it by inspection. It now has a regression test.

**`test_frontend_pad_assignment_rereads_option`** (`mouse_decode_test.c`) sets `virtualjaguar_p2_device=mouse_st` *after* the load — the harness's environment callback serves `cfg.options` live, so this measures what `retro_set_controller_port_device` resolves at call time, which is the thing under test — then issues the JOYPAD and NONE assignments RetroArch sends during its post-load per-port init.

With the fix in place:

```
frontend port-device assignment vs the core option
  ok   RETRO_DEVICE_JOYPAD on port 2 re-reads the core option instead of forcing pad (type=1, expect 1)
  ok   RETRO_DEVICE_NONE on port 2 likewise (type=1, expect 1)
  ok   an explicit frontend mouse on port 2 is honoured
  ok   a frontend mouse on port 1 is still refused
```

**Sabotage 3 — fix A reverted** (drop the `if (type == INPUTDEV_PAD) type = p2_device_from_option();` re-resolution):

```
frontend port-device assignment vs the core option
  FAIL RETRO_DEVICE_JOYPAD on port 2 re-reads the core option instead of forcing pad (type=0, expect 1)
  FAIL RETRO_DEVICE_NONE on port 2 likewise (type=0, expect 1)
  ok   an explicit frontend mouse on port 2 is honoured
  ok   a frontend mouse on port 1 is still refused
```

`type=0` is `INPUTDEV_PAD` — the shipped behaviour, reproduced exactly. Reverted, green again; `make TEST_EXPORTS=1 test` exit 0 with the private corpus, C89 lint clean.

**Scope note, so it is on the record:** A2 is still *not* covered headlessly and I am not claiming otherwise. Its liveness gate depends on `input_state_cb` returning real mouse deltas, which no harness here can supply — it wants the RetroArch check.

Two smaller notes from a second pass:

- `inputdev_live` is deliberately **not** serialized, and the comment now says why: it describes host input routing, not machine state — the same argument `inputdev.h` already makes for the device type and the sensitivity scale. The resulting asymmetry is bounded and stated: a state saved while the mouse was live loads with port 2's pad un-suppressed until the mouse next moves.
- Behaviour change worth recording for finding C: run by hand against a **production-slim** dylib (where `InputDev*` genuinely is not exported), `joymatrix_identity` now exits 77 where it previously exited 0 after two passing assertions. `make test` only ever invokes it under `TEST_EXPORTS=1`, so nothing in CI is affected.

PR #450, opened 2026-08-14
docs(input): mouse user guide, and the evidence that no titledb row should ship (#429)

JoeMatt:
## What this is

PR 4 of 4 of the input-devices track (epic #428): steps **8-9** of
`docs/superpowers/specs/2026-08-14-input-devices-mouse-rotary-design.md` — the
per-title auto-select and the docs.

**The headline is a negative result.** Step 8 was conditional by design — "titledb:
JagDoomEX rows gain `{"virtualjaguar_p2_device","mouse_st"}` — **only if** step 5
confirmed JagDoomEX reads the mouse". It does not. **No titledb row ships**, and the
evidence is recorded at the site so the next person does not add one on the strength of
#429's prose.

So the diff is: one evidence comment in `src/core/titledb.c`, a new user guide, a
settled item in the mapping doc, and two README lines. No behaviour change at all.

### Stacking

Branches off `feat/429-mouse-support` (#449), which branches off
`test/429-input-guardrail` (#446). Targets `develop`. Until those merge the diff shown
here includes their commits; the commit that belongs to this PR is the last one.

---

## Titles: what was added, and what was omitted and why

**Rows added: none.** Every candidate from #429 was checked, and each one failed for a
different, specific reason.

| Candidate | CRC verified? | Row? | Why |
|---|---|---|---|
| **JagDoomEX** (9 CRCs) | **Yes, all nine** | **No** | Disassembly proves it never reads port 2 — see below |
| **Elansar** | n/a | **No** | CD title; titledb is cartridge-CRC-only and structurally cannot match |
| **Philia** (Elansar sequel) | n/a | **No** | Same |
| Defender of the Crown | **No** — absent from corpus | **No** | Cannot verify a CRC; a guessed row is worse than none |
| Gridrunner (Jeff Minter Classics) | **No** — absent | **No** | Same |
| Star Wars Arcade (ST port) | **No** — absent | **No** | Same |

### JagDoomEX: CRCs verified, then the row rejected on the evidence

Every CRC in the Doom EX block was **reproduced from first principles** — the IPS patch
applied to the verified retail dump — not copied from the table:

```
base: test/roms/private/ROMS/Doom - Evil Unleashed (1994).jag   crc=0x5E2CDBC0

JagDoomEX.ips                            len=4194304 crc=0x754096DB
JagDoomEX 2.ips                          len=4194304 crc=0x4643E9DB
JagDoomEX 3.ips                          len=4194304 crc=0x35743B9C
JagDoomEX 4.ips                          len=4194304 crc=0xAD6B68BA
JagDoomEX 5.ips                          len=4194304 crc=0xC4F4CACF
JagDoomEX 6.ips                          len=4194304 crc=0x1F4EE4A5
JagDoomEX[spectral].ips                  len=4194304 crc=0x013A5359
JagDoomEX[transparent].ips               len=4194304 crc=0xB92D1CA3
JagDoom2EX.ips                           len=4194304 crc=0xEA12E234
```

All nine match `src/core/titledb.c` exactly. Also checked: the local
`Doom (World) EX.j64` (`0xEE7B84EB` — a build with **no** titledb row) and the newest
archive `JagDoomEX-ips-250318.7z`, which reapplies to `0x35743B9C`, i.e. the shipped
patch set is the one already covered.

**And then the row was rejected anyway, because none of these builds reads port 2.**
Three independent mechanisms, any one of which is sufficient:

**1. It never selects a port-2 row.** Scanning every `move.w #imm,$F14000.l` in each
image gives exactly four immediates, identical in retail Doom and all ten EX builds:

```
Doom - Evil Unleashed (1994).jag crc=0x5E2CDBC0  rowselects=['$81F7','$81FB','$81FD','$81FE']  port2-selecting=[]
Doom (World) EX.j64              crc=0xEE7B84EB  rowselects=['$81F7','$81FB','$81FD','$81FE']  port2-selecting=[]
JagDoomEX.ips.j64                crc=0x754096DB  rowselects=['$81F7','$81FB','$81FD','$81FE']  port2-selecting=[]
JagDoomEX 2.ips.j64              crc=0x4643E9DB  rowselects=['$81F7','$81FB','$81FD','$81FE']  port2-selecting=[]
JagDoomEX 3.ips.j64              crc=0x35743B9C  rowselects=['$81F7','$81FB','$81FD','$81FE']  port2-selecting=[]
JagDoomEX 4.ips.j64              crc=0xAD6B68BA  rowselects=['$81F7','$81FB','$81FD','$81FE']  port2-selecting=[]
JagDoomEX 5.ips.j64              crc=0xC4F4CACF  rowselects=['$81F7','$81FB','$81FD','$81FE']  port2-selecting=[]
JagDoomEX 6.ips.j64              crc=0x1F4EE4A5  rowselects=['$81F7','$81FB','$81FD','$81FE']  port2-selecting=[]
JagDoomEX[spectral].ips.j64      crc=0x013A5359  rowselects=['$81F7','$81FB','$81FD','$81FE']  port2-selecting=[]
JagDoomEX[transparent].ips.j64   crc=0xB92D1CA3  rowselects=['$81F7','$81FB','$81FD','$81FE']  port2-selecting=[]
JagDoom2EX.ips.j64               crc=0xEA12E234  rowselects=['$81F7','$81FB','$81FD','$81FE']  port2-selecting=[]
```

The port-2 nibble is `$F` in all four — every port-2 row deselected.

**2. It masks the mouse's bits away at the read.** The poll, disassembled from the
`0x1F4EE4A5` build:

```
00802200  223cf0fffffc           move.l #$f0fffffc, d1
00802206  74ff                   moveq #$ff, d2
00802208  33fc81fe00f14000       move.w #$81fe, $f14000.l
00802210  203900f14000           move.l $f14000.l, d0
00802216  8081                   or.l d1, d0
00802218  e898                   ror.l #$4, d0
0080221A  c480                   and.l d0, d2
0080221C  33fc81fd00f14000       move.w #$81fd, $f14000.l
...
```

`or.l #$F0FFFFFC` before the `and.l` into an 8-bit accumulator keeps only `$F14000`
bits 8-11 and `$F14002` bits 0-1 — the port-1 columns. `$F14000` bits 12-15, the four
lines a mouse adapter drives, **cannot reach the result**.

**3. There is no third path to the register.** All **11** absolute-long `$F14000`
references in the image are accounted for: 8 in that poll (4 writes + 4 reads), one
`move.l #$100,$F14000` at init, and two `lea $F14000,a0` in the EEPROM driver (which
reads bit 0 and `$F15000`). A driver reaching the register through a register-indirect
base would still need one of these to load the address, so the enumeration is closed.

**Conclusion:** a row would disconnect the port-2 RetroPad for a title that reads
nothing from port 2, in exchange for no function. #429's checklist item *"Auto-select
the mouse for JagDoomEX by CRC"* is **not applicable to any released build**. If a
future JagDoomEX adds mouse support it will hash differently from all nine CRCs above
and needs its own row and its own re-check — which is exactly what the comment in
`titledb.c` says.

### Elansar: the wiring decision, and why it was dissolved rather than made

The task asked whether to set only the device and leave the wiring at default, since
Elansar has an in-game Atari/Amiga selector that a titledb-forced wiring case could
fight. **Neither: no row is possible at all.**

`libretro.c` gates the titledb feed on `info->data && !is_cd_content`, and its own
comment states *"v1 only covers cartridge CRCs, and hashing a disc image would find
nothing this table knows about while risking a collision handing a CD title some
cartridge's per-title overrides."* Elansar and Philia are **CD** titles. They can never
match a row, so the device-vs-wiring question does not arise — and it would have been
the wrong call to make in a database regardless, because the user's in-game selector is
the more specific signal.

That is a better outcome than either choice: it means Elansar's wiring case is
permanently a documented user decision, and it is now documented as one, in the user
guide and in `docs/jaguar-mouse-adapter-mapping.md` §8 A.

### The three absent titles

Defender of the Crown, Gridrunner (Jeff Minter Classics) and Star Wars Arcade are not in
the private corpus (confirmed with `find -L`, since `test/roms/private` is a symlink and
`find` does not follow those). A row keyed on a guessed CRC either does nothing or fires
on the wrong game, so none was added. They are listed in the user guide as
community-reported mouse titles with an explicit note that the core has not verified
them.

---

## What actually changed

- **`src/core/titledb.c`** — a comment block on the Doom EX family recording the
  evidence above and instructing the next person not to add the row without re-verifying
  against the build in hand. No table rows changed.
- **`docs/input-devices-user-guide.md`** (new) — the core options, the three wiring
  cases and how to tell you picked the wrong one, why raising sensitivity past the
  game's poll rate buys lag rather than speed, the mouse-capable titles, the
  reproduced-not-broken hardware quirks (phantom keypad digits, LMB reading as
  A+B+C+Option, RMB perturbing the controller-type field), and the frontend limitation
  below.
- **`docs/jaguar-mouse-adapter-mapping.md`** — §8 F's "disassemble Elansar or JagDoomEX"
  is now split: **JagDoomEX done, negative** (new §8 F.1 with the full evidence);
  **Elansar still not attempted**, and §8 A still open, with the added note that its
  answer can never live in the database.
- **`README.md`** — one feature bullet and one docs-index entry.

### The documented limitation

The guide states plainly that **real-frontend mouse routing is unverified**: RetroArch
has historically bound the host mouse to port index 0, and #449 deliberately shipped
**no speculative port-0 fallback** — silently reading port 1's mouse into a port-2
device would be a guess, invisible until it misbehaved. The guide tells a user that "the
pointer does not move at all" is the symptom of that, not of the emulation, and asks for
a frontend + version on #429.

---

## Verification

**`bash scripts/c89-lint.sh src/core/titledb.c`**

```
C89 lint passed
```

**`DEVELOPER_DIR=/Library/Developer/CommandLineTools make -j10`** — clean, zero
warnings, links:

```
cc -ovirtualjaguar_libretro.dylib  ./libretro.o ... -Wl,-dead_strip -fPIC -dynamiclib -Wl,-exported_symbols_list,exports.list
```

**`DEVELOPER_DIR=/Library/Developer/CommandLineTools make TEST_EXPORTS=1 test`** —
`exit=0`, with the private corpus symlinked so the optional rows actually ran:

```
```

Zero `FAIL` lines in the whole run.

**Guardrail — the digests did not move.** This PR touches no input code, so any movement
would mean something was wrong. Constants untouched:

```
```

**`test_pertitle_db` — all six cases pass, including the `--case 5` `yarc.j64` non-DB
control.** That trap is trivially safe here: no row was added, so no CRC can start
matching that did not before.

```
=== Results: 2 passed, 0 failed, 0 skipped ===
  PASS: [case1_hires_db_applied] shadowHiresN == 2 (DB applied at default)
  PASS: [case1_titledb_log_present] [titledb] substitution logged (option at default)
=== Results: 2 passed, 0 failed, 0 skipped ===
  PASS: [case2_gate_disabled_is_stock_hires] shadowHiresN == 1 (defaults disabled -> stock)
  PASS: [case2_gate_disabled_is_stock_truecolor] shadowFBActive == 0 (defaults disabled -> stock)
  PASS: [case3_manual_choice_with_db_off] shadowHiresN == 2 (explicit user choice honored)
=== Results: 2 passed, 0 failed, 0 skipped ===
  PASS: [case4_hires_still_applied] shadowHiresN == 2
  PASS: [case4_default_valued_option_substituted] shadowFBActive != 0 (true_color=disabled == default, DB substituted enabled)
=== Results: 2 passed, 0 failed, 0 skipped ===
  PASS: [case6_title_cached_after_load] TitleDBTitleName() non-NULL after matched load
  PASS: [case6_unload_clears_titledb] retro_unload_game() cleared titledb state
./test/tools/test_pertitle_db ./virtualjaguar_libretro.dylib test/roms/yarc.j64 --case 5 --quiet
  PASS: [case5_nondb_rom_stock] shadowHiresN == 1 (no DB entry for this CRC)
  PASS: [case5_no_substitution_log] no [titledb] substitution log line (correctly no match)
  PASS: [case5_miss_logged] [titledb] miss line logged for unlisted CRC
```

The private-corpus symlink was removed after the run; its target was not touched.

---

## Note for the reviewer

Do not read the empty titledb diff as unfinished work — it is the deliverable. The spec
made the row conditional precisely so that this could be the answer, and the value here
is that the condition was actually tested rather than assumed. The alternative on offer
was a row that fires on nine real ROMs and silently disconnects their port-2 pad for no
benefit.

Refs #429, #428. Stacked on #446 and #449.


JoeMatt:
Rebuilt on the current `develop` (e6dbb11) rather than merged.

The old branch carried its own copy of the mouse implementation from before #449 was reviewed — merging it would have resurrected the pre-fix version of `inputdev.c`/`libretro.c` on top of the merged one. So this branch is now `develop` + the docs commit only, force-pushed. The diff is docs + one comment block in `titledb.c`; no code changes.

While re-checking it against what actually merged, three claims in the guide turned out to describe the pre-review behaviour and are corrected in a second commit:

- the port-2 RetroPad is **not** dropped the moment a mouse is selected — `inputdev_live[]` defers the suppression until the frontend has proven it routes mouse state, so a mouse selected but not routed leaves port 2 behaving exactly like a joypad;
- a frontend setting `RETRO_DEVICE_JOYPAD`/`NONE` *releases its claim* back to the core option rather than forcing a pad (that was fix A on #449 — the bug that kept the feature from ever working in RetroArch);
- the symptom of the unverified-routing limitation is therefore "the mouse does nothing but the pad still works", not "port 2 is dead", which is what a user filing the bug would actually see.

Note on review: Kimi is out of quota for this billing cycle (HTTP 403 `access_terminated_error`), and Copilot's credits are exhausted, so neither bot has reviewed this. Reviewed by hand instead.

Issue #439, opened 2026-08-13
[feat] Per-axis input tuning: dead zone, offset, exponent


JoeMatt:
Sub-issue of #428. BigPEmu offers per-axis **dead zone**, **X/Y offset**, and **exponent curve** on its analog device types. Worth matching, but only once we have more than one analog-ish device to tune — otherwise the options are a settings-menu tax on a feature nobody can use yet.

**Estimate:** S — option plumbing plus a small amount of arithmetic, once the devices exist.

## Scope

- Per-axis dead zone, offset, and response exponent, exposed as core options.
- Applied uniformly to every analog input source: mouse (#429), rotary (#436), analog/driving (#437). One implementation, not per-device copies.
- Sensible defaults such that a user who never opens the menu gets correct behaviour.

## Sequencing

Deliberately **last** in #428. Do not start until at least two of the device tickets have landed — the shape of the shared tuning layer should be derived from real consumers, not guessed in advance.

## Guardrail

Tuning applies only to analog device paths. The standard digital pad is untouched.

JoeMatt:
Shipped in v3.4.0: dead zone, offset and response exponent on every analog source, surviving reset and pinned by guardrail digests.

Includes the fix for the failure mode this option exists to prevent — a non-zero idle offset was manufacturing phantom motion every frame from an untouched device.

Re-milestoned from v3.5.0 to v3.4.0, which is where the work actually landed.

PR #451, opened 2026-08-14
feat(input): Tempest rotary controller on both ports (#436)

JoeMatt:
## What this is

PR 3 of 4 of the input-devices track (epic #428): the **Tempest rotary** (#436).

Implements step **7** of `docs/superpowers/specs/2026-08-14-input-devices-mouse-rotary-design.md`. Step 1 (the guardrail) is #446, steps 2-6 (the mouse) are #449, and steps 8-9 (titledb auto-select + docs) are PR 4.

### Stacking on #446 and #449

This branches off `feat/429-mouse-support`, **not** `develop`. Until #446 and #449 merge, the diff shown here **includes their commits** (`test/tools/joymatrix_identity.c`, `src/jerry/quadrature.{c,h}`, `src/jerry/inputdev.{c,h}`, the libretro device scaffolding and the v12 savestate chunk). That work is not being re-landed. Once both merge this rebases away to my one commit.

Refs #436, #428.

---

## The load-bearing fact: the rotary is a matrix device and the mouse is not

These two peripherals share an encoder and nothing else, and keeping that distinction load-bearing is most of this PR.

The **mouse adapter is row-blind**. It does not connect Jaguar pins 1-4 -- exactly the port-2 row-select outputs -- so all six input lines carry its state identically in every row. It cannot be expressed as pad slots, and #449 gives it a row-independent overlay on the raw `$F14000` bits.

The **rotary is the opposite shape**. TR10, section "Rotary 'Tempest' Controller":

> Although originally intended to be a bank switching design all existing rotary controllers are modified standard controllers, consequently they should be read just like a standard controller using Socket 0 row codes ...

Two independent corroborations that this is genuine, and not a detail to generalise away:

- **Wiring** (Zerosquare, AtariAge "Rotary trouble"): the encoder's common wire goes to joystick port **pin 4** -- the row strobe -- "not ground. If you used an actual ground, it won't work properly." Port-1 pin 4 is `J0`, and row 0 is the only row code that drives `J0` low (`J3 J2 J1 J0 = 1110`). So the phases appear in row 0 and **only** row 0.
- **Domin's `oneaxis.s`** reads socket A with `move.w #$817f,JOYSTICK`, masks down to just the L and R bits and indexes a 16-entry `(old<<2)|new` table whose four double-transition entries are commented "undetermined" -- the same decoder shape, and therefore the same rate ceiling, as the mouse.

So the rotary **reuses `joypadNButtons[]` untouched**. No new assertion mechanism:

| Signal | Slot | Port 1 bit | Port 2 bit |
|---|---|---|---|
| Phase 0 | `BUTTON_L` | 10 (J10) | 14 (J14) |
| Phase 1 | `BUTTON_R` | 11 (J11) | 15 (J15) |

`BUTTON_U` / `BUTTON_D` are forced clear -- the controller physically has no up/down, which is why TR10 tells rotary games to use `A = Up` and `C = Down` for menus. Rows 1-3 remain an ordinary pad.

**Polarity is inverted relative to the mouse path**, which is the single easiest way to ship a backwards spinner. TR10 gives the phase sequences as pin levels and "reading a zero means the appropriate button is depressed"; `joystick.c` does `data &= (joypadNButtons[slot] ? mask : 0xFFFF)`, so a *non-zero* array entry pulls the bit low. Hence `pin level 1 -> 0x00`, `pin level 0 -> 0xFF`. `QuadLevels()` returns logic levels precisely so each sink applies its own polarity at one visible place. The encoder itself is `src/jerry/quadrature.{c,h}` from PR A2, reused as-is -- one axis, not two, and no second encoder.

---

## Which poll-shape branch I took, and the evidence

The spec (section 10, step 7) carries a defined branch: if Tempest 2000 polls read-only -- row select written once at init, then reads forever -- the arm-on-write emission rule cannot clock and the stall-breaker must be implemented here.

**Measured, not assumed.** `vjtrace --watch 0xF14000:4:rw` over 400 frames of `Tempest 2000 (1994).jag`:

| Write pc | value | selects | Read pc | count |
|---|---|---|---|---|
| `80FD2E` | `$81FE` | port 1 row 0 | `80FD36` | 330 |
| `80FD42` | `$81FD` | port 1 row 1 | `80FD4A` | 330 |
| `80FD56` | `$81FB` | port 1 row 2 | `80FD5E` | 330 |
| `80FD6C` | `$81F7` | port 1 row 3 | `80FD74` | 330 |
| `80FDA2` | `$817F` | port 2 row 0 | `80FDAA` | 330 |
| `80FDB8` | `$81BF` | port 2 row 1 | `80FDC0` | 330 |
| `80FDD0` | `$81DF` | port 2 row 2 | `80FDD8` | 330 |
| `80FDE6` | `$81EF` | port 2 row 3 | `80FDEE` | 330 |

330 scans over 331 frames -- one full eight-row scan per frame, every read preceded by a row-select write from the paired instruction, each read a `move.l` (one offset-0 access and one offset-2 access).

**Branch taken: shape A/C. The stall-breaker is NOT implemented.** Shape B does not occur, and `rotary_decode_test` pins that as a recorded property rather than an assumption: shape B is asserted to decode zero motion, so it remains the detector that would justify adding one if a real title ever needs it.

---

## What the measurement also exposed: a defect the spec did not model

Eight armed offset-0 reads per frame, but exactly **one row-0 sample per port**. A rotary's phases exist in row 0 only, so the three other reads in each port's scan decode nothing.

Clocked with the mouse's rule -- one state per armed read -- the encoder drains up to eight states between the samples a driver actually decodes, making the decoded result `(states drained) mod 4`:

| states drained between samples | decoded |
|---|---|
| 1 | +1, correct |
| 2 | "undetermined", discarded |
| 3 | **-1 -- a sign inversion** |
| 4, 8 | 0, no movement at all |

The device would work at exactly one count per frame and nowhere else, and no sensitivity setting can compensate. That is precisely the jitter the spec's section 3 emission rule exists to prevent.

So `InputDevClock()` now takes the decoded socket-0 row per port and **a rotary advances only on an armed row-0 read of its own port**. The row-blind mouse ignores both arguments, exactly as it already ignores them in `InputDevOverlayF14002()`.

### This required a change to `joystick.c`, and that was against my brief

My instructions said `joystick.c` is final because `InputDevOverlayF14002()` already carries the `row0`/`row1` arguments the rotary needs -- which is true of the C2/C3 controller-type reporting, the need the spec author checked, but not of this second need for row information. **Flagging it explicitly for review.** The change:

```c
 offset0 = joypad0Offset[joystick_ram[1] & 0x0F];
 offset1 = joypad1Offset[(joystick_ram[1] >> 4) & 0x0F];

 InputDevClock(offset0 == 0xFF ? 0xFF : offset0 / 4,
               offset1 == 0xFF ? 0xFF : offset1 / 4);
```

Two pure table lookups **hoisted, not added**, and one call signature widened. `joystick.c` gains **no lines** and loses none, preserving the spec's section 12 constraint in substance. `InputDevClock()` is a PR-2 symbol not yet in `develop`, so widening it breaks no shipped ABI.

Alternatives considered and rejected:

- **Advance from the `$F14002` overlay, which already has the rows.** Works for Tempest 2000 and any `move.l` driver, but a driver reading only `move.w $F14000` never reaches that hook and the rotary freezes **silently**. Choosing a shape-dependent mechanism immediately after empirically discovering that poll shape varies is the wrong direction, and eight of the nine rotary titles are homebrew I have no ROM for.
- **Decode the row from the assembled `$F14000` word.** Port 2 works (bit 7 is clear in row 0), but port 1's row-0 mask is `0xFFFF` -- it clears nothing -- so "port 1 row 0" is indistinguishable from "port 1 not addressed". Dead end, and port 1 is the primary case.
- **Read `joystick_ram` from `inputdev.c`.** It is `static` in `joystick.c`.

**If the hoist is rejected, the honest fallback is the `$F14002` oracle with its known silent-freeze hole** -- please say which you prefer rather than merging a rotary that only works at 60 counts/s.

---

## Guardrail: both digests unchanged

The headline check. With no device selected, `$F14000`/`$F14002` must be bit-identical to before -- and the `joystick.c` hoist is exactly the kind of change that could perturb it.

```
harness: core Virtual Jaguar v3.3.0 5c0403a-dirty (./virtualjaguar_libretro.dylib)

```

Neither expected digest was touched. The hoisted lookups are pure table reads, and this is the proof rather than the claim.

---

## Savestate: no bump, no growth

**`STATE_VERSION` stays at 12 and the trailing chunk stays 41 bytes.** This was checked, not assumed -- the finding is that the rotary introduces no new machine-visible state:

| State | Where it already lives |
|---|---|
| backlog / Q8 carry / phase | the port's existing `quad_axis x`, already in the v12 chunk |
| published phase | `joypadNButtons[]`, already serialized by `JoystickStateSave()` |
| controller-type ID flag | option-derived, like the sensitivity scale -- deliberately not saved |

Measured headroom after this change, from the core's own one-shot report:

```
[state] v12 payload 2396971 / 2621440 bytes (224469 free)
```

Identical to A2's measurement, confirming zero bytes added. No second version bump inside one release.

---

## The two known Tempest 2000 rotary bugs are reproduced deliberately

BigPEmu ships a hash-gated script patching two defects: the **high-score initials-entry screen is uncontrollable under rotary** (that screen consumes only the rotary input path), and the **bonus games' rotary accumulators never decay**, so spin velocity runs away.

**Both are game logic sitting above the decode path, and a correct row-0 quadrature implementation reproduces them automatically. That is the intended outcome.** Our contract is the hardware: a title that misbehaves on a real rotary should misbehave here. Fixing them requires ROM-address patching keyed to a specific dump -- which is exactly why BigPEmu hash-gates its script -- and that is **#370's per-title-hooks layer, not this PR's**. Nothing of BigPEmu's was downloaded, decompiled or copied; only the defect descriptions, which are facts about the game.

Independent corroboration of the second bug's class from a rotary builder: detentless optical encoders cause "drifting in the menu ... it's really hard to select what you want" -- the same runaway-accumulation signature.

---

## User-facing surface

Device selection goes through the **existing** `virtualjaguar_p2_device` scaffolding from PR A2 -- no second mechanism.

- `virtualjaguar_p2_device` gains a `rotary` value.
- `virtualjaguar_p1_device` is new: `auto` / `pad` / `rotary`. Port 1 gets no mouse (vendor-documented port-2 device), but TR10 documents the rotary matrix for both ports and Tempest 2000's menu carries independent P1/P2 toggles.
- `virtualjaguar_rotary_sensitivity` (25-400%) and `virtualjaguar_rotary_id`, both hidden unless a rotary is attached.
- `RETRO_DEVICE_JAG_ROTARY = RETRO_DEVICE_SUBCLASS(RETRO_DEVICE_MOUSE, 3)`, driven by **relative mouse X** -- the established libretro spinner convention, per the spec's decision. `RETRO_DEVICE_ANALOG` as an alternate source is explicitly out of scope (spec Q4) and belongs with #439's shared analog layer.

**No titledb row, ever** (spec section 4.7): selecting a rotary removes Up and Down, so a per-title default would break menu navigation for anyone using a pad. That is a functional break, not a preference. Nothing in `src/core/titledb.c` is touched.

**Direction** is a host-side UX convention, stated as such rather than dressed up as a hardware fact: host +X maps to clockwise, because pushing a spinner right turns the knob clockwise. TR10 fixes the hardware half -- anticlockwise is A-leads-B, increasing Gray index. There is deliberately **no invert option**; an unsourced sign knob is how a real wiring bug gets papered over instead of found.

`update_input()` grows a third branch. A rotary port keeps **A, B, C, Option, Pause and the keypad** from the retropad and withholds only U/D/L/R, because those four lines *are* the encoder. This is not a nicety: Tempest 2000's rotary support is hidden behind an unlock requiring **Pause on both controllers simultaneously** from the Options screen, persisted in the game's EEPROM. Zero the whole array for a rotary port and the user can never turn the feature on at all. It has its own assertion below.

---

### `test/tools/rotary_decode_test` (new, wired into `make test`)

Shaped like A2's `mouse_decode_test`: synthetic rotation in, decoded direction and magnitude out, **both directions**, **both ports**. The decoder is written independently of `src/jerry/quadrature.c`, from TR10's phase sequences, and discards two-state jumps the way Domin's table does -- so any emitted jump shows up as `dropped` and fails the run.

```
  ok   InputDevSetType accepts a rotary on port 1
  ok   InputDevSetType accepts a rotary on port 2
port 1 rotary: direction and distance
  ok   +40 host units over 40 polls decode as clockwise (net=-40, expect -40), dropped=0
  ok   -40 host units over 40 polls decode as anticlockwise (net=40, expect +40), dropped=0
  ok   4 units/poll x 30 polls lags rather than jitters (net=-30, expect -30), dropped=0
port 1 rotary: poll shapes (the emission rule)
  ok   shape C (write + TWO reads) decodes as shape A: A=-40 C=-40, dropped A=0 C=0
  ok   shape B (write once, then read-only) is frozen by design: net=0 (expect 0)
port 1 rotary: visible in row 0 ONLY (not row-blind)
  ok   phase lines assert in row 0 ($FBFF) and in NO other row (rows 1-3: $FFFD $FFFB $FFF7)
  ok   rows 1-3 still decode as a standard pad (keypad '1' pulls bit 11 low in row 1: $F7FD)
  ok   Up and Down never assert on a rotary port
port 1 rotary: Tempest 2000's measured 8-write scan
  ok   one clean count per scan across 40 scans (net=-40, expect -40), dropped=0 -- a per-read clock would decode garbage here
port 1 rotary: controller-type identification
  ok   with the diode absent (default) C3 reads 1 -- standard joypad ($F14002=$FF7F bit 0)
  ok   with the diode fitted C2 C3 = 1 0 in the right rows for this port (row 3 bit 0 = 0 -> $FF7E; row 2 bit 0 = 1 -> $FF7F)
port 2 rotary: direction and distance
  ok   +40 host units over 40 polls decode as clockwise (net=-40, expect -40), dropped=0
  ok   -40 host units over 40 polls decode as anticlockwise (net=40, expect +40), dropped=0
  ok   4 units/poll x 30 polls lags rather than jitters (net=-30, expect -30), dropped=0
port 2 rotary: poll shapes (the emission rule)
  ok   shape C (write + TWO reads) decodes as shape A: A=-40 C=-40, dropped A=0 C=0
  ok   shape B (write once, then read-only) is frozen by design: net=0 (expect 0)
port 2 rotary: visible in row 0 ONLY (not row-blind)
  ok   phase lines assert in row 0 ($BF7F) and in NO other row (rows 1-3: $FFBF $FFDF $FFEF)
  ok   rows 1-3 still decode as a standard pad (keypad '1' pulls bit 15 low in row 1: $7FBF)
  ok   Up and Down never assert on a rotary port
port 2 rotary: Tempest 2000's measured 8-write scan
  ok   one clean count per scan across 40 scans (net=-40, expect -40), dropped=0 -- a per-read clock would decode garbage here
port 2 rotary: controller-type identification
  ok   with the diode absent (default) C3 reads 1 -- standard joypad ($F14002=$FF7F bit 2)
  ok   with the diode fitted C2 C3 = 1 0 in the right rows for this port (row 2 bit 2 = 0 -> $FF7B; row 3 bit 2 = 1 -> $FF7F)
Tempest 2000 unlock: Pause on BOTH controllers at once
  ok   rotary on port 1 + pad on port 2 both report Pause ($F14002=$FF7A, p1=1 p2=1)
a rotary on one port does not perturb the other
  ok   port-1 direction bits unchanged by a port-2 rotary ($0F00 vs $0F00), while port 2 decoded net=-25
the mouse is still row-blind (the two shapes stay distinct)
  ok   a port-2 mouse asserts in ALL FOUR rows ($2F7F $2FBF $2FDF $2FEF) where the rotary asserts in one
savestate round-trip mid-rotation
  ok   retro_serialize succeeded mid-rotation
  ok   retro_unserialize accepted the state
  ok   post-restore replay matches (net -25 vs -25, dropped 0 vs 0) -- phase, backlog and Q8 carry all survived the rollback
no device selected -> nothing reaches the matrix
  ok   both ports report pad and nothing is attached
  ok   no rotation reaches $F14000 (net=0)

  PASS: [rotary_decode] 0 failure(s)
```

The two assertions that carry the most weight:

- **row-0-only visibility, asserted both ways** -- the phase lines assert in row 0 and in no other row, *and* rows 1-3 still decode a poked keypad digit. That pair is what fails if the rotary is ever "simplified" onto the mouse's row-blind overlay. `test_mouse_still_row_blind` asserts the converse on the same build, so the two shapes cannot quietly converge.
- **the measured eight-write scan**, replayed literally, decoding one clean count per scan. That is the regression for the defect above.

### C89 lint

```
$ bash scripts/c89-lint.sh src/jerry/inputdev.c src/jerry/joystick.c src/jerry/quadrature.c libretro.c
C89 lint passed
```

### Build

`DEVELOPER_DIR=/Library/Developer/CommandLineTools make -j$(getconf _NPROCESSORS_ONLN)` -- clean, no warnings.

### Full suite, with the private ROM corpus symlinked in

```
$ DEVELOPER_DIR=/Library/Developer/CommandLineTools make -j10 TEST_EXPORTS=1 test
TEST_EXIT=0
...
```

`Skipped checks: none`, so the audio, CD and determinism rows genuinely ran rather than silently skipping. The symlink was removed afterwards; its target was not touched.

### Production ABI unchanged

```
      46
$ nm -gU virtualjaguar_libretro.dylib | grep -c "_retro_"
46
$ nm -gU virtualjaguar_libretro.dylib | grep -c "InputDev\|Quad\|Joystick"
0
```

46 exported symbols, all `retro_*`, zero internals leaked -- matching A2's measurement. The `InputDev*` / `Quad*` / `Joystick*` exports exist only in the wide test ABI.

---

## What is NOT verified

Stating this plainly, as A2 did.

- **No real-frontend test.** Everything above is headless. The rotary has not been driven from RetroArch with an actual mouse or spinner.
- **Tempest 2000 has not been played.** The spec names the section 4.4 two-controller Pause unlock as step 7's acceptance test: perform the unlock, set `P1 ROTARY`, confirm play, then confirm the P2 toggle. `test_pause_unlock` asserts the machine-level precondition (a rotary port and a pad port can both report Pause in the same read), but it cannot tell you the menu item appears or that the spinner feels right.
- **Direction sense is unconfirmed against the game.** TR10 fixes anticlockwise = A-leads-B; which way Tempest 2000 wants the ship to travel is a question only playing it answers. If it is backwards, the fix is the one negation in `InputDevFeed()`.
- **Per-port togglability for P2 is inferred**, from the menu structure rather than observed in the source screenshots.
- **Sensitivity default of 100% is untuned.** A real rotary has no canonical pulses-per-revolution -- Atari's never shipped, so every unit is an aftermarket build (36 PPR detentless, 30 detents and 24 detents are all attested). The default is a starting point, not a measurement.
- **The eight other rotary titles** (Downfall+, Impulse X, Kobayashi Maru, Project W, Rebooteroids, Virtual Xperience's Pong, Kaboom, the Arkanoid ST port) are not in the private corpus and none was exercised. Only Tempest 2000's poll shape is measured; the row-gated rule is shape-independent by construction, but that is an argument, not a test.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -62,6 +62,7 @@ test/tools/netlink_pair
 test/tools/netlink_game
 test/tools/joymatrix_identity
 test/tools/mouse_decode_test
+test/tools/rotary_decode_test
 # vjtrace flight-recorder analyzer/smoke tools -- built on demand by
 # test/tools/vjtrace_selftest.sh (and by hand per each tool's own header
 # comment); never tracked, same Mach-O-on-Linux-CI hazard as above.
--- a/Makefile
+++ b/Makefile
@@ -880,6 +880,7 @@ clean:
 		test/tools/test_frame_timing test/tools/test_runahead_determinism test/tools/test_pertitle_db \
 		test/tools/test_wedge_spin test/tools/i2s_lag_probe \
 		test/tools/joymatrix_identity test/tools/mouse_decode_test \
+		test/tools/rotary_decode_test \
 		test/test_quadrature \
 		test/.skipped-checks
 
@@ -934,7 +935,8 @@ test: test/test_dram_timing test/test_cheat test/test_event_queue test/test_jlin
 		test/tools/test_memory_map test/tools/test_op_gpu_object test/tools/test_option_visibility test/test_memtrack test/test_nvmbios test/test_uart_core test/test_netlink_host \
 		test/tools/netlink_pair test/tools/netlink_latency test/tools/netlink_delay_proxy test/tools/test_pertitle_db \
 		test/tools/i2s_lag_probe test/tools/joymatrix_identity \
-		test/test_quadrature test/tools/mouse_decode_test
+		test/test_quadrature test/tools/mouse_decode_test \
+		test/tools/rotary_decode_test
 	@# Skip ledger: truncate FIRST so a previous run's rows cannot resurface
 	@# as fresh skips (the stale-row failure mode documented for
 	@# cd_boot_matrix.sh).  Every optional check below records into it, and
@@ -1175,6 +1177,13 @@ test: test/test_dram_timing test/test_cheat test/test_event_queue test/test_jlin
 	@# and distance out, all three wiring cases and all three poll shapes,
 	@# plus row-blindness and the v12 savestate round-trip.
 	./test/tools/mouse_decode_test ./$(TARGET) test/roms/yarc.j64 --quiet
+	@# Tempest rotary end-to-end (#436): synthetic rotation in, decoded
+	@# direction and magnitude out in BOTH directions on BOTH ports, the
+	@# row-0-only visibility that distinguishes a matrix device from the
+	@# row-blind mouse, Tempest 2000's measured 8-write scan, the
+	@# two-controller Pause unlock, the C2/C3 ID bits and a savestate
+	@# round-trip mid-rotation.
+	./test/tools/rotary_decode_test ./$(TARGET) test/roms/yarc.j64 --quiet
 	./test/test_memtrack
 	./test/test_nvmbios
 	@# Option visibility is content-type dependent; the disc half runs only
@@ -1496,6 +1505,13 @@ test/tools/mouse_decode_test: test/tools/mouse_decode_test.c \
 		test/harness/harness.c test/harness/harness.h
 	$(CC) -O2 -Wall -std=c99 $(INCFLAGS) \
 		-o $@ test/tools/mouse_decode_test.c \
+		test/harness/harness.c \
+		$(if $(filter Linux,$(shell uname -s)),-ldl) -lm
+
+test/tools/rotary_decode_test: test/tools/rotary_decode_test.c \
+		test/harness/harness.c test/harness/harness.h
+	$(CC) -O2 -Wall -std=c99 $(INCFLAGS) \
+		-o $@ test/tools/rotary_decode_test.c \
 		test/harness/harness.c \
 		$(if $(filter Linux,$(shell uname -s)),-ldl) -lm
 
--- a/libretro.c
+++ b/libretro.c
@@ -201,11 +201,19 @@ static uint8_t *joypad_buttons[2] = { joypad0Buttons, joypad1Buttons };
  * RETRO_DEVICE_MOUSE subclass because it is a relative-motion pointer;
  * the three subclasses are the three adapter/mouse wiring combinations
  * (docs/jaguar-mouse-adapter-mapping.md section 4d), not three different
- * devices.  Port 2 only -- see inputdev.h. */
+ * devices.  Port 2 only -- see inputdev.h.
+ *
+ * The rotary (#436) is also a RETRO_DEVICE_MOUSE subclass, driven by
+ * relative X: that is the established libretro spinner convention and it
+ * covers real spinner hardware.  RETRO_DEVICE_ANALOG as an alternate
+ * rotary source is deliberately out of scope here (design spec Q4) and
+ * belongs with #439's shared analog layer.  Rotary is offered on BOTH
+ * ports; unlike the mouse it is a genuine matrix device (inputdev.h). */
 #define RETRO_DEVICE_JAG_PAD             RETRO_DEVICE_JOYPAD
 #define RETRO_DEVICE_JAG_MOUSE_ST        RETRO_DEVICE_SUBCLASS(RETRO_DEVICE_MOUSE, 0)
 #define RETRO_DEVICE_JAG_MOUSE_AMIGA     RETRO_DEVICE_SUBCLASS(RETRO_DEVICE_MOUSE, 1)
 #define RETRO_DEVICE_JAG_MOUSE_AMIGA_AD  RETRO_DEVICE_SUBCLASS(RETRO_DEVICE_MOUSE, 2)
+#define RETRO_DEVICE_JAG_ROTARY          RETRO_DEVICE_SUBCLASS(RETRO_DEVICE_MOUSE, 3)
 
 /* Device the frontend explicitly selected via
  * retro_set_controller_port_device, or INPUTDEV_PAD when it never did.
@@ -218,6 +226,12 @@ static bool         port_device_forced[2]   = { false, false };
  * changes rather than on every check_variables() call. */
 static InputDevType port_device_active[2]   = { INPUTDEV_PAD, INPUTDEV_PAD };
 static bool         show_mouse_options      = true;
+static bool         show_rotary_options     = true;
+/* Sensitivity ladders, Q8 (256 == 1.0).  Kept separate because a spinner
+ * and a mouse want very different multipliers; the per-port scale is
+ * whichever one matches the device actually attached to that port. */
+static int32_t      mouse_scale_q8          = 256;
+static int32_t      rotary_scale_q8         = 256;
 
 /* Has this port actually received non-zero mouse state from the frontend?
  *
@@ -234,6 +248,17 @@ static bool         show_mouse_options      = true;
  * device type changes, which clears this): a mouse that has moved once is
  * a mouse the frontend is routing, and letting a pad drive the same six
  * lines at that point is the ambiguity the suppression exists to prevent.
+ *
+ * A ROTARY PORT DELIBERATELY DOES NOT USE THIS, and that is not an
+ * oversight to "fix" later.  The rule exists because a mouse port loses
+ * BOTH devices; a rotary port loses neither -- it withholds only U/D/L/R
+ * and keeps A/B/C/Option/Pause and the keypad on the RetroPad, so a
+ * frontend that routes no mouse state still leaves the port usable.  (In
+ * particular Tempest 2000's rotary unlock, Option then Pause on both
+ * controllers, is entirely buttons and works with no spinner routed.)
+ * Deferring the rotary's four-slot withholding would instead mean an
+ * un-serialized flag that test_savestate's exact-replay assertion has to
+ * reason around, for a failure mode that cannot occur.
  *
  * DELIBERATELY NOT SERIALIZED, for the same reason inputdev.h gives for
  * the device type and the sensitivity scale: this is a fact about how the
@@ -258,9 +283,17 @@ static const char *inputdev_type_name(InputDevType t)
       case INPUTDEV_MOUSE_ST:            return "Atari ST / PS2 mouse";
       case INPUTDEV_MOUSE_AMIGA_ADAPTER: return "Amiga mouse (Amiga adapter)";
       case INPUTDEV_MOUSE_AMIGA_ON_ST:   return "Amiga mouse (ST adapter)";
+      case INPUTDEV_ROTARY:              return "Tempest rotary";
       default:                           break;
    }
    return "standard joypad";
+}
+
+static bool inputdev_is_mouse_type(InputDevType t)
+{
+   return (t == INPUTDEV_MOUSE_ST
+           || t == INPUTDEV_MOUSE_AMIGA_ADAPTER
+           || t == INPUTDEV_MOUSE_AMIGA_ON_ST);
 }
 
 static void apply_port_device(int port, InputDevType type)
@@ -487,12 +520,33 @@ static bool update_option_visibility(void)
    {
       bool show_mouse_prev = show_mouse_options;
 
-      show_mouse_options = (InputDevGetType(1) != INPUTDEV_PAD);
+      show_mouse_options = inputdev_is_mouse_type(InputDevGetType(1));
 
       if (show_mouse_options != show_mouse_prev)
       {
          option_display.visible = show_mouse_options;
          option_display.key     = "virtualjaguar_mouse_sensitivity";
+         environ_cb(RETRO_ENVIRONMENT_SET_CORE_OPTIONS_DISPLAY,
+                    &option_display);
+         updated = true;
+      }
+   }
+
+   /* Rotary sensitivity and the rotary controller-type knob mean nothing
+    * unless a rotary is attached to one port or the other (#436). */
+   {
+      bool show_rotary_prev = show_rotary_options;
+
+      show_rotary_options = (InputDevGetType(0) == INPUTDEV_ROTARY
+                             || InputDevGetType(1) == INPUTDEV_ROTARY);
+
+      if (show_rotary_options != show_rotary_prev)
+      {
+         option_display.visible = show_rotary_options;
+         option_display.key     = "virtualjaguar_rotary_sensitivity";
+         environ_cb(RETRO_ENVIRONMENT_SET_CORE_OPTIONS_DISPLAY,
+                    &option_display);
+         option_display.key     = "virtualjaguar_rotary_id";
          environ_cb(RETRO_ENVIRONMENT_SET_CORE_OPTIONS_DISPLAY,
                     &option_display);
          updated = true;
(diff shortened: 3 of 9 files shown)


JoeMatt:
## Rebuilt on the current develop, and one claim in it was wrong

Rebased onto `ecfba36`, force-pushed. This branch previously carried its own **pre-review** copy of the mouse implementation from before #449 was reviewed — merging it would have resurrected the old `inputdev.c`/`libretro.c` on top of the merged, fixed ones. So the rotary delta was re-applied onto develop's files rather than merged, and every #449 review fix survives.

### Integration decisions where develop's fixes met the rotary code

- **The clock gate moved from `joystick.c` into `inputdev.c`.** develop had `if (offset1 != 0xFF) InputDevClock();`. A port-1 rotary must advance on exactly the reads that test rejects, so `joystick.c` now calls unconditionally and passes both decoded rows; the mouse's gate is preserved verbatim as `row[p] != 0xFF` (any row — it is row-blind) and the rotary's is `row[p] == 0`.
- **`p1_device_from_option()` added**, and used by `retro_set_controller_port_device`. develop hardcoded `INPUTDEV_PAD` for port 1, which was right only while port 1 was pad-only; with a rotary reachable there, a frontend's routine post-load JOYPAD assignment would re-detach it for the session — the same bug #449's fix A cured for port 2, on a newly reachable path.
- **develop's pre-multiply `> 1600` sensitivity clamp had been silently dropped by the old PR diff.** Restored, and applied to the rotary ladder.
- **`QUAD_REST_PHASE` is develop's 2**, which matters here: an idle rotary writes `0x00` to all four direction slots and is bit-identical to an idle pad. Under the old snapshot's phase 0 it would have read as Left+Right held.
- **`inputdev_live` stays mouse-only, deliberately.** The deferred-suppression rule exists because a mouse port loses *both* devices; a rotary port keeps A/B/C/Option/Pause and the keypad, so that failure mode cannot occur — and T2K's unlock is all buttons, so it works with no spinner routed.

### A claim I checked and it did not hold

The rebase reported that per-device arm consumption (rather than an unconditional clear) was necessary, and that `joymatrix_identity` proved it. **It does not.** Reverting to the unconditional clear leaves all four digests unchanged and `mouse_decode_test` passing.

Chasing that produced a better answer than the original claim: the described bug is **unreachable**. The row select lives in `joystick_ram[1]`, its only writer is `JoystickWriteWord` at offset 0, and that path arms unconditionally — so the arm is always set at the first read after any row change. Measured, not argued: **2.4M random write/read sequences across six device combinations digest bit-identically with and without the guard.**

So `if (consumed)` guards nothing reachable today. It stays, because it is the honest statement of the contract — an arm permits one *decodable sample*, not one bus read, and the next device that samples on some other condition makes the difference real — but the comment claiming a live timing bug is gone. It is now pinned by `test_arm_is_spent_only_by_a_decodable_read`, which reaches the contract by calling `InputDevArm()`/`InputDevClock()` directly because no aligned bus sequence can, and whose header records that measurement so nobody "simplifies" it into a bus-only test.

That test is a real regression test, not a decorative one: under the sabotage it reports `$F14000 bits 12-15 $D -> $D, expect a change`, and each case is paired with its converse so deleting the clear outright also fails.

Also corrected a pre-existing wrong comment: `$81F7`'s low nibble 7 is port-1 socket-0 **row 3**, not row 0 (`joypad0Offset[0x7] == 0x0C`; row 0 is nibble `0xE`). Harmless where it sits, load-bearing for anyone adding a port-1 rotary case.

### Verification on the rebased tip

```
make -j8 TEST_EXPORTS=1 test   ->  exit 0, 1345 PASS, 0 FAIL
                                   "Skipped checks: none -- every optional check ran"
joymatrix_identity  4/4   0xC24DDCDE / 0xD0CD02E2 / 0xD0CD02E2 / 0x5E1858EA  (unchanged)
rotary_decode_test  32 ok
mouse_decode_test   arm cases ok, both converses ok
```

Reverse mode flip (`TEST_EXPORTS=1` → plain `make`) also checked, since `inputdev.c` gained an `#include "joystick.h"` — a new include-graph edge. C89 lint clean. Savestate untouched: `INPUTDEV_STATE_SIZE` stays 41, `STATE_VERSION` stays 12.

### Known limits — please read before merging

- **No hardware or real-game confirmation.** No Tempest 2000 run, no spinner, no RetroArch session. Every rotary assertion is against a decoder written inside the test from TR10, so `+X = clockwise` and the C2/C3 row-by-port mapping are self-consistent with that reading but **not silicon-confirmed**. 32 green checks are internal consistency.
- **`virtualjaguar_rotary_id` is one option driving both ports**, so a two-rotary setup cannot have one diode fitted and one not. That is the design, and the option text is honest that T2K ignores the bits — flagging it so it is not misread as an oversight.
- **These commits are unsigned.** The signing agent cannot prompt without a TTY. If signed commits are required on merge, this branch needs re-signing from an interactive shell.
- Still no bot review: Copilot credits exhausted, Kimi out of quota for the cycle.

PR #455, opened 2026-08-15
fix(ci): say on the PR when no review was posted

JoeMatt:
## The problem this fixes

The Kimi reviewer soft-fails on purpose — an advisory review must never red-X a PR that compiles and passes its tests. The consequence is that a run which posts **nothing** still goes green, and the only signal was a `::warning::` in the job log.

That is not a theoretical gap. #447, #450 and #451 sat unreviewed for a day because the plan hit its quota:

```
HTTP 403: {"error":{"message":"You've reached your usage limit for this
billing cycle...","type":"access_terminated_error"}}
##[warning]HTTP 403 from the Kimi API; no review was posted.
```

The workflow correctly diagnosed it in the log. But on the PR — where the decision to merge is actually made — an unreviewed PR and a PR the reviewer had nothing to say about look identical. All three read as clean.

## What changes

**Persistent failures now leave a note on the PR.** Quota, auth, and request-shape failures post a short marker-tagged comment saying no review happened and why. It is sticky: `pr_note()` finds the existing comment by an invisible HTML marker and PATCHes it, because every push re-runs the reviewer and a reviewer that is down stays down for the whole billing cycle — a fresh comment per push would bury the thread.

**Transient failures deliberately do not post.** A DNS blip, a refused connection, a timeout — the `::warning::` still covers those. A comment for each would be noise, and unlike a quota limit they usually fix themselves by the next push.

**Two smaller silent paths closed:**

- a failed `gh pr comment` *after* a successful review generated the entire review and then threw it away, printing one unannotated line — the last remaining way for a green run to produce nothing. It now annotates, and names the likely cause (the workflow's `pull-requests: write`).
- a successful review retracts any earlier unavailable-notice, so a PR cannot end up carrying both a review and a claim that it has none.

`pr_note()` never raises. It runs on the failure path of an advisory job, and a broken notice must not become a second, louder failure — every step is wrapped and degrades to a log line.

## Testing

`python3 -m py_compile` clean; `pr_note_body()` exercised against the three real payloads this repo has actually seen — the quota 403, an auth 401, and the `invalid temperature` 400 — and each routes to its own wording rather than to a generic one.

Not exercised end-to-end against the live API, because the quota that motivated the fix is still exhausted. The first real exercise will be the next run on any open PR, which should post exactly one notice.

Closes nothing; follows #452.

### DIFF
--- a/.github/scripts/kimi_review.py
+++ b/.github/scripts/kimi_review.py
@@ -123,6 +123,76 @@ def call_kimi(base_url, key, key_id, model, system, user):
     return payload["choices"][0]["message"]["content"]
 
 
+MARKER_FMT = "<!-- kimi-review:{} -->"
+
+
+def pr_note(repo, pr, marker, body):
+    """Upsert a marker-tagged comment on the PR.
+
+    Sticky on purpose: every push to an open PR re-runs the reviewer, and a
+    reviewer that is down stays down for the whole billing cycle, so posting
+    a fresh comment per push would bury the PR under identical notices.  The
+    marker is an HTML comment, invisible in the rendered thread.
+
+    Never raises: this runs on the failure path of an advisory job, and a
+    broken notice must not become a second, louder failure.
+    """
+    tag = MARKER_FMT.format(marker)
+    try:
+        r = subprocess.run(
+            ["gh", "api", f"repos/{repo}/issues/{pr}/comments",
+             "--paginate", "--jq",
+             '.[] | select(.body | contains("%s")) | .id' % tag],
+            capture_output=True, text=True, timeout=60,
+        )
+        existing = (r.stdout or "").split()
+        p = "/tmp/kimi-note.md"
+        with open(p, "w") as f:
+            f.write(tag + "\n" + body)
+        if existing:
+            cmd = ["gh", "api", "--method", "PATCH",
+                   f"repos/{repo}/issues/comments/{existing[0]}",
+                   "-F", "body=@" + p]
+        else:
+            cmd = ["gh", "pr", "comment", pr, "--repo", repo,
+                   "--body-file", p]
+        n = subprocess.run(cmd, capture_output=True, text=True, timeout=60)
+        if n.returncode != 0:
+            print(f"could not post the unavailable-notice: {n.stderr}")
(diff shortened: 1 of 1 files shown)


JoeMatt:
## Correction: this PR's commit message names the wrong cause

The fix is right and worth keeping. The diagnosis in its commit message is not, and I'd rather correct it here than leave a wrong cause recorded in `git log` for the next person who touches this workflow.

I wrote that #450's review job failed because GitHub computes `refs/pull/N/merge` asynchronously and a force-push sends mergeability back to UNKNOWN, so the job reached checkout before the ref existed. The timeline says otherwise:

```
05:47:54Z  run 31867836534 created (pull_request event, sha a8c73b8)
05:48:09Z  PR #450 merged                      <- 15 seconds later
05:50:18Z  checkout: fatal: couldn't find remote ref refs/pull/450/merge
```

The merge ref was gone because **the PR had already been merged and its branch deleted** while the job sat queued — not because mergeability had not been computed yet. Same error string, different cause, and I asserted the one that fit my prior instead of checking the timestamps.

That is the third time an error diagnosis in this exact file has claimed a cause its evidence contradicted (the first two are recorded in `.github/scripts/kimi_review.py`'s handler comments: every 400 blamed on the model name when it was `temperature`, then the first fix gating on `"model" in detail` when the temperature message contains that word). The lesson the file already carries — never let a diagnostic guess louder than its evidence — applies to commit messages too.

**What this means for the fix:**

- **`continue-on-error` on the job is the part that actually addresses what happened**, and it addresses it correctly. A merged-and-deleted PR, a queued job, a runner hiccup — none of them should red-X a PR, and all of them landed before `kimi_review.py` could soft-fail them.
- **The poll-and-fall-back step is still worth having**, but it is now unproven against a real occurrence rather than a fix for an observed one. The race it guards is real in principle; I have no evidence it has ever fired here. Note it also cannot rescue the merged-and-deleted case — if the branch is gone, `refs/pull/N/head` is gone too, and `continue-on-error` is what catches that.
- It follows that the 100s of polling is dead time in the one scenario we have actually seen. Not worth another PR on its own; worth knowing before someone tunes it.

No code change proposed — the shipped behaviour is correct. This is to keep the record honest.

Issue #462, opened 2026-08-15
[bug] Strange file size limit.


42Bastian:
### Game / ROM

Homebrew / Bootintro

### Steps to reproduce

Load a ROM < 128k

### Expected vs actual

Expected: Load and run.
Actual: Shows error.

### BIOS mode

Real BIOS (built into the core; runs the actual boot sequence)

### Blitter mode

Accurate (Midsummer2)

### Core version

3.3.0

### Frontend + platform

RA 1.22.2

### Logs

```shell

```

### Pre-flight

- [x] Verified the issue is in this libretro core (not standalone Virtual Jaguar or another emulator).
- [x] Searched [existing issues](https://github.com/libretro/virtualjaguar-libretro/issues?q=is%3Aissue) for duplicates.

JoeMatt:
Board sync 2026-08-15: parking in Maintenance as size S. Likely `ParseFileType` in `src/core/file.c` — `JST_ROM` only on `size % 1MiB == 0 || size == 131072`, then a strict `InferRawBinaryLoadAddress` fallback. Small headerless homebrew/bootintros die. Pulling forward as a cheap user-facing fix, unrelated to the timing epic.

JoeMatt:
Fix in https://github.com/libretro/virtualjaguar-libretro/pull/465 — small carts with the `$400` universal header, plus headerless images between `$408` and 1 MiB, now load as `JST_ROM`.

JoeMatt:
Shipped in PR #465: small-ROM load is on `libretro/develop`. Closing this issue.

#469 (Fountain) is a different bug: 128K jagcrypt BIOS crash, not a load reject.

PR #465, opened 2026-08-15
fix(cart): load homebrew carts smaller than 128K (#462)

JoeMatt:
## Summary
- `ParseFileType` only accepted `JST_ROM` at 1 MiB multiples or exactly 128K, so bootintros and small homebrew (issue #462, 42Bastian) were refused.
- A universal header at `$400` is now enough to call the file a cartridge at any size ≥ `$408`.
- Headerless images between `$408` and 1 MiB that are not ABS/JagServer/Alpine/BJL still map as a cart so real-BIOS mode can boot them from `$800000`. 512-byte copier headers and non-megabyte leftovers ≥ 1 MiB stay rejected.

## Test plan
- [x] `bash scripts/c89-lint.sh src/core/file.c`
- [x] `make TEST_EXPORTS=1` + `./test/test_cart_format` — 8 passed (3 new: 64K headered, 4K headered, 2K headerless)
- [ ] Load a <128K homebrew/bootintro in RetroArch with Real BIOS

### DIFF
--- a/src/core/file.c
+++ b/src/core/file.c
@@ -33,14 +33,20 @@
  * whole-file CRC ($0FDCEB66) is catalogued as FF_BAD_DUMP.  Nothing about the
  * image data is wrong, so skip the header instead of refusing the file.
  *
- * Detection is deliberately narrow, because the size test alone would start
+ * Detection is deliberately narrow, because a size test alone would start
  * accepting arbitrary junk.  Both must hold:
  *
- *   - the file overhangs a whole number of megabytes by exactly the header
- *     length, and
+ *   - the file is large enough to hold a 512-byte header plus the payload's
+ *     $400 marker slot, and
  *   - the cartridge universal-header marker ($04040404, which precedes the run
  *     address that JaguarLoadFile reads from ROM offset $404) is present at
  *     that offset measured from the payload rather than from the file.
+ *
+ * A native cart already has the marker at file offset $400, so those are
+ * left alone even if $04040404 also appears at $600.  The payload no longer
+ * has to be a whole number of megabytes: a sub-1 MiB homebrew with the same
+ * 512-byte copier header used to fall through to the headerless JST_ROM
+ * fallback and load skewed (Kimi review on #465).
  */
 #define CART_HEADER_SKIP_SIZE   512
 #define CART_UNIVERSAL_MARKER_OFFSET   0x400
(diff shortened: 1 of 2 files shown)


JoeMatt:
## Kimi review

**Finding 1 (truncated commercial dumps via the headerless fallback) — accepted as a named trade-off, with a correction on which path is responsible.**

A truncated 1 MiB commercial dump almost always still has `$04040404` at file `$400`, so it is accepted by the *marker* check (the `#462` path), not by the headerless `[$408, 1MiB)` fallback. The fallback only runs for files that have no cart header at all. CRC cannot catch a truncation: the truncated file has a different CRC, so `FindRomIdentifier` will not match the good dump. Leaving the fallback in place so headerless bootintros boot; files `>= 1 MiB` that are not an exact megabyte multiple still hit `ReportUnrecognizedContent`.

**Finding 2 (sub-1 MiB + 512-byte copier header) — fixed in 7485fe4.**

`DetectPrependedHeaderSize` no longer requires a whole-megabyte payload. It skips 512 bytes when the marker is at payload `$400` (file `$600`) and *not* already at file `$400` (so a native cart is never stripped). New test `small_headered_with_copier_header_loads`: 64K+512 → `jaguarROMSize == 65536`. Cart-format suite 9/9.

JoeMatt:
## Kimi follow-up

**Finding 1 (lost size-shape guard) — agreed, fixed.**

The previous follow-up keyed only on the `$600` marker, which put every file ≥ 1540 bytes in the candidate pool. Copier output is always `N MiB + 512`, so an **exact-megabyte** file is never a headered dump and is now excluded before any payload math. The sub-1 MiB homebrew arm stays for `64K+512` etc.

A first draft of that guard used `payload = size - 512` and then `payload < 1 MiB`, which still treated a 1 MiB native as a "small payload" (`1 MiB - 512`). The regression test `fullsize_incidental_marker_at_600_not_stripped` caught that; the guard is now on **file** size (`size % 1 MiB == 0 → never strip`). Cart-format suite 10/10.

**Finding 2 (payload `$200` coincidental marker refuses a real strip) — accepted as the same likelihood class as before.**

A copier-headered image whose payload happens to contain `$04040404` at `$200` (file `$400`) is not stripped on the sub-1 MiB arm. The classic `payload % 1 MiB == 0` arm does not consult file `$400`, so Brutal Sports is unchanged. Not tightening further.

Issue #461, opened 2026-08-15
Crash watchdog named the wrong subsystem: gpu_pc_valid too permissive, video_stall fires on static screens


JoeMatt:
Two independent defects in `src/core/crash_detect.c`, both exposed by the Defender 2000 investigation (#460). Together they turned a GPU executing game *data* in an infinite loop into a log line that names the wrong subsystem.

The user's line was:

```
[CRASH-DETECT] video_stall frame=872 fb_hash=$5C3A6B17 unchanged for 300 frames
               gpu_pc=$0014B1CE gpu_run=1 dsp_pc=$00F1B332 dsp_run=1
```

The GPU had run away into a data buffer and was spinning in a `STORE`/`JR` fragment of ordinary game data. Nothing said so.

## 1. `gpu_pc_valid()` accepts anything <= `MAPPED_CODE_HI` (`$E3FFFF`)

So a GPU spinning inside a data buffer in main RAM is "valid" and never raises `gpu_pc_escape`. `$0014B1CE` is exactly that. The check exists to catch a GPU that has left its local RAM, but the legal range is so wide that the most common runaway destination — main RAM — is inside it.

It is also **alias-dependent**, which is the sharper bug: `$212786` is accepted while `$FD012786` — *the same loop, through a different alias* — would be rejected. The same fault is a clean bill of health at one alias and an escape at another.

**Proposed fix.** `VJT_EV_GPU_GO` already records `G_PC` at every GPU start, so the set of addresses the GPU was ever legitimately started at is known. Add a `gpu_runaway` signature: running with PC outside local RAM **and** outside any region the GPU was ever started at. Alias-normalise `gpu_pc_valid()` so the same address cannot be valid at one alias and an escape at another.

**A "run of `$0000` opcodes" predicate would NOT work** and is worth ruling out explicitly: the GPU walks through zeros only in transit. Its resting state is a `STORE`/`JR` loop inside real game data, which is indistinguishable from code by any opcode-shape test.

## 2. `video_stall` fires on healthy static screens

It fired at **default options** on the Defender 2000 title screen at frame 575 — the same shape as the user's frame-872 line. So the signature that did fire carries almost no information: it cannot distinguish "the game is showing a static screen" from "the machine is wedged".

[...]

PR #466, opened 2026-08-15
fix(core): gpu_runaway watchdog + quieter video_stall (#461)

JoeMatt:
## Summary
- \`gpu_pc_valid()\` now masks to 24-bit, so \`$FD012786\` and \`$012786\` are the same address.
- New \`gpu_runaway\` signature: GPU running outside local RAM **and** outside any 4 KB page it was started at (\`CrashDetectNoteGPUGo\` on G_CTRL GO). Names the Defender 2000 3x data-buffer spin instead of blaming video.
- \`video_stall\` no longer fires on a still image while the GPU/DSP is executing in local RAM (AvP idle, pause screens).

## Test plan
- [x] \`bash scripts/c89-lint.sh src/core/crash_detect.c src/tom/gpu.c\`
- [x] \`make TEST_EXPORTS=1\`
- [ ] Defender 2000 + \`m68k_clock_scale=3x\`: log should show \`gpu_runaway\` rather than a lone \`video_stall\`
- [ ] AvP idle / a pause screen should not \`video_stall\`

### DIFF
--- a/src/core/crash_detect.c
+++ b/src/core/crash_detect.c
@@ -172,10 +172,12 @@ static int gpu_pc_valid(uint32_t pc)
 
 static int dsp_pc_valid(uint32_t pc)
 {
-   /* High-byte garbage aliases on the 24-bit bus (JTRM): $FD012786
-    * fetches from $012786.  Treating that as an escape was a false
-    * positive; the fetch is mapped.  DSP has no gpu_runaway twin --
-    * a Defender-style jump into a data buffer is still invisible here. */
+   /* High-byte garbage aliases in this core: GPUReadWord / DSP fetch
+    * and the 68K bus path all do `addr &= 0x00FFFFFF` (gpu.c:331,
+    * jaguar.c m68k_read_memory_*).  $FD012786 fetches from $012786.
+    * Treating that as an escape was a false positive against our own
+    * decode.  DSP has no gpu_runaway twin -- a Defender-style jump
+    * into a data buffer is still invisible here. */
    pc = pc_canonical(pc);
    if (pc <= MAPPED_CODE_HI) return 1;
    if (dsp_pc_in_local(pc)) return 1;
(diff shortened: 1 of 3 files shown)


JoeMatt:
## Kimi review

**Finding 1 (4 KB start-page only / silent table-full) — fixed in 685abe1.**

`gpu_pc_in_start_page` now accepts `[start_page, start_page + 64 KB)`. That covers a program that flows across a 4 KB boundary (or several) from the GO entry without hiding a Defender-style jump into a distant data buffer (`$1CB000` is a different 64 KB bank from a typical `$14xxx` start). A GO that arrives after the 32-slot table is full now logs `gpu_go_pages full` instead of dropping silently.

**Finding 2 (DSP canonicalize weakens escape) — accepted, documented, not mirrored.**

Jaguar addresses are 24-bit. `$FD012786` fetches from `$012786`; calling that an escape was a false positive on the same bus the GPU already canonicalizes. Comment added on `dsp_pc_valid`. A `dsp_runaway` twin is a follow-up if we see a Defender-style DSP jump into a data buffer; none is in #461.

**Finding 3 (`gpu_healthy_spin` requires the PC to differ) — incorrect reading; renamed the counters so it cannot recur.**

`gpu_same_pc_frames` was never a sampled-PC counter. It increments only when `gpu_exec_opcode_count == last_gpu_opcount` (zero opcodes this frame) and resets to 0 when any opcode runs. A tight `jr .` that still executes thousands of opcodes per frame has `gpu_zero_opcode_frames == 0` and *is* a healthy spin — that is Super Burnout / AvP idle, which is why the wedge predicate was changed off sampled-PC in the first place (issue #378). Renamed to `gpu_zero_opcode_frames` / `dsp_zero_opcode_frames`.

**Finding 4 (include path) — fixed, and the MSVC concern does not apply here.**

`gpu.c` now includes `../core/crash_detect.h`, matching `vjtrace.h`. `Makefile.common` already has `-I$(CORE_DIR)/src/core` for every host, including the MSVC buildbot rows (same Makefile, no separate vcxproj). The `windows-msvc` job on this PR was already green before the path change.

Notes on savestate / Defender logging both signatures: agreed, no change.

JoeMatt:
## Kimi follow-up

**Finding 1 (`video_stall` ignored GO windows) — agreed, fixed.**

`gpu_healthy_spin` now uses `gpu_pc_in_start_page()` (local RAM **or** a recorded 64 KB GO window) plus `gpu_zero_opcode_frames == 0`. A still image with the GPU spinning in a main-RAM program it was started at no longer logs `video_stall`. Defender-style runaway is still outside every window, so both signatures still fire together there. DSP stays local-RAM-only (no start-page table).

**Finding 2 (G_PC write while running) — this core does implement the jump; recorded now.**

`GPUWriteLong` `case 0x10` is `gpu_pc = data` with no latch and no running check. The next `GPUExec` fetches from that address (`GPUReadWord(gpu_pc)`). `CrashDetectNoteGPUGo(gpu_pc)` now runs on every G_PC write. A write-then-GO records the same page twice (deduped). 

`gpu_pc` at the GO edge **is** the freshly written start address: G_PC is case `0x10`, G_CTRL GO is case `0x14` on the same write path. Same assumption as `VJT_EV_GPU_GO`.

**Finding 3 (JTRM 24-bit cite) — comment corrected, masking kept.**

I should not have cited JTRM for "top byte ignored on every GPU/DSP fetch." The mask matches **this core's** decode: `GPUReadWord` does `addr &= 0x00FFFFFF` (`gpu.c:331`), and the 68K bus path does the same. `$FD012786` fetches `$012786` *here*; treating it as an escape was a false positive against our own fetch. No JTRM claim in the comment now.

Issue #322, opened 2026-08-05
[low-pri] CHD support: rip-format analysis + MAME/chdman patches needed first


JoeMatt:
### What

Low-priority tracking ticket for CHD, noted here so the analysis isn't lost. CHD loading was deliberately removed during the CD overhaul; CLAUDE.md's CHD mention is stale.

**Why Jaguar CHD is broken today**: Jaguar CDs are two-session discs with game data mastered inside audio-type tracks (byte-exact requirement: boot header at byte offset ≡ 2 mod 4 for the I2S capture skew; BIOS auth reads pregap audio). Classic CHD CD metadata has no session field; chdman historically flattened multi-session input, normalizes pregaps on ingest, and can shift the session-2 start LBA that depends on the real ~11,400-sector inter-session gap (`src/cd/cdintf.c:309` synthesizes it for BIN/CUE).

**Deliberately low priority**: the CHD rips in circulation are already broken at rip/convert time — re-adding a loader would not make them work.

### MAME-facing contributions (when someone wants this)

1. Audit current chdman multi-session support against a known-good Jaguar BIN/CUE: sessions, true gap size, pregap *contents* (not synthesized silence) must survive the round-trip; patch chdman where they don't.
2. Contribute our synthetic multi-session disc generator (`test/test_cd_synth_*` pattern) + a byte-survival checker as a chdman regression test — copyright-free validation for every data-in-audio system.
3. Mastering-spec doc for MAME so `createcd` defaults stop producing broken Jaguar CHDs.

### Our side (after the above proves round-trip fidelity)

Re-add libchdr as a bounded task; acceptance = boot matrix green on CHD conversions of the local corpus.

JoeMatt:
https://github.com/mamedev/mame/pull/15886/changes

This chdman MAME PR is merged, we're unblocked, but most user won't have the updated binary, so we should try to include our own build and helper tools for converting and detecting bad existing rips

JoeMatt:
Closing as completed — every item this ticket tracked has shipped:

- **Rip-format analysis** → [docs/jagcd-chd.md](https://github.com/libretro/virtualjaguar-libretro/blob/develop/docs/jagcd-chd.md) (two-session structure, CHSE requirement, session-gap synthesis rules, byte-offset constraints).
- **MAME/chdman patches** → [mamedev/mame#15886](https://github.com/mamedev/mame/pull/15886) merged; `chdman` now writes `CHSE` session metadata.
[...]

PR #468, opened 2026-08-16
feat(cd): load Jaguar CD CHDs that carry CHSE session tags (#322)

JoeMatt:
Refuse pre-15886 rips, vendor libchdr, and ship pinned chdman helpers on tagged releases and nightlies so users can convert CUE/BIN themselves.

## Test plan

- [ ] `make -j$(getconf _NPROCESSORS_ONLN)` builds clean on host
- [ ] `make test` passes (or N/A — explain why)
- [ ] `bash scripts/c89-lint.sh` passes (no mid-block declarations)

## Branch base

- [ ] **Base is `develop`** (the integration branch)
- [ ] OR base is `master` *and* this is a `hotfix/*` or `release/*` branch


JoeMatt:
## Kimi review

The review was truncated (no `cdintf.c` / tests / most of libchdr), so the findings are about the build wiring. Taking them in order:

**`unity.o` omitting `$(INCFLAGS)` — not a bug.** `FLAGS += $(INCFLAGS)` and then `CFLAGS += $(FLAGS)` (Makefile ~752 / ~781). The dedicated `unity.o` recipe already compiles with `$(CFLAGS)`, which is how a local `make deps/libchdr/unity.o` gets `-I./deps/libchdr/include` and the lzma/miniz/zstd dirs. Adding `$(INCFLAGS)` again would just duplicate them. The host builds that already went green on the first CI pass (Linux aarch64, both MSYS2, tvOS, macOS x86_64) compiled this object.

**`actions/checkout@v7` / `upload-artifact@v7` — already the repo standard.** `c-cpp.yml` uses the same majors; those jobs ran on this PR. A `@v7` tag that did not resolve would have failed every existing workflow, not just `chdman-tools.yml`.

**`jagcd-chd-check` defines — agreed, now aligned.** The checker only reads `CHSE`/`CHT2` metadata (`chd_get_metadata`), so `WANT_RAW_DATA_SECTOR` / `WANT_SUBCODE` / `VERIFY_BLOCK_CRC` do not change its verdict. `13ebc66` still compiles it with the same `LIBCHDR_CFLAGS` as the core so the shipped binary is not a narrower decode config.

**bare `cp` — agreed for the house rule.** Same commit uses `command cp -f`.

**How the synth fixtures were made — documented.** `docs/jagcd-chd.md` now says `test/roms/synth_jagcd.chd` / `synth_jagcd_nosession.chd` were built with the pinned post-15886 `chdman` (`-c none`). Homebrew 0.288 cannot recreate the CHSE file; that is why `jagcd_roundtrip.sh` exits 77.

The actual first-pass CI failures were elsewhere (not in the truncated diff):

- Linux Clang: `unity.c` includes miniz/lzma/zstd (and thus `<stdio.h>`) *before* `libchdr_chd.c`'s `_POSIX_C_SOURCE` / `_FILE_OFFSET_BITS`, so `-std=c99` treated `fseeko`/`ftello` as undeclared.
- MSVC x64/x86: the hardcoded `cl.exe` line never added `/Ideps\libchdr\include` or compiled `unity.c`.

Both are in `13ebc66`.

PR #473, opened 2026-08-17
fix(core): boot JaguarDemos BootIntros via real BIOS path

JoeMatt:
## Summary
- Auto-enable the cart boot ROM for jagcrypt/GPU-only carts under default HLE (`JaguarCartNeedsBIOS` + `apply_cart_bios_autodetect`)
- Add embedded Model-M boot ROM and `virtualjaguar_bios_type` core option (`k` / `m`); map selected ROM at `$E00000`
- Update JaguarDemos suite to probe BootIntros with `--bios` + PAL, prefer `vj.j64` / `*_K` ROMs, and refresh baseline

## Results
- `make jaguar-demos-full`: **46 PASS / 3 FAIL / 49** (was 8/41/49)
- `make jaguar-demos-smoke`: **8/8 PASS** (Fountain now passes)
- `make TEST_EXPORTS=1 test`: pass
- Remaining FAILs (accuracy gaps, recorded in `BASELINE.txt`): `256/Outline2026/rtb/rtb.j64`, `512/solidYARC512/solidYARC512_K.j64`, `64/columns32/columns32_K.j64`

## Test plan
- [x] `make jaguar-demos-full`
- [x] `make jaguar-demos-smoke`
- [x] `make TEST_EXPORTS=1 test`
- [x] `bash scripts/c89-lint.sh src/core/file.c libretro.c`
- [ ] CI (c-cpp + jaguar-demos workflows)


JoeMatt:
## Review follow-up (pushed)

Addressed the PR review blockers:

1. **Rayman Demo false positive** — FF-at-\`$802000\` heuristic now requires \`payload < 1 MiB\` (and \`>= 0x2100\` for the scan window). Confirmed real Rayman Demo → \`NeedsBIOS=false\`; Fountain \`vj.j64\` still \`true\`.
2. **FF-scan OOB** — no longer reads past \`payload\`.
3. **Path vs user/titledb** — \`cart_bios_type_from_path\` only runs when \`virtualjaguar_bios_type\` is still at its registered default (via \`get_variable_pertitle\`).
4. **Regression test** — \`test/test_cart_needs_bios.c\` (unit table + default-HLE autodetect arms), wired into \`make test\`.

\`make TEST_EXPORTS=1 test\` passed locally after these changes.

JoeMatt:
Pushed two `.info` follow-ups (`a1b75f4`, `e983222`) — packaging metadata only, no code touched.

**`a1b75f4` — advertise the optional Model-M boot ROM.** `stage_cart_boot_rom()` added one file to the system-directory contract: `<system>/jagboot_m.rom`, read as a 0x20000-byte override for the embedded Model-M image when `virtualjaguar_bios_type` is `m`. `dist/info/` still listed only the two CD BIOS overrides, so no frontend could report that slot. Bumped `firmware_count` 2 → 3 and added the optional `firmware2_*` triplet. No entry for Series K — that path has no external override.

**`e983222` — document every accepted CD BIOS filename.** Not a name correction: the two `.j64` paths already in `.info` are the first entries of `retail_names[]`/`dev_names[]`, i.e. the first names searched. The gap was documentation. `load_external_cd_bios()` accepts 12 filenames across 5 locations and `try_load_cd_bios_file()` identifies the image by checksum, so the filename only sets search order — a user with `Jaguar CD BIOS.rom` had no way to know why Core Information showed the `.j64` row as Missing while real-BIOS CD boot worked fine. That's now spelled out in `notes`.

The firmware block is deliberately left at 3 entries. `firmware*_path` is a single relative path with no alias or one-of-N semantics, so the subdirectory search is inexpressible, and enumerating the spellings would give a core that requires zero files eleven "Missing (optional)" rows.

Verified: `scripts/check-info-version.sh` passes (also runs as a pre-commit hook), `notes` is still a single pipe-delimited line, and `make -j8` links clean at `e983222`.

**Follow-up owed on merge:** [libretro-super](https://github.com/libretro/libretro-super/blob/master/dist/info/virtualjaguar_libretro.info) needs both changes — one new optional firmware entry plus the notes rewrite.

JoeMatt:
Re the [Kimi review](https://github.com/libretro/virtualjaguar-libretro/pull/473#issuecomment-5321431017) — both findings validated; both are intended behaviour that the code failed to explain. Addressed in `9a71037` as documentation at the two sites, no code change.

**Path-only loads (`libretro.c:2359`):** cannot occur on a compliant frontend. `retro_get_system_info` sets `need_fullpath=false`, so cart content always arrives with a data buffer; CD content is the only path-loaded class (via `CONTENT_INFO_OVERRIDE`) and takes the `is_cd_content` branch, never reaching `apply_cart_bios_autodetect`. `test_memory_map` asserts both halves (Tests 1b/1c). The `info->data && info->size` guard is defensive against non-compliant frontends only — now stated in a comment as suggested.

**Forcing `useJaguarBIOS` over an explicit user 'HLE' (`libretro.c:2364`):** deliberate. A GPU-only/jagcrypt cart has no 68K program at `$802000` for HLE to start, so honouring 'HLE' would be a guaranteed black screen. The K/M *type* hint defers to user/titledb values; the *enable* intentionally does not. Now documented both in the code and in the `virtualjaguar_bios` help text, so the "(Restart)" description on `bios_type` is no longer the only place.

**One the review missed, caught while verifying:** the `virtualjaguar_bios` help text claimed "the console boot ROM is never loaded from the system directory" — made false by this same branch's `jagboot_m.rom` override. Rewritten.

The suffix-matching nit (`cart_bios_type_from_path` not checking extensions) is left as-is per the review's own assessment: a false match only picks a boot ROM revision, and the K branch is a no-op at the default.

Also re-ran the full suite at the new tip before these doc edits: `make TEST_EXPORTS=1 test` all green at `e983222` — including the audio pair (clipping: Atari Karts/Skyhammer/IS2; presence: IS1 ×2), 0 failed, 1 recorded skip (CUE→CHD round-trip, no CHSE-capable chdman on PATH). `9a71037` is comments + option text over that.

Issue #437, opened 2026-08-13
[feat] Analog / driving / ADC-Reg controller types


JoeMatt:
Sub-issue of #428. BigPEmu exposes **analog**, **driving**, and **ADC-Reg** controller types alongside the standard pad. These read genuinely analog values rather than matrix bits, so unlike the mouse (#429) and rotary (#436) this is not purely a `$F14000` synthesis job.

**Estimate:** M — pending the research below; could be S if only one title ever used it, or L if it needs an ADC path we do not currently model.

## Open questions to answer before implementation

1. **Which real titles used these?** A controller type with no software that reads it is not worth shipping. Establish the title list first — this determines whether the ticket is worth doing at all.
2. **What is the hardware path?** ADC-Reg implies an analog-to-digital register rather than the joypad matrix. Identify the register, its address, and its timing from the JTRM / jag_sim netlists before designing anything.
3. **Head tracker (Jaguar VR)** — BigPEmu lists it. Almost certainly out of scope (no released software), but record the finding rather than leaving it unasked.

## Scope once answered

Accept `RETRO_DEVICE_ANALOG` per port, map axes to whatever the hardware path turns out to be, advertise via `SET_CONTROLLER_INFO`, reuse #429's device-selection scaffolding.

## Guardrail

Default off; with no analog device selected, input behaviour must be bit-identical to today.

JoeMatt:
Shipped in v3.4.0 (PR #486) via the TR10 bank-switching protocol. The hardware finding: the Jaguar has **no console-side ADC** — Atari deleted it from production silicon — so the controller carries its own microcontroller and answers the ordinary $F14000 row scan.

Caveat carried in the release notes: no released software reads this controller, so verification is synthetic (`test/tools/analog_decode_test`) and labelled as such.

JoeMatt:
Correction to a claim on this issue (and in the v3.4.0 release notes): *"no released software reads it; verification is synthetic"* was too broad.

The narrow version stands — no released title is known to read the **TR10 `$F14000` bank-switching protocol** this device implements, so the synthetic register-level suite in `test/tools/analog_decode_test.c` is still the right verification.

[...]

PR #474, opened 2026-08-18
feat(input): per-axis dead zone, offset and response exponent (#439)

JoeMatt:
Closes #439. Sub-issue of #428.

BigPEmu exposes per-axis dead zone, X/Y offset and a response exponent on its analog device types. This adds the same three controls as **one shared layer**, not per-device copies.

## Why this shape

#439 was deliberately sequenced last in #428 so the layer would be derived from real consumers rather than guessed. Two have shipped — the ST/Amiga mouse (#429, PR #449) and the Tempest rotary (#436, PR #451) — and they decided every design question:

**One insertion point, because both consumers already share one.** Mouse and rotary both reach the machine through a single call, `QuadFeed()` from `InputDevFeed()`. `src/jerry/axistune.c` sits exactly there. #437's analog/driving axes will arrive at the same call, so there is no per-device arithmetic to keep in sync and there must not be.

**Both consumers are relative axes, so the dead zone is a gate, not a re-base.** On an absolute axis a dead zone is conventionally re-based so there is no step at the boundary. On a relative axis there is no centre to re-base toward — the quantity is a *speed*, and a dead zone is a noise gate on it. Re-basing would subtract a constant from every sample, systematically shortening slow motion; on a device the game integrates into a position that is pointer drift versus the hand, which is worse than the step it removes. An absolute-axis consumer (#437) will have to answer the re-basing question for itself. That is deliberately **not** pre-answered here with an unused flag — one caller does not justify a mode switch.

**The offset is subtracted in host orientation, before the gate.** The dead zone and the curve are magnitude-symmetric and commute with a sign flip; the offset does not. The rotary negates its delta as a UX convention, so `inputdev_feed_axis()` negates the *tuned* delta and the code says why, because that is exactly the line a future reader "simplifies".

A real mouse reports zero at rest and is unaffected by any offset. The case it exists for is a frontend routing something else to the port's relative axis — a stick mapped to `RETRO_DEVICE_MOUSE` is the common one — where a mis-centred source drifts forever.

**The exponent's reference is `QUAD_MAX_BACKLOG` (64), not a taste constant.** A power curve needs a reference: `out = REF * (in/REF)^e` is the only form with a fixed point. `quadrature.h` already establishes 64 Gray states as the most a single feed can carry (a poll-rate-clocked encoder drains one state per poll; everything past 64 is clamped away), so 64 host units per poll at unity sensitivity is the top of the range this path can express. Anchoring anywhere else would have been the guess. At and above the reference every exponent is unity gain, so **no fast motion is ever lost to the curve**, and `test_axistune` asserts `AXISTUNE_REF == QUAD_MAX_BACKLOG` so the two headers cannot drift apart.

Worth stating plainly because users will hit it: an exponent above 1.0 makes ordinary movement *slower*, since ordinary movement is well below 64 units per poll. That is what an expo curve is; the paired sensitivity option is how you get top speed back. The option text says so.

**The curve comes out as a gain, not a curved delta.** This is not stylistic — it is the only form that keeps small movement alive. A curved delta has to be an integer, and at e=2.0 a 4-unit sample curves to `4 * 4/64 = 0.25`, which truncates to zero: the whole low end would quantise away and the mouse would read as dead. Handed back as a Q8 gain it multiplies into the scale `QuadFeed()` already applies, and QuadFeed's existing fractional carry — which exists for exactly this reason — accumulates it. The gain is floored at 1 so **the dead zone stays the only thing that ever drops a sample**.

**Integer-only, no libm.** The input-device guardrails are host-independent FNV digests (`joymatrix_identity.c` says so in its own header). A float `pow` would make them depend on the host's libm rounding and the constants would stop meaning anything across x86-64 / arm64 / i686. The fractional power uses the successive-square-root identity in Q16, verified against `libm` to <3% wherever the gain is meaningful and exact at halves and squares.

## Defaults are the identity, structurally

`AxisTuneApply()` returns its argument and a unity gain via an early return placed **before any arithmetic**. A zero-filled `axis_tune` is that identity too — `InputDevShutdown()` memsets its port array (iOS cannot dlclose a core), and a zeroed exponent field would otherwise leave a degenerate curve on every analog axis after a reload, i.e. a bug reachable on exactly one platform.

## Options

All hidden unless the relevant device is attached (existing visibility machinery, extended), all defaulting to the identity.

| Key | Values | Default |
| --- | --- | --- |
| `virtualjaguar_mouse_deadzone_x` | Off, 1, 2, 3, 4, 6, 8 units | `0` (Off) |
| `virtualjaguar_mouse_deadzone_y` | Off, 1, 2, 3, 4, 6, 8 units | `0` (Off) |
| `virtualjaguar_mouse_offset_x` | −4 … +4 | `0` (Off) |
| `virtualjaguar_mouse_offset_y` | −4 … +4 | `0` (Off) |
| `virtualjaguar_mouse_exponent_x` | Linear (1.00), 1.25, 1.50, 1.75, 2.00, 2.50, 3.00 | `100` (Linear) |
| `virtualjaguar_mouse_exponent_y` | Linear (1.00), 1.25, 1.50, 1.75, 2.00, 2.50, 3.00 | `100` (Linear) |
| `virtualjaguar_rotary_deadzone` | Off, 1, 2, 3, 4, 6, 8 units | `0` (Off) |
| `virtualjaguar_rotary_offset` | −4 … +4 | `0` (Off) |
| `virtualjaguar_rotary_exponent` | Linear (1.00), 1.25, 1.50, 1.75, 2.00, 2.50, 3.00 | `100` (Linear) |

A rotary is a single wheel, so it gets one tune and no X/Y split. The rotary options are **shared by both ports** — two rotaries cannot be tuned independently — which is a deliberate carry-forward of how `virtualjaguar_rotary_sensitivity` already behaves, and the reason they sit in the un-prefixed `input` category rather than a per-port one.

Clamping lives in `AxisTuneSet()` alone, so a hand-edited config is bounded in exactly one place regardless of which option it came through.

## Guardrail: the digital pad is untouched

A port whose type is `INPUTDEV_PAD` never enters the tuned paths at all. Proved two ways — the three existing `joymatrix_identity` digests are unmoved, and the new `tuning_identity` sweeps both ports as pads with an aggressive tuning configured and requires the digest to be identical to the untuned sweep.

## Tests

Two new tests, because the feature has two separable claims.

`test/test_axistune.c` — pure unit test, compiles `src/jerry/axistune.c` directly with no core (same shape as `test_quadrature.c`). Asserts the identity **exhaustively over the whole int16 range** rather than sampling it: `AxisTuneApply()` is the single insertion point, so "returns its argument unchanged at unity gain for every representable input" *is* the statement that a defaults-only configuration is byte-identical to the pre-#439 core. Also pins the zero-fill identity, gate-not-re-base (a re-basing implementation returns 1 where this requires 4), offset-before-gate ordering, `AXISTUNE_REF == QUAD_MAX_BACKLOG`, hand-computed curve values, no-zero-gain across every exponent, and monotonicity of `mag*gain`.

`test/tools/tuning_identity.c` — the same feature end to end through `$F14000`. It **self-baselines** (two runs of the same binary, one with the tuning API never called and one with it called at the defaults) instead of asserting a constant, because a constant would have to be measured on a build that already contains the change it is judging.

The third assertion is a **negative control** and is not optional: a non-default tuning must *move* the analog digest. Without it, both equalities would still pass on a build where the layer was never wired into `InputDevFeed()` at all.

Sweep order in that file is load-bearing and documented: a tune is a static that `InputDevReset()` deliberately does not clear (same as the sensitivity scale), so "the API was never called" is only observable before the first `InputDevSetTune()` in the process. Written the obvious way, the analog baseline inherits the pad sweep's tuning and the test reports the exact opposite of the truth — which is how it failed on the first run here.

## Test plan

All run on macOS arm64 with `DEVELOPER_DIR=/Library/Developer/CommandLineTools`.

```
$ make -j8 TEST_EXPORTS=1
  -> clean, no warnings

$ make -j8                      # reverse ABI flip, shipped-slim exports
  -> exit 0, clean

$ bash scripts/c89-lint.sh src/jerry/axistune.c src/jerry/inputdev.c \
                           libretro.c test/test_axistune.c
  -> C89 lint passed

$ ./test/tools/joymatrix_identity ./virtualjaguar_libretro.dylib \
      test/roms/yarc.j64 --quiet
  -> 4 passed, 0 failed
     joymatrix_identity_full             0xC24DDCDE (expected 0xC24DDCDE)
     joymatrix_identity_port1            0xD0CD02E2 (expected 0xD0CD02E2)
     joymatrix_identity_port1_with_mouse 0xD0CD02E2 (expected 0xD0CD02E2)
     joymatrix_identity_full_with_mouse  0x5E1858EA (expected 0x5E1858EA)

$ ./test/tools/mouse_decode_test  ./virtualjaguar_libretro.dylib test/roms/yarc.j64 --quiet
  -> PASS: [mouse_decode]  0 failure(s)

$ ./test/tools/rotary_decode_test ./virtualjaguar_libretro.dylib test/roms/yarc.j64 --quiet
  -> PASS: [rotary_decode] 0 failure(s)

$ ./test/test_axistune
  -> All axistune checks passed (37 checks, 0 failed)

$ ./test/tools/tuning_identity ./virtualjaguar_libretro.dylib test/roms/yarc.j64 --quiet
  -> PASS: [tuning_identity] 0 failure(s)
     pad    digests: untouched 0x4061DDC5  defaults 0x4061DDC5  tuned 0x4061DDC5
     analog digests: untouched 0xED6E4AD9  defaults 0xED6E4AD9  tuned 0x3BEA0665

$ make -j8 TEST_EXPORTS=1 test
  -> EXIT=0
     Core Option Visibility: 5 checks, 0 failed
     16 skipped checks, all "no private ROM/disc" (unchanged from develop)
```

The three digest constants were **not** touched. They are unmoved because the change cannot reach the pad path, which is the point of the exercise.

## Deliberately not done

- **No absolute-axis / re-basing mode.** #437 has no device yet; adding a flag with one caller would be the abstraction the issue's scope warns against. The header states the open question instead.
- **No per-title DB rows.** This is the mechanism, not the data.
- **No savestate change.** The tuning is option-derived and constant for a session and changes only how host motion becomes feeds, never anything the machine can observe — same reasoning `inputdev.h` already gives for the sensitivity scale. `INPUTDEV_STATE_SIZE` and `STATE_VERSION` are unchanged.
- **No invert / sign knob.** Out of scope and, per `inputdev.c`'s existing note, an unsourced sign knob is how a real wiring bug gets papered over instead of found.
- **No hardware verification**, because there is none to do: this layer is entirely host-side pre-processing and touches no emulated register behaviour. The JTRM has nothing to say about how a frontend's mouse deltas are conditioned before they become quadrature pulses.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -66,6 +66,7 @@ test/tools/netlink_game
 test/tools/joymatrix_identity
 test/tools/mouse_decode_test
 test/tools/rotary_decode_test
+test/tools/tuning_identity
 # vjtrace flight-recorder analyzer/smoke tools -- built on demand by
 # test/tools/vjtrace_selftest.sh (and by hand per each tool's own header
 # comment); never tracked, same Mach-O-on-Linux-CI hazard as above.
--- a/Makefile
+++ b/Makefile
@@ -923,8 +923,8 @@ clean:
 		test/test_titledb test/test_titlehook test/tools/test_hook_gate \
 		test/tools/test_wedge_spin test/tools/i2s_lag_probe \
 		test/tools/joymatrix_identity test/tools/mouse_decode_test \
-		test/tools/rotary_decode_test \
-		test/test_quadrature \
+		test/tools/rotary_decode_test test/tools/tuning_identity \
+		test/test_quadrature test/test_axistune \
 		test/.skipped-checks
 
 # Self-contained unit tests (parser + list management + simulated
@@ -980,8 +980,8 @@ test: test/test_dram_timing test/test_cheat test/test_event_queue test/test_jlin
 		test/tools/netlink_pair test/tools/netlink_latency test/tools/netlink_delay_proxy test/tools/test_pertitle_db \
 		test/tools/test_hook_gate \
 		test/tools/i2s_lag_probe test/tools/joymatrix_identity \
-		test/test_quadrature test/tools/mouse_decode_test \
-		test/tools/rotary_decode_test \
+		test/test_quadrature test/test_axistune test/tools/mouse_decode_test \
+		test/tools/rotary_decode_test test/tools/tuning_identity \
 		tools/jagcd/jagcd-chd-check
 	@# Skip ledger: truncate FIRST so a previous run's rows cannot resurface
 	@# as fresh skips (the stale-row failure mode documented for
@@ -1259,6 +1259,16 @@ test: test/test_dram_timing test/test_cheat test/test_event_queue test/test_jlin
 	@# two-controller Pause unlock, the C2/C3 ID bits and a savestate
 	@# round-trip mid-rotation.
 	./test/tools/rotary_decode_test ./$(TARGET) test/roms/yarc.j64 --quiet
+	@# Per-axis input tuning (#439), no core: the identity at defaults over
+	@# the whole int16 range, the dead zone as a gate rather than a re-base,
+	@# the curve's fixed point at QUAD_MAX_BACKLOG and its exact values.
+	./test/test_axistune
+	@# ...and the same feature measured end to end through $F14000: the
+	@# digital pad is bit-identical with tuning at ANY setting, the analog
+	@# paths are bit-identical at defaults, and a non-default tuning does
+	@# move the analog digest (the negative control that stops the two
+	@# equalities passing on a disconnected layer).
+	./test/tools/tuning_identity ./$(TARGET) test/roms/yarc.j64 --quiet
 	./test/test_memtrack
 	./test/test_nvmbios
 	@# Option visibility is content-type dependent; the disc half runs only
@@ -1632,6 +1642,20 @@ test/tools/rotary_decode_test: test/tools/rotary_decode_test.c \
 
 test/test_quadrature: test/test_quadrature.c src/jerry/quadrature.c src/jerry/quadrature.h
 	$(CC) -O2 -Wall $(INCFLAGS) -o $@ test/test_quadrature.c src/jerry/quadrature.c
+
+test/test_axistune: test/test_axistune.c src/jerry/axistune.c src/jerry/axistune.h \
+		src/jerry/quadrature.h
+	$(CC) -O2 -Wall $(INCFLAGS) -o $@ test/test_axistune.c src/jerry/axistune.c
+
+# #439's guardrail: needs the wide test ABI's InputDev* / Joystick* /
+# joypad0Buttons / joypad1Buttons exports.
+test/tools/tuning_identity: test/tools/tuning_identity.c \
+		src/jerry/inputdev.h src/jerry/axistune.h \
+		test/harness/harness.c test/harness/harness.h
+	$(CC) -O2 -Wall -std=c99 $(INCFLAGS) \
+		-o $@ test/tools/tuning_identity.c \
+		test/harness/harness.c \
+		$(if $(filter Linux,$(shell uname -s)),-ldl) -lm
 
 test/tools/test_option_visibility: test/tools/test_option_visibility.c
 	$(CC) -O2 -Wall -std=c99 $(INCFLAGS) \
--- a/Makefile.common
+++ b/Makefile.common
@@ -81,6 +81,7 @@ SOURCES_C :=  \
 	$(CORE_DIR)/src/bios/jagdevcdbios.c \
 	$(CORE_DIR)/src/jerry/joystick.c \
 	$(CORE_DIR)/src/jerry/quadrature.c \
+	$(CORE_DIR)/src/jerry/axistune.c \
 	$(CORE_DIR)/src/jerry/inputdev.c \
 	$(CORE_DIR)/src/core/settings.c \
 	$(CORE_DIR)/src/core/memtrack.c \
--- a/libretro.c
+++ b/libretro.c
@@ -242,6 +242,21 @@ static bool         show_rotary_options     = true;
 static int32_t      mouse_scale_q8          = 256;
 static int32_t      rotary_scale_q8         = 256;
 
+/* Per-axis tuning (#439), resolved from the options and applied to
+ * whichever device is actually in each port -- the same "one ladder per
+ * device kind, selected by what is plugged in" shape as the scales above.
+ * A rotary is one wheel, so it has a single tune and no Y entry.
+ *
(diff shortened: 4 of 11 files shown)


JoeMatt:
Pushed `bc24d5d` — fixes a real defect found reviewing this PR.

**The `offset` knob manufactured input from an idle device.** `AxisTuneApply()` subtracted the offset before checking whether there was a sample at all. `libretro.c` feeds every attached mouse/rotary port on every `retro_run`, passing 0 when the frontend reports no motion — so with any non-zero `virtualjaguar_mouse_offset_x/y` or `virtualjaguar_rotary_offset`, and the default dead zone of 0, the idle path computed `v = 0 - offset`, cleared the gate, and fed `QuadFeed` a phantom delta **every frame with the device physically untouched**.

That is precisely the failure the option exists to cure. The rotary offset's own help text promises it stops a source that "would otherwise spin the knob forever with the controls untouched"; the mouse text claims "a real mouse reports exactly zero at rest and is unaffected". Both were false for any non-default offset.

**Fix:** on a *relative* axis `raw == 0` does not mean "a reading at the biased centre" — it means no sample this poll. Offset corrects a source that reports movement; it has nothing to correct when the source reports none. So return before the subtraction.

**Test:** `test_offset_leaves_idle_alone()` sweeps every offset in [−64, 64] at dead zone 0, repeats with a non-identity exponent (the early-out must precede the curve too), and asserts the offset still does its real job — a device genuinely reporting +3 at rest is still cancelled, real motion still de-biased. Verified non-vacuous: with the early-out removed, both sweeps fail.

**Verification:** `make -j8 TEST_EXPORTS=1 test` exit 0. Guardrail digests unmoved — joymatrix `0xC24DDCDE` / `0xD0CD02E2` / `0xD0CD02E2` / `0x5E1858EA`, `tuning_identity` 0 failures. `c89-lint` passed.

Three lower-confidence findings from the same review pass are deliberately not addressed here: the user guide's option table is now stale (nine new options), `test_axistune.c`'s "monotone for every exponent" docstring overclaims (real dips exist below `exponent_q8 = 100`, a range the option pipeline clamps out of reach), and `inputdev.c`'s enumerating comment about what survives a reset doesn't mention the new `tune_x`/`tune_y`. None block.

Issue #469, opened 2026-08-16
[bug] Crash in Demo ROM `Outline2026/Fountain/vj.j64`


42Bastian:
### Game / ROM

Fountain (Homebrew)

### Steps to reproduce

ROM: 
https://codeberg.org/42Bastian/JaguarDemos/src/branch/main/256/Outline2026/Fountain/vj.j64

Just load.

### Expected vs actual

Expect: Runs.
Actual: Crash of core and RetroArch

### BIOS mode

Real BIOS (built into the core; runs the actual boot sequence)

### Blitter mode

Accurate (Midsummer2)

### Core version

3.3.0

### Frontend + platform

1.22.2

### Logs

```shell
-- no log generated --
```

### Pre-flight

- [x] Verified the issue is in this libretro core (not standalone Virtual Jaguar or another emulator).
- [x] Searched [existing issues](https://github.com/libretro/virtualjaguar-libretro/issues?q=is%3Aissue) for duplicates.

JoeMatt:
Crash site and fix are in https://github.com/libretro/virtualjaguar-libretro/pull/470 (against develop).

**Crash site (BIOS-only):** around **frame 13–15**, GPU PC leaves `$F03xxx` (last good `$F036A8` `JUMP T,(r30)`) after the 68K, running through **PRNG exception vectors**, writes GPU RAM and `G_PC`. The GPU then executes TOM MMIO as code; `tomWidth` clamps to **1024** while `retro_get_system_av_info.max_width` is **652** → RetroArch video path aborts with no core log. HLE hangs at `$802000` (empty entry) and does not host-crash.

**Root cause:** `JaguarReset` PRNG-fills RAM including vectors 2–255 in real-BIOS mode. Fountain’s jagcrypt path starts the GPU before the boot ROM installs handlers. RAM RTE stubs do not work: BIOS memcpy during decrypt overwrites `$400`.

**Fix:** point vectors 2–255 at the boot ROM’s existing `BRA.S *` at `$E005DC` until BIOS writes real handlers.

Headless 180-frame BIOS+PAL run stays at presented width 326 (no 1024). I could not drive RetroArch in this session — please confirm there before closing.

JoeMatt:
[...]

PR #477, opened 2026-08-18
feat(core): external cart boot ROM loader with known-hash identification

JoeMatt:
## Summary

Delivers the design agreed in #469's discussion: "list the known BIOSes explicitly, then support a generic loader as an extra." Extends `virtualjaguar_bios_type` from `{k, m}` to `{k, m, custom}`. `custom` searches the system directory for an external 128 KB cart boot ROM, identifies it by CRC32 against a small known-image table, and falls back to the embedded Series K image if nothing usable is found. `k` and `m` (including the existing `jagboot_m.rom` override) are unchanged.

Builds directly on develop's #473 (K/M selector) — this PR only adds the third option value and the search/identify machinery behind it.

## Known-image hash table (`src/core/biosdb.c`)

Sourced from the six-image hash table @42Bastian and I worked out in #469. Only the four images we can actually confirm are listed; BJL and the Alpine board's own boot ROM are intentionally **not** included yet — their hashes are still unconfirmed on the issue thread. Adding either later is a one-line table entry.

| CRC32 | Name |
|---|---|
| `fb731aaa` | Series K boot ROM |
| `ae25bdf5` | Model M boot ROM |
| `e60277bb` | Stubulator '93 dev boot ROM |
| `8d15dbc6` | Stubulator '94 dev boot ROM |

Any other CRC32 → `custom/unrecognized image` (still loaded — that's the point of the option).

All four entries were cross-checked against real dumps during this session (not just against the embedded K/M images): the embedded `jaguarBootROM` / `jaguarBootROM_M` blobs hash to `fb731aaa` / `ae25bdf5`, and the private corpus's `[BIOS] Atari Jaguar (World).j64`, `Stubulator '93`, and `Stubulator '94` `.j64` files hash to `fb731aaa`, `e60277bb`, and `8d15dbc6` respectively — all four match the issue's table exactly.

## Search order (`load_external_cart_boot_rom()` in `libretro.c`)

For `custom`, the system directory (and `Atari - Jaguar`, `jaguar` subfolders — the CD BIOS loader's `sub_dirs` list minus the CD-specific entries) is searched **filename-first**, in this order:

1. `jagboot.rom`
2. `boot.rom`
3. `boot0.rom` (MiSTer convention)
4. `[BIOS] Atari Jaguar (World).j64`
5. `[BIOS] Atari Jaguar Stubulator '94 (World).j64`
6. `[BIOS] Atari Jaguar Stubulator '93 (World).j64`

For each filename, all three directories are tried before moving to the next filename — so a higher-priority filename several sub-directories down still beats a lower-priority filename sitting in the system directory root (pinned by `custom_filename_priority_beats_subdirectory`). Only exactly-131072-byte files are accepted; the first one found wins. Its identity is logged either way:

```
[BOOT] external cart boot ROM <path>: Series K boot ROM (crc fb731aaa)
[BOOT] <path> is an unrecognized cart boot ROM image (crc <x>) -- loading anyway, custom images are the point
[BOOT] external cart boot ROM <path>: custom/unrecognized image (crc <x>)
```

## Fallback semantics

`custom` selected but no usable (131072-byte) file found anywhere in the search path → embedded Series K, with a warning log:

```
[BOOT] Custom cart boot ROM selected but no usable file found -- falling back to embedded Series K
[BOOT] cart boot ROM: Series K (fallback)
```

Never fails the load.

## Test plan

```
bash scripts/c89-lint.sh          # -> C89 lint passed
DEVELOPER_DIR=/Library/Developer/CommandLineTools make -j8
DEVELOPER_DIR=/Library/Developer/CommandLineTools make TEST_EXPORTS=1 test
```

Result: `make test` exits 0. Two new suites, both part of `make test`:

- `test/tools/test_biosdb.c` (direct-link unit test, no dlopen, no ROMs): table integrity, the two embedded images hash+identify correctly, the two Stubulator CRCs look up correctly, an unrecognized synthetic buffer falls back cleanly, and (optional, skips cleanly when the private corpus is absent) a real `[BIOS] Atari Jaguar (World).j64` identifies as Series K. **14/14 pass** (including the optional check, present in this session's corpus).
- `test/test_cart_bios_loader.c` (dlopens the built core, drives `retro_load_game()` with a synthetic cart): custom+file-present loads the file's bytes at `$E00000`, custom+no-file falls back to embedded K, k/m unchanged (still load their embedded image even with a same-named file sitting right there), a wrong-size candidate is skipped and falls through to the fallback, filename search finds a file in a sub-directory, filename priority beats sub-directory depth, and (optional, present in this session's corpus) a real recognized K dump loads through the full path byte-identical to the embedded image. **8/8 pass**.

`test/tools/test_option_visibility` (existing) also passes unmodified — `virtualjaguar_bios_type` visibility rules didn't need to change, only its value set did.

Production ABI restored after testing (`make -j8`, no `TEST_EXPORTS`).

## Deliberately not done

- **BJL and Alpine boot ROM hashes** — pending 42Bastian's reply on #469. The table is a plain array of `{crc32, name}`; adding either is a one-line entry, no other code changes needed.
- No change to `virtualjaguar_bios_type`'s visibility rules (still hidden for CD content) — CD content stages this boot ROM too (unconditionally, unchanged pre-existing behavior for `k`/`m`), so a `custom` selection carried over from cart play will log the fallback warning on CD loads. Not a regression — consistent with how `m`'s `jagboot_m.rom` probe already behaves on CD loads — but worth a maintainer glance.


JoeMatt:
## Root cause

`test/test_cart_bios_loader.c` called `mkdtemp()` at all 8 test sites. On Linux/glibc, `mkdtemp`'s prototype is hidden under `-std=c99` unless a POSIX feature-test macro (`_POSIX_C_SOURCE`/`_XOPEN_SOURCE`/`_DEFAULT_SOURCE`) is defined — macOS's libc doesn't gate it the same way, which is why this only ever showed up in CI, never locally.

- **Under GCC** the missing prototype is only a warning. The implicit declaration defaults the return type to `int`, truncating the real 64-bit pointer to 32 bits. The corrupted `dir` pointer then segfaults the moment it's dereferenced (e.g. inside `snprintf(path, ..., "%s/jagboot.rom", dir)`), which is exactly the `Segmentation fault (core dumped)` in the linked job log.
- **Under Clang** the same missing prototype is a hard compile error: `call to undeclared function 'mkdtemp'`.

Both are the same root cause manifesting differently per toolchain, which is why it took down all five checks at once (build-Linux x86_64 GCC/Clang, build-Linux aarch64, coverage, and ASan+UBSan — the sanitizer job also builds with clang).

**Reproduced directly** in a Linux x86_64 container (Ubuntu 24.04): with the old code, `gcc -std=c99` produced the implicit-declaration warning and the built test binary segfaulted exactly as CI showed; `clang -std=c99` failed to compile with the "call to undeclared function" error. Both are clean after the fix.

## Fix

`test/test_cart_bios_loader.c` already had two siblings that hit this exact issue before — `test/test_cd_pregap.c` and `test/test_cd_synth_cdda.c` — both documenting the same workaround in comments. Applied that established pattern here: replaced `mkdtemp()` with a `getpid()`-keyed `mkdir()` at all 8 call sites. These scratch dirs need no cryptographic uniqueness — tests run sequentially in-process and clean up after themselves — so a name keyed on the PID is sufficient and portable across both toolchains.

## Verification

- Linux x86_64 container (gcc 13 and clang 18): compiles clean, `test/test_cart_bios_loader` runs clean, 7 tests PASS + 1 SKIP (no private ROM in the container), exit 0. Confirmed both toolchains failed before this fix and pass after.
- macOS: `make TEST_EXPORTS=1 test` — full suite passes (9/9), including `test_biosdb` (14/14 assertions) and `test_cart_bios_loader` (8/8, including the private-corpus real-BIOS case).
- `scripts/c89-lint.sh test/test_cart_bios_loader.c` — passes.

No production code changed — the loader in `libretro.c` (`load_external_cart_boot_rom()` / `try_load_cart_boot_rom_file()`) was correct all along; the defect was entirely in the test's temp-directory setup.

JoeMatt:
Thanks — the `clean:` duplication is real and is now fixed in `779b1f9`.

Verified before fixing: `test_titledb`, `test_titlehook`, `test_hook_gate`, `test_wedge_spin` and `i2s_lag_probe` did each appear twice. The cause was the develop merge (which brought in `test_texdump` from #478) being resolved as a union of both sides — correct for the build lists, but it duplicated names in `clean:` that were already present below. Only `test_biosdb` and `test_cart_bios_loader` were genuinely new, exactly as you said.

Fix collapses the block so each entry appears once; `make -n clean` still parses. Confirmed no remaining duplicates:

```
awk '/^clean:/{f=1} f&&/^$/{exit} f' Makefile | tr ' ' '\n' | grep -E "^test/" | sort | uniq -c | awk '$1>1'
```
returns nothing.

On the cleared items — the partial-write hygiene analysis is the one worth confirming out loud, since it was the non-obvious risk in this PR: a short read can leave a torn image in the `$E00000` window mid-search, and it is only safe because every exit path (a later candidate succeeding, or the embedded fallback `memcpy`) rewrites the full 128 KB. That invariant is load-bearing, so if a future change ever adds an early return between the read and the fallback, this becomes a real bug.

PR #478, opened 2026-08-18
feat(texdump): texture dump mode (issue #369, deliverable 1)

JoeMatt:
Implements texture dump mode per the approved design spec (`docs/texture-dump.md`, first commit on this branch): every unique blit **source** tile a title pushes through the blitter is written once per session to `<system_dir>/vj_texdump/<cart CRC32>/` as a preview PNG plus an append-only `manifest.tsv` row, for HD-pack authoring and title inspection.

## What's here

- **`src/tom/texdump.c`/`.h`** — capture at the shared blit-launch site (beside `BlitMemoLaunch`, before engine dispatch): SRCEN-gated, register-described source walk, FNV-1a 64 key, open-addressed 2^17 dedupe set (allocated on first enable, freed + statics reset in `retro_deinit`/`retro_unload_game` — iOS never dlcloses), first-sight PNG render + manifest row. Windows >1 MB or touching non-populated address space are skipped.
- **Source-channel decode verified against `docs/jtrm-blitter.md`**, not source comments: B_CMD bit 0 (SRCEN) reads from **A2**, or from **A1 when DSTA2 (bit 11)** is set. Per-pixel addressing transliterates the engines' `PIXEL_OFFSET_*` formulas (width 6-bit float, pitch phrase-gap `{0,1,3,2}`, per-bpp packing).
- **PNG encoder** — in-tree miniz 3.1.1 (`tdefl_write_image_to_png_file_in_memory_ex`), enabled by defining `MINIZ_DEFLATE_APIS` for the libchdr unity TU; writes go through the libretro VFS (`rfopen`/`rfwrite`). No new vendored code.
- **Options** — `virtualjaguar_texture_dump` (disabled default, runtime-toggleable, no restart) and `virtualjaguar_texdump_16bpp` (cry|rgb|both), the latter visible only while the dump is enabled (same visibility machinery as the existing gates). Inserted as one contiguous block after `virtualjaguar_bios_type` to keep merge conflicts with the concurrent BIOS-loader branch minimal.
- **Zero savestate fields, no STATE_VERSION bump.** All state is host-transient.

## Identity contract

The key is **FNV-1a 64 over `'VJTD' | ver=1 | src_bpp | w:u16 | h:u16 | raw source bytes, row-major, packed as stored`** (little-endian serialization). **The palette is deliberately NOT part of the key**: replacement happens at blit time and the Jaguar blitter never sees a palette — CLUT lookup is an Object Processor display-time concept, so two CLUT variants of one indexed tile are indistinguishable at the swap point. Palette information is advisory manifest metadata (`clut=` CRC on first sight, extra sighting rows for later palettes), never identity. The manifest header carries `texdump v1` so a forced re-key is visible, never silent.

## Test plan

All gates run inside `make test` via `test/tools/test_texdump.c` (in-tree `test/roms/yarc.j64`; never skips):

```
DEVELOPER_DIR=/Library/Developer/CommandLineTools make -j8 TEST_EXPORTS=1 test   # exit 0
DEVELOPER_DIR=/Library/Developer/CommandLineTools make -j8                        # reverse ABI flip links clean
bash scripts/c89-lint.sh src/tom/texdump.c        # C89 lint passed (ditto texdump.h, blitter_mmio.c, libretro.c)
./test/tools/test_texdump ./virtualjaguar_libretro.dylib test/roms/yarc.j64 --frames 600
```

Result (9/9):

- `contract_freeze` — 28 hashes match the committed golden list `test/expected/texdump_yarc.txt`
- `determinism` — two consecutive dump-on runs, identical hash sets
- `engine_independence` — fast vs accurate blitter, identical hash sets
- `inertness_on` — 600 per-frame framebuffer hashes AND savestate digests at frames 300/600 identical dump-on vs dump-off (the machine cannot see the feature)
- `png_validity` / `synth_png_validity` — every emitted PNG parsed structurally: signature, IHDR dims vs manifest, every chunk CRC-32 via `src/core/crc32.c` (no external decoder)
- `synth_population` / `tile_population` — 28 unique tiles (2 ROM boot + 26 scripted)
- `palette_not_identity` — a CLUT-change re-blit of known bytes mints NO new hash and appends exactly one sighting row

## Deviations from spec

1. **Golden-list population**: yarc.j64 blits only its two boot-time code copies through SRCEN (it renders via the OP) — and so does `jagniccc.j64`, the only other in-tree public ROM (measured at 600 and 900 frames). Instead of the spec's "add a second committed ROM" fallback (there is no adequate one), the test drives a **synthetic blit battery** through the core's real bus/launch path (`JaguarWriteLong` → TOM → `BlitterWriteWord` → `TexDumpLaunch`, TEST_EXPORTS ABI): 26 scripted tiles covering 1/2/4/8/16/32 bpp, both source channels (A2 and DSTA2's A1), and the palette re-blit check. The committed golden list freezes ROM + battery together — a stronger tripwire than any in-tree ROM run.
2. **`MINIZ_DEFLATE_APIS` lives in `LIBCHDR_CFLAGS`** (Makefile.common) rather than on the `unity.o` rule alone: the theos_ios path compiles `unity.c` through theos' own rules, bypassing `unity.o`, and would otherwise ship without the encoder. The effect is identical everywhere else (the flag reaches only unity.c and the standalone CHD tools). The CI MSVC job also compiles unity.c with `/DMINIZ_DEFLATE_APIS`, confirming the deflate code under cl.exe; the spec's fallback wrapper was not needed (plain extern linkage suffices).
3. **`flags=` manifest tokens are `bcompen`/`dcompen`**, not the spec's `transen`: B_CMD has no TRANSEN bit (verified against `docs/jtrm-blitter.md`) — bit-comparator (26) and data-comparator (27) transparency are what the hardware actually has.
4. Palette-sighting dedupe reuses the main hash set (key ⊕ CLUT CRC inserted alongside the primary key) instead of a parallel per-entry structure — same behavior as specified (one sighting row per new palette per tile), no extra allocation.

Also fixed in passing: `test/test_blitter_mmio.c` gained the texdump stubs its standalone link now needs, and `test/tools/test_option_visibility.c` asserts the 16bpp knob is hidden while the dump option is at its default.

**Found while testing (pre-existing, not addressed here):** a fast→accurate engine flip leaves static residue that changes the next session's savestate bytes when the core stays resident (iOS no-dlclose regime) — reproduced with texture dump never enabled. The test sequences its runs to sidestep it; see the "Run order note" in `test/tools/test_texdump.c`.

Part 1 of #369; replacement pipeline is deliverable 2 (v3.5).

### DIFF
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -417,9 +417,9 @@ jobs:
           /I. /Isrc /Isrc\core /Isrc\tom /Isrc\jerry /Isrc\cd /Isrc\bios /Isrc\m68000 /Ilibretro-common\include ^
           /Ideps\libchdr\include /Ideps\libchdr\deps\lzma-25.01\include /Ideps\libchdr\deps\miniz-3.1.1 /Ideps\libchdr\deps\zstd-1.5.7 ^
           /D__LIBRETRO__ /DINLINE="_inline" /DBLITTER_SIMD_SCALAR ^
-          /D_7ZIP_ST /DWANT_RAW_DATA_SECTOR=1 /DWANT_SUBCODE=1 /DVERIFY_BLOCK_CRC=1 ^
+          /D_7ZIP_ST /DMINIZ_DEFLATE_APIS /DWANT_RAW_DATA_SECTOR=1 /DWANT_SUBCODE=1 /DVERIFY_BLOCK_CRC=1 ^
           libretro.c ^
-          src\tom\blitter.c src\tom\blitter_compare.c src\tom\blitter_mmio.c src\jerry\dac.c src\jerry\dsp.c src\core\file.c ^
+          src\tom\blitter.c src\tom\blitter_compare.c src\tom\blitter_mmio.c src\tom\texdump.c src\jerry\dac.c src\jerry\dsp.c src\core\file.c ^
           src\tom\gpu.c src\core\jaguar.c src\jerry\jerry.c src\tom\tom.c src\tom\op.c ^
           src\cd\cdintf.c src\cd\cdrom.c src\core\crc32.c src\core\event.c ^
           src\jerry\eeprom.c src\core\filedb.c src\jerry\joystick.c src\core\settings.c ^
--- a/Makefile
+++ b/Makefile
@@ -893,6 +893,10 @@ endif
 # libchdr is C99; the rest of the core is gnu89. A dedicated rule so
 # the generic %.o: %.c line cannot compile unity.c as C89.  Must sit
 # AFTER `all:` so it is not the default goal.
+#
+# MINIZ_DEFLATE_APIS (texture dump's PNG encoder, issue #369) rides in
+# LIBCHDR_CFLAGS -- see the comment in Makefile.common -- so the
+# theos_ios path, which compiles unity.c without this rule, gets it too.
 $(LIBCHDR_DIR)/unity.o: $(LIBCHDR_DIR)/unity.c
 	$(CC) -c $(OBJOUT)$@ $< $(CFLAGS) $(LIBCHDR_CFLAGS) $(LIBCHDR_WARNFLAGS)
 
@@ -921,7 +925,7 @@ clean:
 		test/tools/test_memory_map test/tools/test_option_visibility test/test_memtrack test/test_nvmbios test/tools/test_dsp_audio_diag \
 		test/tools/test_frame_timing test/tools/test_runahead_determinism test/tools/test_pertitle_db \
 		test/test_titledb test/test_titlehook test/tools/test_hook_gate \
-		test/tools/test_wedge_spin test/tools/i2s_lag_probe \
+		test/tools/test_wedge_spin test/tools/test_texdump test/tools/i2s_lag_probe \
 		test/tools/joymatrix_identity test/tools/mouse_decode_test \
 		test/tools/rotary_decode_test \
 		test/test_quadrature \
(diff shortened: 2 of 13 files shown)


JoeMatt:
## AddressSanitizer + UndefinedBehaviorSanitizer fix

**Root cause:** a 32-byte heap leak, not a memory-safety bug in the core. LeakSanitizer caught it in `test/tools/test_texdump.c:709` (`main`):

```
Direct leak of 32 byte(s) in 1 object(s) allocated from:
    #0 malloc
    #1 main test/tools/test_texdump.c:709:33
SUMMARY: AddressSanitizer: 32 byte(s) leaked in 1 allocation(s).
```

`main()` pre-parses the tool's own `--golden`/`--update-golden`/`--json`/`--quiet` flags into a filtered `fargv` array (so the shared harness's arg parser doesn't mistake `--golden`'s value for a positional ROM path), then reassigns `argv = fargv`. That heap array was never freed on any exit path.

**Fix:** `free(argv)` right after the last read of `argv[0]` — once on the `harness_init_from_args` failure branch (before `return 1`), once on the normal continuation. Confirmed by grep that `argv`/`fargv` isn't referenced anywhere later in `main`.

**Scope:** test-tool only — no core `src/` file touched, no change to the FNV-1a-64 hashing/identity path. The golden list (`test/expected/texdump_yarc.txt`) is byte-for-byte unaffected; `contract_freeze` (28/28 hashes) still passes.

**Verification:**
- Reproduced the exact CI flags in a Linux (aarch64) container mirroring the workflow's `SAN_FLAGS`/`ASAN_OPTIONS`/`UBSAN_OPTIONS`: before the fix, `test/tools/test_texdump` reproduced the identical LeakSanitizer report; after the fix, `make test TEST_EXPORTS=1` under the same sanitizer build exits 0 with no ASan/LSan/UBSan output anywhere in the suite.
- `test/tools/test_texdump` alone: 9/9 gates pass (png_validity, determinism, engine_independence, inertness_on, synth_png_validity, synth_population, palette_not_identity, tile_population, contract_freeze).
- Native macOS: `DEVELOPER_DIR=/Library/Developer/CommandLineTools make -j8 TEST_EXPORTS=1 test` exits 0 (same 9/9 texdump gates).
- `bash scripts/c89-lint.sh test/tools/test_texdump.c` → `C89 lint passed`.

Pushed as `5799b6a` on this branch.

PR #486, opened 2026-08-18
feat(input): TR10 bank-switching analog / driving controller (#437)

JoeMatt:
Implements #437 — the third device in the #428 input epic, on the scaffolding established by the mouse (#429), rotary (#436) and shared per-axis tuning (#439/#474).

## What the hardware research established (sources first)

**The analog / driving controller is a bank-switching matrix device, not an ADC register.** The Jaguar Technical Reference V10 (SgM Electrosoft; `docs/atari-jaguar-1999/Technical Reference v10.pdf`, gitignored) specifies it completely in three sections — *"Analogue Joystick and 'Driving' Controllers"*, *"Identifying Controller Types"*, *"Reading Bank Switching Controllers"*:

- There is **no console-side ADC**: *"Early versions of the Jaguar included an 8 bit ADC on the motherboard. This has been deleted — analogue controllers now require their own ADC chip."* The controller carries a microcontroller (Atari's prototype: 68HC05P9) that digitises two 100 K pots and answers the ordinary `$F14000`/`$F14002` row scan with data bits. So, like the rotary, this is row-select-driven matrix synthesis — no new bus plumbing.
- **Two banks**, switched *"automatically when the controller sees a transition from row 3 to row 0 (of the same controller socket)"* — i.e. clocked by row-select **writes**, not reads — with interleaved reads of other sockets explicitly required not to desynchronise it. Bank 0 carries X/Y as four nibbles (rows 0/1 = X low/high, rows 2/3 = Y low/high) plus buttons A–D on the B column; bank 1 carries the hat/gear-shift switches in row 0 and the all-ones type-ID pattern elsewhere.
- **Identification**: C2 C3 = 0 1 ("Bank Switching"), row-0 C-column = bank-0 flag, last-bank B-column 1111 = "Analogue Joystick or Driving Controller".
- **"Analog" and "driving" are one wire protocol** — TR10 defines one device with two interpretations (stick: X=roll/Y=pitch; driving: X=steering/Y=accelerator-brake). The two core option values differ only in host mapping.

**No released title reads this protocol.** Atari never shipped the controller. BigPEmu's celebrated Checkered Flag analog support is a *game patch* (script), not this device; the one known consumer of console-side analog input, JANALOG.ABS, targets the deleted early-board ADC. Per the ticket's research-first mandate, the deliverable therefore pivots as instructed: **the device type plus a documented synthetic verification** (`test/tools/analog_decode_test.c` — a TR10-faithful driver written independently of the emulation, scanning banks out of `$F14000`/`$F14002` exactly as TR10 tells a 68K game to). The test's header and the user guide both label it as such.

**ADC-Reg is out of scope as unsourced, and documented so.** The early-board ADC's pins exist (PAD0X/PAD0Y/PAD1X/PAD1Y; TRM rev 8 labels GPIO5 `$F17C00-$F17FFF` "Paddle Interface") but no document we hold gives the register layout, it was deleted from production hardware, and only dev-unit test software ever read it. Emulating it would mean inventing register behaviour. Recorded in `docs/input-devices-user-guide.md` so the question stays asked-and-answered; same for the **head tracker** (type ID exists in TR10, no released software — Missile Command 3D was a prototype).

## The absolute-axis tuning decision (#439's open question, now answered)

`axistune.h` deliberately deferred the gate-vs-re-base question to the first absolute consumer. With one in hand, the answer is **re-base**, via a new `AxisTuneApplyAbs()` in the same shared layer (no second tuning path):

- On a *position* there is no integration, so the drift argument that forbade re-basing on relative axes doesn't exist — and a gate would put a step at the dead-zone edge (centre → edge-value jump = steering snap). The surviving magnitude is remapped `(mag − dz) · range / (range − dz)`: continuous at the edge, full deflection preserved.
- **The identity rule differs from #474's relative rule, deliberately**: on a relative axis `raw == 0` means "no sample" and must stay 0 under every tune. On an absolute axis `raw == 0` **is** a sample (the stick at rest): dead zone and exponent are magnitude-symmetric and fix the centre, and **offset alone may move it** — cancelling a mis-centred rest position is its entire job, so there is no `raw == 0` early-out.
- The exponent anchors at the caller's `range` (full deflection), not `AXISTUNE_REF` — e > 1 buys finer centre control at zero range cost, which is why the analog ladder has no sensitivity option.

All spelled out in `axistune.h` ("THE ABSOLUTE-AXIS ANSWER") and pinned by new `test/test_axistune.c` cases plus register-level assertions in the decode test.

## Option / device shape

- `virtualjaguar_p1_device` / `p2_device` gain `analog` and `driving` values (both ports — TR10 restricts neither socket). Default remains the pad; the plain `RETRO_DEVICE_ANALOG` id deliberately maps to the pad so "RetroPad w/ Analog" users are never silently switched.
- `SET_CONTROLLER_INFO` advertises `RETRO_DEVICE_SUBCLASS(RETRO_DEVICE_ANALOG, 0/1)` on both ports.
- Shared tuning ladder `virtualjaguar_analog_{deadzone,offset,exponent}_{x,y}` (both ports, like the rotary's), units = ADC counts, shown only while an analog type is attached.
- **Liveness guardrail**: the port keeps its RetroPad — and the device stays entirely inert, controller-type probes included — until the stick deflects past the threshold ("engagement", the mouse's `inputdev_live` rule adapted to an absolute source, where a centred stick is indistinguishable from no routing). Documented consequence: a title probing controller types at boot sees a pad unless the stick moves first.

## Savestate

Bank counter, last observed row, latched ADC bytes and switch mask are machine-visible (TR10's own driver recipe depends on bank sync surviving) and join the input-device chunk (41 → 51 bytes). v12 is develop-only (v3.3.0 shipped v11), so per the one-bump-per-release policy the v12 layout is **extended in place**; a pre-extension state reads the zero-fill tail, which is inert. Engagement is host-routing-derived and not serialized (same argument as `inputdev_live`). iOS statics reset in `retro_deinit` / `InputDevShutdown`.

## Guardrails held

- `joymatrix_identity` digests **unchanged** (0xC24DDCDE / 0xD0CD02E2 / 0x5E1858EA) — pad-only behaviour is bit-identical, and the never-fed analog device is proven bit-identical to a pad in the decode test.
- `mouse_decode_test`, `rotary_decode_test`, `tuning_identity` all pass unmodified (only the hand-restated chunk-size mirror moved, 41 → 51, as its comment intends).
- `InputDevOverlayF14000` gained row parameters (the analog device drives a different X/Y slice per row); the mouse ignores them, and `joystick.c`'s row tables were hoisted to file scope so the write path can decode rows for the bank clock.

## Self-review findings

A reviewer pass over the staged diff returned **no correctness bugs at confidence ≥ 80**. It independently re-derived and confirmed the chunk arithmetic (41 + 2×5 = 51), the AND-only overlay property, the bank-toggle condition, the port-dependent C-column mapping, the `continue` that stops the analog branch falling into the mouse path, and the export-list coverage (`InputDev*` wildcards already cover the two new entry points).

Two observations it raised that I judged **not** to be defects, with reasons:

- *"In row 0 the analog bank-flag bit is the same `$F14002` bit `joystick.c` decodes as PAUSE (bit 0 port 1 / bit 2 port 2)."* Correct observation, not a collision: an engaged analog port has its whole pad array cleared by `update_input()` before any read, and a real analog controller has no Pause button — that line *is* the bank flag on the hardware. Same shape as the rotary's existing C-bit note. Verified independently.
- *"`AxisTuneApplyAbs`'s `dz >= range` branch is dead at the analog call site."* True there (`AxisTuneSet` clamps dz to 64, the call site passes range 127), but `AxisTuneApplyAbs` is a general API in a shared layer, so the guard stays and is exercised directly by `test/test_axistune.c`.

The reviewer also flagged a methodology caveat — its session had no shell, so it read whole files rather than the diff, and suggested confirming nothing outside those files was touched. Confirmed separately: `git diff --cached --stat` covers exactly the 14 files listed in this PR.

Separately, and because it has bitten two PRs this week: every touched C file was compiled with `-std=c99 -Werror=implicit-function-declaration -Werror=incompatible-pointer-types` (the Linux-Clang failure mode) — all clean. `InputDevOverlayF14000`'s prototype and definition were changed together and agree exactly.

## Testing

- `DEVELOPER_DIR=/Library/Developer/CommandLineTools make -j8 TEST_EXPORTS=1 test` — **exit 0**.
- `joymatrix_identity` — all three digests unmoved: `0xC24DDCDE` / `0xD0CD02E2` / `0x5E1858EA` (plus the with-mouse port-1 digest). `mouse_decode`, `rotary_decode`, `tuning_identity` — 0 failures each, unmodified.
- New `test/tools/analog_decode_test` (in `make test`): bank-0/1 layout, value mapping (centre 128, TR10 +Y = stick forward), bank cycling incl. other-socket interleave immunity, type-ID bits with the port-dependent C-row mapping, engagement inertness, analog≡driving on the wire, absolute tuning semantics at the register level, 51-byte chunk + mid-cycle rollback replay.
- `test/test_axistune`: absolute identity over the domain, re-base continuity + monotonicity, centre-fixing rule, offset-moves-centre, curve anchored at full deflection, dz≥range degenerate.
- `bash scripts/c89-lint.sh` clean on every touched file.

Closes #437.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -66,6 +66,7 @@ test/tools/netlink_game
 test/tools/joymatrix_identity
 test/tools/mouse_decode_test
 test/tools/rotary_decode_test
+test/tools/analog_decode_test
 test/tools/tuning_identity
 # vjtrace flight-recorder analyzer/smoke tools -- built on demand by
 # test/tools/vjtrace_selftest.sh (and by hand per each tool's own header
--- a/Makefile
+++ b/Makefile
@@ -928,7 +928,8 @@ clean:
 		test/test_titledb test/test_titlehook test/tools/test_hook_gate \
 		test/tools/test_wedge_spin test/tools/test_texdump test/tools/i2s_lag_probe \
 		test/tools/joymatrix_identity test/tools/mouse_decode_test \
-		test/tools/rotary_decode_test test/tools/tuning_identity \
+		test/tools/rotary_decode_test test/tools/analog_decode_test \
+		test/tools/tuning_identity \
 		test/tools/blitter_static_leak \
 		test/test_quadrature test/test_axistune \
 		test/.skipped-checks
@@ -987,7 +988,8 @@ test: test/test_dram_timing test/test_cheat test/test_event_queue test/test_jlin
 		test/tools/test_hook_gate \
 		test/tools/i2s_lag_probe test/tools/joymatrix_identity \
 		test/test_quadrature test/test_axistune test/tools/mouse_decode_test \
-		test/tools/rotary_decode_test test/tools/tuning_identity \
+		test/tools/rotary_decode_test test/tools/analog_decode_test \
+		test/tools/tuning_identity \
 		test/tools/blitter_static_leak \
 		tools/jagcd/jagcd-chd-check
 	@# Skip ledger: truncate FIRST so a previous run's rows cannot resurface
@@ -1283,6 +1285,14 @@ test: test/test_dram_timing test/test_cheat test/test_event_queue test/test_jlin
 	@# two-controller Pause unlock, the C2/C3 ID bits and a savestate
 	@# round-trip mid-rotation.
 	./test/tools/rotary_decode_test ./$(TARGET) test/roms/yarc.j64 --quiet
+	@# TR10 bank-switching analog / driving controller (#437): NO RELEASED
+	@# TITLE reads this protocol, so this synthetic register-level suite IS
+	@# the device's verification (the tool's header says so at length): a
+	@# TR10-faithful driver scans both banks out of $F14000/$F14002 and
+	@# asserts the layout, the bank cycling (incl. other-socket interleave
+	@# immunity), the type ID bits, the engagement guardrail, the absolute
+	@# tuning semantics and the extended v12 savestate chunk.
+	./test/tools/analog_decode_test ./$(TARGET) test/roms/yarc.j64 --quiet
 	@# Per-axis input tuning (#439), no core: the identity at defaults over
 	@# the whole int16 range, the dead zone as a gate rather than a re-base,
 	@# the curve's fixed point at QUAD_MAX_BACKLOG and its exact values.
@@ -1692,6 +1702,13 @@ test/tools/rotary_decode_test: test/tools/rotary_decode_test.c \
 		test/harness/harness.c test/harness/harness.h
 	$(CC) -O2 -Wall -std=c99 $(INCFLAGS) \
 		-o $@ test/tools/rotary_decode_test.c \
+		test/harness/harness.c \
+		$(if $(filter Linux,$(shell uname -s)),-ldl) -lm
+
+test/tools/analog_decode_test: test/tools/analog_decode_test.c \
+		test/harness/harness.c test/harness/harness.h
+	$(CC) -O2 -Wall -std=c99 $(INCFLAGS) \
+		-o $@ test/tools/analog_decode_test.c \
 		test/harness/harness.c \
 		$(if $(filter Linux,$(shell uname -s)),-ldl) -lm
 
--- a/docs/input-devices-user-guide.md
+++ b/docs/input-devices-user-guide.md
@@ -189,10 +189,54 @@ row scans. We reproduce that, so three odd-looking behaviours are faithful:
 All three happen on real hardware with a real adapter. They are commented as
 such at the code sites so nobody "fixes" them later.
 
+## Analog / driving controller (#437)
+
+Atari specified — but **never released** — an analog controller: a
+bank-switching matrix device with its own ADC chip, documented in the Jaguar
+Technical Reference V10 ("Analogue Joystick and 'Driving' Controllers",
+"Reading Bank Switching Controllers"). The core emulates that specification
+faithfully on both ports, as two option values that are one wire protocol:
+
+| Option value | Host mapping |
+|---|---|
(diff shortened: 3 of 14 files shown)


JoeMatt:
Heads-up on a doc interaction: PR #487 (open, docs-only) rebuilt `docs/input-devices-user-guide.md`'s option table from `libretro_core_options.h` — 2 rows to 14, grouped by device with visibility gating verified against `update_option_visibility()`. It deliberately **excluded** this PR's analog options because #437 hadn't merged when it was written.

So whichever of us merges second should add the analog rows (and the analog device to the port-2 device list, which #487 also corrected from four to five). Flagging rather than doing it here to avoid a conflict in the same file while both are open.

JoeMatt:
Rebased onto develop (`def8635`) — the conflict was the `docs/input-devices-user-guide.md` interaction flagged above.

#487 merged first, so develop already carried the rebuilt 14-row option table. The conflict reduced to a single prose line about which epic items remain open; resolved to keep #487's phrasing while recording analog as shipped. The analog option rows were already present in this branch's copy, so nothing was lost from either side.

Verified on the new base rather than trusting the clean rebase: full `make -j8 TEST_EXPORTS=1 test` **exit 0**, guardrail digest `0xC24DDCDE` unmoved, `analog_decode_test` in-suite and passing.

One build note for anyone rebasing another branch today: clear `deps/libchdr/unity.o` first. `MINIZ_DEFLATE_APIS` now rides in `LIBCHDR_CFLAGS` (texture dump's PNG encoder), and the build-config stamp doesn't track that flag — a stale object link-fails with an undefined `tdefl_write_image_to_png_file_in_memory_ex` that looks like a texdump bug but isn't.

Issue #438, opened 2026-08-13
Lightgun support — first establish whether any title uses one


JoeMatt:
Sub-issue of #428. Placeholder so the epic's device list is tracked in issues rather than prose.

**Estimate:** S to research, unknown to implement — **the first question is whether this should exist at all.**

## Step 1: does any Jaguar title use a lightgun?

The Jaguar had no first-party lightgun. Before any implementation work, establish whether any released or homebrew title reads one, and through what hardware. If the answer is "none", **close this issue with the finding recorded** — that is a successful outcome, not a failure.

## If a title does exist

- libretro provides `RETRO_DEVICE_LIGHTGUN` with screen-relative coordinates.
- The emulation-side question is the same as every other device in #428: what does the peripheral put on JERRY's `$F14000` matrix (or elsewhere), and when — a lightgun classically latches on a video-timing event, which would make this the only device in the epic that couples to TOM's beam position.
- Source the mapping authoritatively before coding; do not guess.

## Guardrail

Default off; no effect on pad-only titles.

JoeMatt:
## Answered: the hardware supports it, the peripheral never existed, and no commercial title reads it

This issue's gate was "first establish whether any title uses one." Result: **zero commercial titles, two homebrew tech demos.**

### The silicon is real — this part is primary-sourced

From the JTRM in `docs/atari-jaguar-1999/` (Technical Reference v8/v10 carry extractable text; the numbered volumes are image scans):

- **v10 pinout:** Port 1 pin 6 = *"Button input / Light Gun on Port 1"*. Port 2 pin 6 is `B2` only.
- **v10 prose:** *"A TTL rising edge on the LP signal (pin 6 of Port 1, shared with B0) causes the light pen registers (LPH and LPV) to be latched."*
- **v8 registers:** `LPH $F00008 RO`, `LPV $F0000A RO`; chip pin list carries `XLP  Light pen input.`

Two details that matter. The manual calls the registers *light pen* but the pinout says *Light Gun* — same signal, and the inconsistency is itself a fact. And because **LP is shared with B0**, a trigger is readable as a plain button via `JOYBUTS` without touching LPH/LPV — so an **LPH/LPV read is the necessary signature of positional gun use**, which is what makes it scannable.

**Atari's own controller-ID scheme (v10 p.35, diodes C1/C2/C3) enumerates four types — 4-player adaptor, bank-switching, rotary, reserved. There is no lightgun type.** A gun was never a planned product.

[...]

PR #487, opened 2026-08-18
docs(input): PR #474 review follow-ups (option table, docstring, comment)

JoeMatt:
## Summary

Three documented-but-non-blocking findings from [PR #474's review](https://github.com/libretro/virtualjaguar-libretro/pull/474#issuecomment-5322384227) (last paragraph). All comment/doc accuracy fixes — **no behavior change**.

- **`docs/input-devices-user-guide.md`**: the options table predated the nine per-axis tuning options (#474/#439) and the rotary device/sensitivity options (#451) — two options documented where fourteen now exist. Rebuilt from `libretro_core_options.h` on develop, grouped by device selection / mouse tuning / rotary tuning, with visibility gating pulled from `update_option_visibility()` in `libretro.c` rather than guessed. Also fixes the "offers the same four devices" line (`retro_controller_info` lists five for port 2 now) — the same count-mismatch class the #450 review flagged for "All three live under Port 2" — and updates "Not yet implemented" now that rotary and per-axis tuning have shipped, correcting an #437/#438 issue-number swap along the way (#437 is analog/driving controllers, #438 is lightgun — verified via `gh issue view`).

  `feat/437-analog-controllers` has **not** merged as of this PR (no open PR for that branch, issue #437 still open), so its options are intentionally not in the table, per the review's instruction not to invent unmerged content.

- **`test/test_axistune.c`**: invariant 9's docstring claimed mag\*gain monotonicity "for every exponent", but `test_response_is_monotone()` only sweeps `exponent_q8` in `[100, 800]`, and the invariant is genuinely false below that (verified independently here, not transcribed: `exponent_q8=1` dips mag 15→16 from mag\*gain 16290→16288). That region is unreachable through the shipped option pipeline — the exponent core options only enumerate 100%-300% (`exponent_q8` 256-768). No assertion or shipped arithmetic changed.

- **`src/jerry/inputdev.c`**: `inputdev_reset_dynamic()`'s enumerating comment named the device type and sensitivity scale as option-derived state that survives a machine reset, but omitted `tune_x`/`tune_y` (#439), which survive by the same rule — `test/tools/tuning_identity.c`'s sweep-order comment documents that survival as load-bearing. Minimal one-block comment edit to keep any merge with the in-flight #437 branch (which also touches this file) trivial.

## Verification

- `bash scripts/c89-lint.sh src/jerry/inputdev.c test/test_axistune.c` — passed.
- `DEVELOPER_DIR=/Library/Developer/CommandLineTools make -j8 TEST_EXPORTS=1 test` — full clean run: 0 failures. Guardrail digests unmoved (per the original PR's own stated values): joymatrix `0xC24DDCDE` / `0xD0CD02E2` / `0xD0CD02E2` / `0x5E1858EA`, `tuning_identity` 0 failures.
- Diff is comment/doc-only: `git diff -U0` on the two `.c` files shows every changed line inside a `/* */` block; the third file is Markdown.
- A later `make test` run hit an unrelated `netlink_latency_test.sh` failure (`mktemp: mkstemp failed ... File exists`) caused by several other concurrent agent worktrees in this environment running `make test` at the same time (confirmed via `ps aux`, not caused by this change); the test passed cleanly on an isolated retry.

## Test plan

- [x] `c89-lint` on both touched `.c` files
- [x] `make TEST_EXPORTS=1 test` exit 0, guardrail digests unmoved
- [x] Diff-is-comment/doc-only assertion made from the actual diff, not memory
- [x] Checked `feat/437-analog-controllers` merge status before finishing (still open, correctly excluded)

No Copilot review requested, per repo policy.

### DIFF
--- a/docs/input-devices-user-guide.md
+++ b/docs/input-devices-user-guide.md
@@ -26,14 +26,14 @@ plugged into either port draws on the same five rotary options.
 | *Port 2 > Mouse Sensitivity* (`virtualjaguar_mouse_sensitivity`) | 25% – 400% | 100% |
 | *Port 2 > Mouse Dead Zone (X)* (`virtualjaguar_mouse_deadzone_x`) | Off, 1 – 8 units | Off |
 | *Port 2 > Mouse Dead Zone (Y)* (`virtualjaguar_mouse_deadzone_y`) | Off, 1 – 8 units | Off |
-| *Port 2 > Mouse Offset (X)* (`virtualjaguar_mouse_offset_x`) | -4 – +4 | Off |
-| *Port 2 > Mouse Offset (Y)* (`virtualjaguar_mouse_offset_y`) | -4 – +4 | Off |
+| *Port 2 > Mouse Offset (X)* (`virtualjaguar_mouse_offset_x`) | -4 – -1, Off, +1 – +4 | Off |
+| *Port 2 > Mouse Offset (Y)* (`virtualjaguar_mouse_offset_y`) | -4 – -1, Off, +1 – +4 | Off |
 | *Port 2 > Mouse Response Curve (X)* (`virtualjaguar_mouse_exponent_x`) | Linear (1.00) – 3.00 | Linear (1.00) |
 | *Port 2 > Mouse Response Curve (Y)* (`virtualjaguar_mouse_exponent_y`) | Linear (1.00) – 3.00 | Linear (1.00) |
 | *Rotary Sensitivity* (`virtualjaguar_rotary_sensitivity`) | 25% – 400% | 100% |
 | *Rotary Reports Controller Type* (`virtualjaguar_rotary_id`) | Standard Joypad (no diode), Tempest Rotary (diode fitted) | Standard Joypad |
 | *Rotary Dead Zone* (`virtualjaguar_rotary_deadzone`) | Off, 1 – 8 units | Off |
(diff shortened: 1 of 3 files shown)


JoeMatt:
Good catch — real inconsistency, fixed in `bf10639`.

Verified against the source before choosing which side to correct. `libretro_core_options.h` defines the offset enums as:

```c
{ "-4", "-4" }, { "-3", "-3" }, { "-2", "-2" }, { "-1", "-1" },
{ "0",  "Off" },
{ "1",  "+1" }, { "2",  "+2" }, { "3",  "+3" }, { "4",  "+4" },
```

So `0` **is** in the enum and its display label **is** `Off` — exactly the convention the dead-zone rows already document as `Off, 1 – 8 units`. That means the Default column was right and the Values column was incomplete, rather than the other way round. Taking your second suggestion (changing Default to `0`) would have made the table disagree with what the user actually sees in the menu.

Values now read `-4 – -1, Off, +1 – +4` on all three rows (mouse X, mouse Y, rotary), which keeps the dead-zone convention and puts `Off` where it genuinely sits in the enum rather than tacking it on the front.

On the two things you flagged as unverifiable from the diff alone — the visibility-gating prose and the enumerated value lists — those were derived by reading `update_option_visibility()` and `libretro_core_options.h` directly rather than inferred, and this finding is a fair demonstration that the transcription step was the weak link. Appreciated.

Issue #400, opened 2026-08-11
Doom savestate rollback is non-deterministic — breaks RetroArch run-ahead (we advertise savestate_features=3)


JoeMatt:
`test/tools/test_runahead_determinism` **fails on Doom** while passing on every title the suite actually exercises. Since we advertise `savestate_features = 3` ("deterministic") in `dist/info/virtualjaguar_libretro.info`, RetroArch enables single-instance run-ahead on that promise — and single-instance run-ahead is exactly `retro_serialize` → run forward → `retro_unserialize`. If the rollback isn't bit-exact, run-ahead corrupts the running game.

## Evidence (300 frames, `develop` @ bae6aff)

| ROM | result |
|---|---|
| **Doom - Evil Unleashed** | **FAIL** — video diverged at replay frame 227/300; `repeated_rollback_agrees` disagrees at replay frame 183 |
| Hover Strike | PASS (4/4) |
| yarc.j64 | PASS (4/4) |
| Wolfenstein 3D | PASS |
| Atari Karts | PASS |

Pre-existing, not a recent regression: identical divergence reproduces at `fd06e59` (before #395 merged), same frames.

## Why the suite never caught it

`make test` runs this tool against `test/roms/yarc.j64` and Iron Soldier only (Makefile ~line 1062) — both pass. No Doom-class title is covered.

## Narrowing hint

`state_reconverges` **passes** (state after replay is byte-identical, 2621440 bytes) while video during replay differs. So the persisted blob is complete enough to re-converge; the divergence is in transient state that is not restored in the same phase — a scheduled event's remaining time, an execution-slice boundary, or similar. The bus arbiter fields *are* serialized (`libretro.c:1146-1149`), so that's already ruled out.

## Secondary: the tool reports false audio failures above ~300 frames

At `--frames 600` every ROM (including known-good yarc) reports `audio diverged at replay frame 360 ... pass1 800 samples, pass2 0 samples`, with the divergence running to the end of the run. That is the harness's `HARNESS_MAX_AUDIO_FRAMES = 1200` capture cap being hit across the two passes, not a core defect — the recorded frames simply stop. The tool should either cap its frame count, reset audio capture between passes, or report the exhaustion explicitly, so this doesn't get mistaken for a real regression.

## Impact

[...]

PR #489, opened 2026-08-18
feat(input): light gun on port 1 (#438)

JoeMatt:
Closes #438. Epic #428.

Adds `RETRO_DEVICE_LIGHTGUN` as a **port-1** device type, following the mouse
(#449) and rotary (#451) precedent, and implements `docs/lightgun-design.md`
(merged as #482) — with two of its load-bearing decisions corrected by the ROM.

**Default off.** `virtualjaguar_p1_device` stays `auto` → pad; a pad-only session
is bit-identical and `joymatrix_identity`'s digests
(`0xC24DDCDE / 0xD0CD02E2 / 0x5E1858EA`) are untouched.

## How it works

TOM's `LPH` (`$F00008`) / `LPV` (`$F0000A`) are synthesised in
`TOMExecHalfline()` from the aim point, latching on the half-line the beam
reaches it. The forward map is the direct inverse of what the scanline
renderers already do:

| quantity | source |
|---|---|
| HC origin of framebuffer column 0 | `TOMGetLeftVisibleHC()` — the same origin `startPos = HDB1 - leftHC` is measured from |
| video clocks per column | `PWIDTH`, `VMODE` bits 9-11 ("width of pixel in video clock cycles", value written + 1) |
| half-line of framebuffer row *r* | `topVisible + 2*r` — where `TOMExecHalfline()` writes row *r* |

All read live from the registers, so a title that reprograms its display
geometry is followed automatically (Balloons programs its own `HDB1`, `HDE`,
`VDB` and `PWIDTH=3`), and PAL/NTSC needs no branch here because those
accessors already resolve it.

**No beam hit-test window and no per-clock horizontal counter.** MiSTer's model
needs a live beam position; ours advances `HC` once per half-line. It doesn't
matter: a corpus scan finds **zero** readers of `HC`/`VC` (`$F00004`/`$F00006`)
across the whole ROM set, Balloons included — the latched value *is* the entire
observable. This is the design's §3.4 call, now backed by evidence rather than
by proportionality.

**Savestates need no code.** `state.h` saves `tomRam8` wholesale, so the latched
pair rides along. No new chunk, no `STATE_VERSION` bump.

**Hi-res guardrail (#400).** `libretro.c` divides `shadowHiresN` out before the
aim reaches the core, so a 2x session latches byte-identical `LPH`/`LPV` to a 1x
one. Asserted by the test, not assumed.

**Off-screen** suppresses the latch rather than clamping to an edge — a real gun
pointed away from the CRT sends no pulse and the console keeps its last reading.
The trigger still registers, because the pin is electrically the same wherever
the barrel points.

## Deviations from the design

Balloons landed in the private corpus after the design was written. Three of its
open questions became measurements, and two of the answers contradict it. All
three are written up in a new **"AS BUILT"** section at the top of
`docs/lightgun-design.md`; the original research text below it is untouched so
the reasoning that led there stays readable.

### 1. The LP latch is CONTINUOUS, not trigger-gated (§3.4 was wrong)

§3.4 says to synthesise on "the trigger's rising edge". Balloons' loops
(`$4266`, `$43D6`, `$452C`) do:

```
stop   #$2000                  ; wait for the video interrupt
move.w LPH,$7E50               ; read the pair UNCONDITIONALLY, every field
move.w LPV,$7E52
...
move.w d0,$4E28                ; and write the decoded position straight into
move.w d0,$4E2A                ; the CROSSHAIR sprite's X / Y
```

The trigger is tested separately, afterwards. A trigger-edge-only latch freezes
the crosshair. It's also the only thing that makes sense electrically: a real
gun's photodiode pulses `LP` every field the beam crosses the aim point, and
**there is no "`LP` was pulsed" status bit anywhere in the register map** — so
software cannot detect a shot from `LPH`/`LPV` at all. The trigger has to be
something else, which leads to:

### 2. The trigger is `BUTTON_B`, not `BUTTON_PAUSE` (§3.2 was wrong)

§3.2 reasons from TR10's "pin 6 of Port 1, shared with B0" to
`joypad0Buttons[BUTTON_PAUSE]`. The pin-sharing fact may well be true of the
silicon; it just isn't how a game reads a shot.

Balloons scans all four rows into a bitmask (`$4A42`), edge-detects it into
`$4C60`, and tests **bit 25**. Simulating `joystick.c`'s decode against that
exact scan sequence maps bit 25 to row 1 / `JOYBUTS` bit 1 = **`BUTTON_B`**.
(`BUTTON_PAUSE` lands on bit 28 and is never tested.)

Mapping as shipped — mechanism is the rotary's (ordinary matrix slots), as §3.2
concluded, just a different slot:

| libretro | Jaguar |
|---|---|
| Trigger | **B** |
| Aux A / Aux B | A / C |
| Start / Select | Option / Pause |

### 3. `LPH` is emitted as a gap-free count, not TOM's bit-10 encoding

The design left the encoding open. It matters, and only the ROM could settle it.

TOM encodes `HC` as bit 10 = "second half-line" plus a `0..HP` offset, so raw
values jump `845 → 1024` mid-line (NTSC) — a 179-count hole. Balloons decodes
with a bare `(LPH - calibration_LPH) / PWIDTH` and no bit-10 handling.

**Both encodings were built and run against the ROM**, sweeping 21 aim points
across the width:

| encoding | result |
|---|---|
| gap-free (shipped) | exact at every column, 0 → right edge |
| bit-10 raw `HC` | exact to column 237; column **238** decodes to object X **318** instead of **259** — `+59 px`, i.e. `179 / PWIDTH(3)`, exactly where Balloons' half-line boundary falls |

**Balloons' calibration screen passes under both**, because it only records the
raw `LPH` it is handed and range-checks `LPV`. So "calibration converges" — the
design's own §6 test — is *not* sufficient, and a single-point aim check on the
left half would have shipped the wrong encoding. Nothing else in this core reads
`LPH`, so there is no second consumer to keep consistent.

## Verification

`test/tools/test_lightgun`, wired into `make test`, in two halves.

**Part A — public ROM, never skipped.** `LPH`/`LPV` affine in the aim
column/row with slope `PWIDTH` / 2; off-screen does not relatch; and the same
normalised aim latches byte-identical registers at 1x and at
`internal_resolution=2x`.

**Part B — Balloons, private-corpus gated** (exit 77 + skip ledger when absent;
that path was exercised by hand, not just written):

- its two-target **calibration** screen, driven with the gun, accepted:
  `cal1=$010D cal2=$04F1`
- a **21-point sweep** across the full width asserting the crosshair the *game*
  computes from our `LPH` lands on the object coordinate the aimed-at pixel
  actually occupies — exact at every point (`startPos=-21, topVis=25, pwidth=3`)
- a scripted **balloon kill**, detected via the game's own respawn counter
  (`$7E48`, which `$4724` is the only thing that raises) — hit on the first shot
- `VIDEN` still set, so Balloons' own "aim below half-line `$1BE` and quit" path
  can never be misread as a miss

```
=== Results: 7 passed, 0 failed, 0 skipped ===
  PASS: [lightgun_transform] LPH = leftHC(188) + col*pwidth(4), LPV = topVis(25) + 2*row, 4 points across the width
  PASS: [lightgun_offscreen] off-screen kept LPH=1164 LPV=185 (last on-screen 1164/185)
  PASS: [lightgun_hires_identity] 1x and 2x latch identical LPH/LPV for the same normalised aim (first 252/65, last 1164/185)
  PASS: [balloons_calibration] cal1=$010D cal2=$04F1 (both targets accepted: yes)
  PASS: [balloons_still_running] display stayed enabled for the whole run
  PASS: [balloons_crosshair_tracks] 21 point(s) across the width; game-computed crosshair == the aimed pixel exactly (startPos=-21, topVis=25, pwidth=3)
  PASS: [balloons_hit] hit registered after 1 shot(s)
```

Also: `make -j8 TEST_EXPORTS=1 test` exits 0 (joymatrix digests unmoved),
`scripts/c89-lint.sh` clean on every touched file, no new source files so
`Makefile.common` and the MSVC list are untouched.

## Still open

Bullets (`cubanismo/bullets`) remains unbuildable without an SDK that isn't
vendored, so which button *it* reads is unverified; if it wants A, that is
`AUX_A` today — a mapping question, not a design one. Team Tap compatibility
(§5.3) and the gun's controller-ID bits (§5.4) are still unsourced and nothing
was invented for them.

Not implemented, deliberately (design §4): no crosshair overlay (frontends draw
their own from the same state), no `RETRO_DEVICE_ID_LIGHTGUN_RELOAD` (no
register-level analogue), and none of Bullets' `170` or MiSTer's
`X_SKEW_PIX`/`Y_SKEW_PIX` fudge constants — the direct-inverse transform has no
residual error for them to cancel.

Please do not run a Copilot review on this one.


JoeMatt:
Rebased onto `develop` after #486 (analog / driving controllers) landed. Every conflict was a parallel addition to the same framework, so all of them were resolved as the **union** — nothing from either feature was dropped.

**What conflicted, and how it was resolved**

- `src/jerry/inputdev.h` — `InputDevType` enum: `INPUTDEV_ANALOG`/`INPUTDEV_DRIVING` and `INPUTDEV_LIGHTGUN` now all present.
- `src/jerry/inputdev.c` — six sites. Port struct keeps both the `an_*` block and the `gun_*` block; `inputdev_reset_dynamic()` and the type-change path clear both; `InputDevSetType()` now enforces mouse = port 2 only, light gun = port 1 only, rotary/analog = either; `InputDevFeedAnalog()`, `InputDevFeedLightgun()` and `InputDevLightgunAim()` all kept intact side by side. `inputdev_attach_mask` is type-agnostic (set for anything that is not a plain pad), so the gun registers as attached exactly as before.
- `libretro.c` — device macros (`RETRO_DEVICE_JAG_ANALOG`/`_DRIVING` **and** `_LIGHTGUN`), the device-name switch, the `p1_device` option parser, `retro_set_controller_port_device`, and the `SET_CONTROLLER_INFO` table (port 1 is now 5 devices, port 2 stays 7). In `update_input()` the analog `if` block and the light gun `if` block are both present as separate early-`continue` branches — the gun keeps its hi-res (#400) divide-down guardrail untouched.
- `libretro_core_options.h` — the `virtualjaguar_p1_device` values list is the union (`auto`/`pad`/`rotary`/`lightgun`/`analog`/`driving`) and the description text carries both features' prose.
- `Makefile` — `test_lightgun` and `analog_decode_test` both in the build deps and both invoked by `make test`; the Balloons ROM gate and its skip ledger entry are unchanged.
- `docs/input-devices-user-guide.md` — kept the analog section, added the light gun section back. The core-options table on `develop` was already stale w.r.t. #486 (it listed 14 options and no analog rows), and the port-count prose could not be made correct without fixing it, so the table now lists all 20 options and both new devices. Flagging that explicitly since it is slightly wider than pure conflict resolution.

**Verification** (all on the rebased branch)

- `make -j8 TEST_EXPORTS=1` builds and links clean.
- `test/tools/test_lightgun` with `BALLOONS.BIN`: **7 passed, 0 failed** — including `balloons_calibration`, `balloons_crosshair_tracks` (game-computed crosshair lands exactly on the aimed pixel across 21 points) and `balloons_hit`.
- `test/tools/analog_decode_test`: pass, savestate chunk still 51 bytes.
- `test/tools/joymatrix_identity`: **4 passed**, digests unmoved — `0xC24DDCDE`, `0xD0CD02E2` (also with a port-2 mouse), `0x5E1858EA`. That is the proof the union did not perturb the pad-only path, which matters here because `InputDevFeedLightgun()` writes into the shared matrix slots.
- `bash scripts/c89-lint.sh` clean on `libretro.c`, `src/jerry/inputdev.c`, `test/tools/test_lightgun.c`.

PR #491, opened 2026-08-18
test: make scratch paths per-process to survive concurrent suites

JoeMatt:
Fixes a spurious-failure class observed 2026-08-18 while several agent sessions ran `make test` concurrently.

`test/tools/test_runahead_determinism.c` wrote savestates to the fixed path `/tmp/vj_runahead_a.state`. Two suites raced; one deleted the other's file mid-test and the victim failed with `harness: cannot open save state` / `FAIL: retro_unserialize failed`. That reads exactly like a real determinism regression, which is the worst way for a flake to present.

Scratch paths are now per-process (`getpid()`-keyed), matching the pattern the CD tests and `test_texdump`/`test_texreplace` already use. Paths that were already race-safe were left alone: the five CD tests that already key on `getpid()`, and every `mkstemp`/`XXXXXX` template (race-safe by construction — `netlink_latency_test.sh` was investigated during an earlier flake and cleared).

Verification is deferred to CI and to the v3.4.0 release branch: this repo currently has several agent sessions building concurrently, so a local full-suite run is exactly the condition this PR fixes and would not be a trustworthy signal either way.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -968,6 +968,16 @@ else
 # rev (+ -dirty) against this before running -- a stale dylib fails loudly
 # instead of silently testing the wrong code (see scripts/build-id.sh).
 test: export VJ_EXPECT_BUILD := $(shell ./scripts/build-id.sh)
+# Per-process EEPROM fixture scratch: two concurrent `make test` runs
+# (different worktrees/sessions sharing /tmp) would otherwise recompile
+# and overwrite the same generator binary and regenerate/truncate the
+# same fixture ROM out from under each other mid-suite.  $(shell echo
+# $$PPID) is evaluated once, at parse time, in a subshell whose parent
+# is this `make` process -- so every recipe line below sees the same
+# value for the life of one `make test` invocation, and two concurrent
+# invocations get different values.
+test: EEPROM_GEN_TOOL := /tmp/vj_gen_eeprom_test_rom_$(shell echo $$PPID)
+test: EEPROM_FIXTURE := /tmp/vj_eeprom_lifecycle_$(shell echo $$PPID).j64
 test: test/test_dram_timing test/test_cheat test/test_event_queue test/test_jlink test/test_jlink_tcp test/test_jlink_netpacket test/test_uart_loopback test/test_blitter_simd test/test_dsp_mac40 test/test_titledb test/test_titlehook test/test_biosdb \
 		$(TARGET) test/test_m68k_ops test/test_m68k_irq_ssp test/test_gpu_ops test/test_dsp_ops \
 		test/test_dsp_unit test/test_hle_bios test/test_subsystem_init \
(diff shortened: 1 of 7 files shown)


JoeMatt:
Adding the full classification writeup and concurrency-demo evidence for the record (this PR was merged very quickly after push, before I could attach it as the PR body):

## Classification

**(a) Already safe — left untouched:**
- `test_cd_synth_read.c`, `test_cd_synth_cdda.c`, `test_cd_pregap.c`, `test_cd_synth_subq.c`, `test_cd_synth_butch.c`, `test_cart_bios_loader.c`, `test_texdump.c` — already `snprintf(..., "%ld", (long)getpid())`
- `test_wedge_spin.c`, `test_hook_gate.c`'s `log_path`, `test_pertitle_db.c` — already `mktemp`/`XXXXXX` templates (race-safe by construction)
- `jagcd_roundtrip.sh`, `run_avp_fixture.sh`, `run_dl_branch_fixture.sh` — already `mktemp -d` or `$$`-suffixed

**(b) Fixed in this PR:**
| File | What changed |
|---|---|
| `test/tools/test_runahead_determinism.c` | 3 savestate paths (`state_a/b/c`): array-initializer string literals → runtime `char[64]` buffers built with `snprintf(...,"%ld",(long)getpid())`. Existing `remove()` cleanup kept working. |
| `test/tools/test_hook_gate.c` | `statefile` local `const char*` → per-process `char[64]` buffer. Existing `unlink()` kept working. |
| `test/test_cart_needs_bios.c` | 3 fixture ROM paths → per-process buffers; added `remove()` cleanup that didn't exist before. |
| `test/test_fountain_crash.c` | `DUMMY_CART_PATH` (genuinely-written scratch) → per-process buffer + added `remove()` cleanup. `FOUNTAIN_ROM_DEFAULT` left fixed — see below. |
| `test/tools/test_screenshot.c` | Default `--out` path → per-process (still overridable via `--out`). Not part of `make test`, but shares the same `/tmp` collision risk across concurrent agent sessions on the same machine. |
| `test/tools/flicker_detect.c` | Default `--prefix` → per-process (still overridable via `--prefix`). Same rationale. |
| `Makefile` (`test:` target) | The EEPROM lifecycle test compiled `test/tools/gen_eeprom_test_rom.c` to a fixed `/tmp` path, ran it to produce a fixed fixture `.j64`, then read that fixture from a *separate* recipe line (`test_eeprom_read_race`). Two concurrent suites would recompile/overwrite each other's generator binary and regenerate/truncate the shared fixture mid-suite. Fixed with two target-specific variables (`EEPROM_GEN_TOOL`, `EEPROM_FIXTURE`) suffixed with `$(shell echo $$PPID)`, evaluated once per `make test` invocation (the parent of that shell subprocess is the invoking `make` process), so every recipe line in one run agrees and two concurrent runs differ. Cleaned up with `rm -f` after use. |

**(c) Judgement — left alone, not part of `make test`, diagnostic/manual defaults:**
- `test/headless.py --save-dir` default `/tmp/vj_save` — manual/interactive tool, `mkdir` is idempotent
- `test/jaguar-demos/run.sh` `LOGDIR` default — manual demo runner, per-ROM logs for human review
- `test/tools/fb_ab_sweep.sh`, `cart_boot_matrix.sh`, `blit_memo_sweep.sh` — manual sweep/matrix tools, not invoked by `make test`
- `test/acid/Makefile` (`/tmp/acid_baseline.log`, `/tmp/acid_run.log`) — separate `make acid` target, not `make test`
- Doc-comment example paths in `test/tools/vjtrace_smoke.c`, `test/harness/trace_probe.h`, `test/acid/README.md`, `cd_boot_matrix.sh` usage text, `test_eeprom_lifecycle.c`'s usage string — never executed as code

**`FOUNTAIN_ROM_DEFAULT` (`/tmp/fountain_vj.j64`) deliberately left fixed**, despite superficially looking like a MUST-FIX: it's read-only, not scratch. `test_fountain_crash.c` only ever uses it to build a SKIP *message*; the actual live-ROM path always arrives via `argv[2]`, and the `Makefile` checks that exact literal path itself (`if [ -f /tmp/fountain_vj.j64 ]`) before passing it explicitly on the command line. It's a well-known spot for a manually-placed, not-vendored, optional public ROM — changing the default would make it undiscoverable.

No `unlink`/`remove()` added to `test_screenshot`'s or `flicker_detect`'s outputs — those are diagnostic artifacts meant for a human to open afterward; deleting them would break the tools.

## Concurrency demonstration

Ran two full `make TEST_EXPORTS=1 test` suites simultaneously from **separate git worktrees** sharing `/tmp` (same machine, same user, real disk contention):

1. **Pre-Makefile-fix pair** (only the C-side commit applied): both suites completed clean, 0 FAIL, full summary reached in both. Did not reproduce the EEPROM-fixture race — timing-dependent races don't reproduce on demand — but the code-level justification (two suites recompiling/overwriting the same `/tmp/gen_eeprom_test_rom`, regenerating/truncating the same `/tmp/eeprom_lifecycle_test.j64` read by a later recipe line) is real, so the Makefile fix was added anyway.
2. **Post-Makefile-fix pair**: rebuilt both worktrees, reran. One suite completed clean (0 FAIL, exit 0 confirmed via absence of any `make: *** [test] Error N` in the log) and its EEPROM fixture line showed the fix working — the generated and consumed fixture path both carried that invocation's own `make` PID as the suffix, matching on the write and read lines. No leftover `/tmp/vj_*` scratch files after the run.
3. The other suite in that pair hit an **unrelated, one-off flake** in `test/tools/netlink_latency_test.sh`'s `mktemp "${TMPDIR}/vj_nl_lat_XXXXXX.log"` (`mkstemp failed ... File exists`). Root cause undetermined — a true random collision on a 6-char alnum template is implausible, no stale files were found, `/usr/bin/mktemp` is stock BSD. That file is **unmodified by this PR** (already uses the correct `mktemp`/`XXXXXX` pattern) and the sibling suite passed the identical script at the identical wall-clock time, so this isn't a regression from this change. Re-ran that suite alone afterward: passed clean, 0 FAIL, flake did not reproduce.
   - Separately flagged (not fixed here, out of scope): the script doesn't check `mktemp`'s exit status, so a failure there cascades into several confusing `No such file or directory` lines before the real error surfaces.

## Test plan
- [x] `bash scripts/c89-lint.sh` passes on every touched `.c` file; `flicker_detect.c` is on the lint skip list
- [x] No implicit declarations — `<unistd.h>` added to every file that gained a `getpid()` call; `-std=c99` build confirmed no implicit-decl warnings
- [x] `DEVELOPER_DIR=/Library/Developer/CommandLineTools make -j8 TEST_EXPORTS=1 test` exits 0
- [x] Two full suites run concurrently from separate worktrees; both reach completion with 0 FAIL
- [x] No test's assertions/thresholds/digests changed — only path plumbing

Issue #513, opened 2026-08-20
[feat] Team Tap support — the 4-player adapter is not emulated


JoeMatt:
Found while auditing the docs after v3.4.0: the website's front page and "why this core" page both listed the **Team Tap** among emulated peripherals. It is not emulated, and never has been. The claim has been removed from the site; this issue is the actual work.

## Current state, verified in-repo

The joystick interface models exactly **two sockets**:

- `src/jerry/joystick.c` — `joypad0Buttons[21]` / `joypad1Buttons[21]`, with `joypad0Offset[16]` / `joypad1Offset[16]` decoding the JOYSTICK register's row-select nibbles into a per-socket base offset. `joystick_ram[4]`.
- `libretro.c` registers controller types for ports 1 and 2 only (`virtualjaguar_p1_device`, `virtualjaguar_p2_device`).

So there is no third or fourth pad anywhere in the core, and `RETRO_ENVIRONMENT_SET_CONTROLLER_INFO` advertises two ports.

Our own [`docs/lightgun-design.md`](docs/lightgun-design.md) already records "**No Team Tap support**" as an explicit non-goal (§4.2), and lists the gun's Team Tap interaction as open question 3 — *"Plausible, not TR10-confirmed by this document. Re-verify before asserting in code."* That question becomes live the moment this is implemented.

## What the work is

The Team Tap multiplexes additional pads onto one controller port's row/column matrix, so this is a decode change in `joystick.c` plus two more logical pads — not new register surface. Concretely:

- [ ] **Settle the protocol against TR10 first.** How the adapter presents four pads through the 4-to-16 mux, what a game writes to select among them, and how a title detects a Team Tap is present. `docs/jtrm-jerry.md` §"Joystick Interface" and `docs/jtrm-register-map.md` §"Joystick Registers" document the base mux; they do not document the adapter. **Do not implement from a wiki page or from memory** — that is the standing rule in this repo, and the lightgun work is the precedent for why.
- [ ] Extend the socket decode to N pads per port, keeping the two-pad path bit-identical when no Team Tap is selected (the `test/tools/` identity guardrail from #446 is the model — `$F14000`/`$F14002` must read the same byte-for-byte with the feature off).
- [ ] Expose it as a per-port device type, defaulting off, consistent with how mouse/rotary/analog/lightgun are offered.
- [ ] Advertise the extra ports through `SET_CONTROLLER_INFO` so frontends can bind pads 3 and 4.
[...]

PR #527, opened 2026-08-20
feat(input): Team Tap 4-player adapter (#513)

JoeMatt:
Closes #513. Implements the Team Tap 4-player adaptor, from the TR10 protocol established in the spike (#521).

**Shipping default-off.** This was parked as a draft on a risk/reward call — two known retail titles against a `STATE_VERSION` bump, a widened button array and a decode rewrite. That call has been reversed: the goal is a best attempt at every potential controller, shipped off by default, documented honestly, with users and developers invited to test. The scope statement below is unchanged, because the scope has not changed — only the decision about whether to ship it.

## Scope — read this before the code

- **Known retail support is exactly two titles**: *White Men Can't Jump* (**requires** the adaptor for 3- and 4-player games) and *NBA Jam Tournament Edition* (optional). Both come from the hardware's documented history, not from anything measured here.
- **Homebrew support is unestablished.** The spike's 27-ROM static scan produced **no negatives** — about half the corpus composes row codes in registers and is UNRESOLVED, and that half includes both titles above.
- **No game has been validated in-game.** The only program confirmed to drive the adaptor in this core is the PD Joypad-TeamTap Tester (Domin, 2000). It proves detection answers correctly; it does not prove any game plays correctly through the adaptor.
- **Default is off** (`auto` → plain joypad), with no per-title auto-select, and the tap-absent path is proven bit-identical to `develop` (evidence below).

`docs/input-devices-user-guide.md` carries all of the above plus an explicit call for testing: what to try, what "working" looks like, and what to put in an issue.

## Decode

`joypad0Offset[16]` / `joypad1Offset[16]` become socket+row tables covering all sixteen row codes, per TR10 p.18. Each port's button array grows 21 → 4×21 with **socket 0 keeping slots [0..20]**, so every existing reference — `update_input`, the `$F14002` row table, the guardrail, the export list — indexes exactly the bytes it always did. `base = socket*21 + row*4`.

The socket-0 column arrived pre-cross-checked: the four non-`0xFF` entries of the removed tables sat at exactly the four socket-0 nibbles, at exactly the right rows.

**Detection** is one term at one (socket, row): socket 3 row 1 clears `B0` (port 1) / `B2` (port 2). Nothing else touches those bits in row 1 — which is why the no-tap answer was already correct before any code existed for it.

**Non-pad devices in tap sockets are excluded.** Sockets 1–3 pass `0xFF` to every `InputDev*` call — byte for byte what those twelve codes passed before — and `apply_port_device()` resolves device and adapter together in one function both callers go through, so they cannot disagree.

**Frontend:** new `teamtap` value on both `virtualjaguar_pN_device` options plus a `RETRO_DEVICE_JOYPAD` subclass; eight ports advertised statically. Frontend ports 1/2 keep their meaning, extras are 3–5 (Jaguar port 1) and 6–8 (Jaguar port 2), so one tap on port 1 puts four players on frontend ports **1, 3, 4, 5**. Per-port remapping stays socket-0 only.

## Coexistence with the paddle ADC (#544)

#544 landed on `develop` while this was parked and touches the same option keys and the same device-resolution switch. The two are **mutually exclusive by construction**, in both directions:

- One option key per port resolves both, through a single `else-if` chain — `teamtap` and `paddle` are different values of the same key, so no configuration can select both.
- On the frontend path they arrive from different directions, so `apply_port_device()` resolves them together: `if (type != INPUTDEV_PAD) teamtap = false;`. `INPUTDEV_PADDLE` is not `INPUTDEV_PAD`, so selecting the paddle detaches the tap. This is correct for the hardware too — a Team Tap occupies the physical port the ADC reads.
- Device-type IDs do not collide: the paddle is `RETRO_DEVICE_SUBCLASS(RETRO_DEVICE_ANALOG, 2)`, the tap is `RETRO_DEVICE_SUBCLASS(RETRO_DEVICE_JOYPAD, 0)`, and `RETRO_DEVICE_JAG_PAD` is the *base* `RETRO_DEVICE_JOYPAD` rather than subclass 0.

## Savestate: `STATE_VERSION` 13, appended after the paddle chunk

- **13 is still correct after the rebase.** `v3.4.0` shipped 12 and `develop` is *still* 12 with `CORE_BASE_VERSION := v3.4.0`, so this release cycle has not bumped yet — one bump per release, and this is it. #544 deliberately avoided a bump by trailing its chunk under the existing v12 gate; that stays valid because this chunk is appended *after* it.
- **Chunk order is paddle → Team Tap, identical in `retro_serialize` and `retro_unserialize`.** Every v12 blob `develop` writes ends with the paddle chunk, so appending here keeps those loadable: a v12 state stops after the paddle and takes the `JoystickTeamTapStateReset()` branch.
- **Trailing chunk.** `test/test_state_compat.c` synthesises its v1/v2/v3 fixtures by locating module blocks inside a live blob and splicing; the joystick chunk is mid-blob, so growing it by 126 bytes would have invalidated every fixture. `JoystickStateSave/Load` stay byte-identical, length pinned to `JOYPAD_SOCKET_SLOTS` rather than `sizeof`.
- Adapter **presence** is deliberately not serialized — user configuration, same rule as `InputDevType`.
- Margin, since two branches appended trailing chunks to a fixed buffer this cycle: `[state] v13 payload 2397109 / 2621440 bytes (224331 free)`. Informational — it provably fits, because a `retro_serialize` overrun returns false and `test_state_compat`, `test_runahead_determinism` and `teamtap_state_roundtrip` all pass at 2,621,440 bytes.

## A merge hazard this rebase caught

`retro_controller_info` carried hand-written count literals (`{ port1_devices, 6 }`, `{ port2_devices, 8 }`). Those sat *outside* the conflict region, so git merged both branches' new array entries in and kept the old counts — silently truncating **Light Gun** off port 1 and the just-shipped **paddle** off port 2. Frontend-only symptom; no build, lint or headless test can see it. Both are now `ARRAY_SIZE(...)`, which kills the class rather than the instance.

**This fix is not covered by any test, and the green suite below does not validate it** — the same blindness that let the bug through applies to the fix. The only check that exists is reading the diff: `port1_devices` should have 7 entries and `port2_devices` 9, and both counts are now derived rather than written down. Worth a reviewer's eye, since it is the one change here that affects a device (`paddle`) this PR does not otherwise touch.

## Evidence, re-measured on the rebased tree

Pre-rebase validation proves nothing after 44 commits of `develop`, so everything below was re-run.

### Tap-absent is bit-identical to `develop` — measured, not asserted

The load-bearing guardrail: an adaptor nobody selected must change nothing. Proven by building clean `libretro/develop` (`eef318d`) and this branch with the same flags and diffing per-frame framebuffer-hash CSVs (`test/tools/frame_hash_ab`, 600 frames each, no tap selected):

| ROM | `develop` vs branch |
|---|---|
| `test/roms/yarc.j64` | **identical** (`cmp` clean, 600 rows) |
| `test/roms/jagniccc.j64` | **identical** |
| Joypad-TeamTap Tester (plain dump) | **identical** |

With controls, because "identical" is only meaningful if the instrument can move:

- **Sensitivity floor:** the niccc CSV holds **443 distinct** framebuffer hashes over 600 frames — this is not a constant signal.
- **Positive control:** the same run on the tester ROM *with* `virtualjaguar_p1_device=teamtap` **differs** from the tap-off run. The measurement detects the feature when it is on, and reports no change when it is off.

### Decode digests — unchanged, not re-anchored

All five `joymatrix_identity` digests come back at their originally-recorded values after the rebase. Nothing was re-baselined:

```
PASS [joymatrix_identity_full]              0xC24DDCDE (expected 0xC24DDCDE); 524288 reads
PASS [joymatrix_identity_port1]             0xD0CD02E2 (expected 0xD0CD02E2)
PASS [joymatrix_identity_port1_with_mouse]  0xD0CD02E2 (expected 0xD0CD02E2)
PASS [joymatrix_identity_full_with_mouse]   0x5E1858EA (expected 0x5E1858EA)
PASS [teamtap_full_digest]                  0x480880C4 (expected 0x480880C4); all 84 slots/port seeded
PASS [teamtap_detach_restores_identity]     0xC24DDCDE / 0xD0CD02E2
```

The two behavioural subtleties are still covered: `teamtap_socket_readable` pins the socket-visibility change, and the tap-fitted sweep runs **all 256 row-byte values** — so every combination of simultaneously-low row-select lines is exercised with a tap fitted, not just the single-line cases.

`teamtap_ports` (frontend→socket fill + the savestate chunk) also passes, including `teamtap_socket0_chunk_intact` — the assertion that the paddle and Team Tap chunks do not overlap.

### Detection matrix — one correction to the earlier record

Re-running the matrix produced a result that **differs from what this PR previously claimed**, so it is stated rather than folded in. The claim was that the plain and `[a1]` dumps both detect and `[a2]` does not. In fact **only the plain dump is a valid instrument at all**:

| dump | bytes | clean `develop` (no tap code) | none | p1 | p2 | both |
|---|---|---|---|---|---|---|
| plain | 176,946 | `0x00` | `0x00` | `0x01` | `0x02` | `0x03` |
| `[a1]` | 176,900 | **`0x01`** | `0x01` | `0x01` | `0x01` | `0x01` |
| `[a2]` | 179,602 | `0x00` | `0x00` | `0x00` | `0x00` | `0x00` |

- **The plain dump answers all four configurations exactly correctly.** That is the evidence.
- **`[a1]` was never detecting.** It reads `0x01` on clean `develop` too — a core with *no Team Tap code whatsoever* — and never varies across configurations. The earlier "`[a1]` detects" reading was a stuck value that happened to look like "port 1 found".
- **`[a2]` reads `0x00` everywhere**, on `develop` as well.

Mechanism confirmed at the byte level: the documented detection opcode `33FC 81FA 00F14000` sits at file offset `0x272` in the plain dump only. `[a1]` has `68A5 13FC 0016...` there and `[a2]` has `0A30 33C0 0003...` — differently-linked binaries in which `$68AA` is simply not the verdict byte. Neither alternate dump's value is changed by any tap configuration, so **neither is evidence for or against the implementation**; only the plain dump should be used, and `test/tools/teamtap_rom_probe`'s header says so.

### Suite

`make TEST_EXPORTS=1 test` → **exit 0**. Provenance, stated exactly: the suite ran at `79e0610`, and the two commits added after it (`ce44918`, `7a4f302`) touch one markdown file and one C comment block respectively — no compiled unit changes, so the result stands at HEAD. The detection matrix above was re-run separately on a clean tree with the build-identity guard active (`VJ_EXPECT_BUILD=ce44918`), which is what caught an uncommitted doc edit and refused to measure against it.

The audio pair ran against real ROMs rather than skipping:

```
=== Clipping check: Skyhammer ===         0.000% saturated, window RMS 3083.3, first audio frame 171
=== Clipping check: Iron Soldier 2 ===    0.000% saturated, window RMS 2598.3, first audio frame 21
=== Clipping check: Atari Karts (negative control) ===  0.000% saturated, RMS 410.6
=== Presence check: Iron Soldier 1 ===          PASS: audio is present and within expected envelope
=== Presence check: Iron Soldier 1 (risc=2x) === PASS: audio is present and within expected envelope
```

Skyhammer's 3083.3 sits on develop's recorded 3079.8; nothing moved. `make -j8` clean, and `scripts/c89-lint.sh` passes on all five touched `.c` files (`libretro.c`, `src/jerry/joystick.c`, `test/tools/{teamtap_ports,teamtap_rom_probe,joymatrix_identity}.c`).

## Not settled

- **Which retail titles use it** — see Scope. A dynamic check is cheap now; out of scope here, and the user guide says so.
- **Light gun behind a tap.** `docs/lightgun-design.md` corrected in three places: TR10 says the adapter's six inputs are wire-ORed and pin 6 is `B0/LP`, so `LP` **survives** the adapter but is **not isolated** from other sockets' `B0`. Nobody has documented the actual behaviour, so gun and tap stay mutually exclusive and no code comment claims immunity.
- **RetroArch's default device type for advertised-but-unconfigured ports 3–8 is unverified** — frontend behaviour, untestable headlessly. The guide carries a "check Controls if the extra pads do nothing" note rather than a claim, and asks testers to report their frontend version.


JoeMatt:
**Parked as draft — maintainer call, not a defect.**

Known retail users are **White Men Can't Jump** and **NBA Jam T.E.**; homebrew likely uses it too but that is unestablished. Against two known titles, this PR carries a `STATE_VERSION` bump to 13, a widened per-port button array, eight advertised frontend ports, and a decode rewrite — a lot of surface for a small payoff, and the savestate bump in particular is the kind of thing you only want to spend once per release and for a good reason.

Nothing here is wrong: CI is green, the three develop-measured identity digests come back byte-identical, and the real tester ROM reports all four tap combinations correctly. It is purely a risk/reward decision to take this late in the cycle or not at all.

I stopped the corpus-wide usage investigation rather than spend more on establishing a title list that would not change the answer.

If it is picked up later, two things from the review are worth re-reading first: the `[a2]` dump of the TeamTap Tester does **not** detect while `[a1]` and the plain dump do (different binary, 179,602 vs ~176,900 bytes — so likely a different build, but unproven), and RetroArch's default device type for advertised-but-unconfigured ports 3–8 was never verified because no harness can test frontend behaviour.

Issue #534, opened 2026-08-20
[perf][tests] Deterministic A/B perf measurement in CI via cachegrind instruction counts


JoeMatt:
Sub-issue of #509. Companion to #533 (benchmark ROM choice) and the first consumer is #532.

## Problem

Every perf number this project produces currently depends on a quiet developer laptop. That dependency has burned us twice:

- #515 measured -O3 at **+12.5% sequentially and -11.7% interleaved** on the same two binaries, same machine, minutes apart. The sign flipped because of host drift.
- The #517 and #520 agent runs both had to gate their measurements on load average, and #520's Phase B was **blocked outright** (load 19.08 on 12 cores) so the dispatch-vs-body timing split never ran.

`test/tools/opt_ab.sh` correctly refuses above 1 load per core. That is the right guard, but it means measurement is unavailable exactly when the machine is busy — which, on a dev box running agents, is most of the time.

## The idea: count work instead of timing it

**This emulator is deterministic.** That makes retired-instruction counting viable, and `valgrind --tool=cachegrind` produces **bit-identical** counts across runs and across machines for the same binary. Not low-variance — zero-variance. It is what LLVM, CPython (pyperformance) and Rust use for perf-regression CI for exactly this reason.

Critically it **needs no privileges**, so it runs on free GitHub-hosted runners. `perf stat` mostly does not: hosted runners keep `perf_event_paranoid` locked down.

Consequence: an A/B that needs no quiet host, runs on every PR, and cannot flake.

## Scope

- A script (`test/tools/ir_ab.sh` or similar) that builds both arms **in one session**, runs each under cachegrind against a fixed ROM + frame count, and reports the Ir (instructions retired) delta plus D1/LL miss counts (also deterministic, simulated).
- A workflow that runs it on PRs touching hot paths (`src/tom/gpu.c`, `src/jerry/dsp.c`, `src/tom/blitter.c`, `src/m68000/`, the Makefile).
- A flagging threshold. Because variance is zero, this can be tight — a >0.5% Ir change is real, not noise.
- Cachegrind is ~50-100x slower, so use a short frame count (60-100). Determinism means fewer frames costs confidence in *coverage*, not in *precision*.

## Known limitation — state it in the docs, do not paper over it

[...]

Issue #533, opened 2026-08-20
[perf][tests] yarc.j64 is unrepresentative as the default benchmark ROM — its GPU is 73% spin loop


JoeMatt:
Surfaced by the #520 opcode census. This is a **measurement-methodology** issue: it affects the trustworthiness of every perf number this project has produced or will produce, including how #532 gets validated.

## The finding

`make benchmark` defaults to `test/roms/yarc.j64` (600 frames), and #515's -O3 measurement used the same ROM. The #520 census (7 titles x 900 frames, 1.26 B GPU ops / 2.37 B DSP ops) found **yarc's GPU time is 73.3% `jr`** — a spin loop.

So the default benchmark spends most of its GPU budget on a branch that does no work. A change that speeds up real interpreter workloads can look flat on yarc, and a change that only helps spin loops can look like a win.

## Second-order consequence

#510's perf-interface reading — the one that made the GPU look dominant and the DSP negligible — was taken on yarc. The census says the split is **GPU 47.2% / DSP 34.9%** of process time, so the DSP is not negligible and is often the larger workload by instruction count. Any future 'the GPU is the target' reasoning should cite a corpus number, not a yarc number.

## What to do

- Pick a small benchmark set that reflects real workloads rather than one ROM, or at minimum swap the default to a title whose GPU actually works. The #520 census data is the input for choosing.
- Keep a fixed, documented set so numbers stay comparable commit-to-commit (the whole point of `make benchmark`).
- Public-ROM-only for anything CI touches; the private corpus can back the deeper sweep.
- Note this does **not** invalidate #515's -O3 result — that was a build-wide flag change confirmed independently on-device (A10X, ~50 to 60 fps). It does mean a *targeted interpreter* change measured only on yarc would be untrustworthy.

## Related

#532 (loop-invariant hoisting) is the first change that will need a trustworthy benchmark, so this is worth settling first or alongside.

JoeMatt:
**Before writing anything new — check what is already here.** Two leads found while filing #534:

[...]

Issue #520, opened 2026-08-20
[perf] Interpreter dispatch: what is actually left after direct threading (context-threading assessment)


JoeMatt:
Sub-issue of #509. Raised from a maintainer suggestion to look at **context threading** (Zaleski, Brown & Stoodley, *"Context threading: a flexible and efficient dispatch technique for virtual machine interpreters"*, CGO 2005 — [IEEE](https://ieeexplore.ieee.org/document/1402073), [PDF](https://www.cs.toronto.edu/syslab/pubs/demkea_context.pdf)). The paper reports ~25% mean execution-time reduction for Java and 19–37% for OCaml, from reduced branch hazards.

Filing the assessment rather than the task, because the first two things you need to know change what the task is.

## 1. We already do direct threading. That win is taken.

`src/tom/gpu.c:1556` `executeOpcode()` and `src/jerry/dsp.c:1154` `dsp_executeOpcode()` already dispatch through a computed-goto table:

```c
static const void *gpu_dispatch[64] = { &&gpu_op_add, &&gpu_op_addc, ... };
goto *gpu_dispatch[index];
gpu_op_add:  gpu_opcode_add();  return;
```

and the function-pointer table is explicitly disabled at the call site:

```c
#if 0
      gpu_opcode[index]();
#else
      executeOpcode(index);
#endif
```

So the "a single switch confuses the branch predictor" problem the paper opens with is **already solved here**, for both RISC interpreters, gated on `__GNUC__` with the switch path presumably still reachable for MSVC. Any proposal framed as "replace the switch dispatch" is proposing work that was done.

## 2. Context threading needs runtime code generation — which is exactly what iOS/tvOS forbids

This is the part worth stating plainly, because the popular summaries of the paper get it backwards and it would send someone down a dead end aimed at our worst platform.

Context threading builds a **Context Threading Table**: a buffer of native `call` instructions, one per guest instruction, which the CPU then executes so its return-stack predictor can track the flow. Writing native call opcodes into memory and jumping into them **is** runtime code generation. It needs W^X to be relaxed, the same entitlement a dynarec needs.

The A10X Apple TV in #509 is precisely where we cannot do that. So context threading as published is **not available on the platform this epic exists for**. Subroutine threading has the same problem for the same reason.

What *is* available without codegen — direct threading and token threading — is what we already have.

## 3. What is actually left on the table

The current threading is one level short. Each label makes an out-of-line call and returns:

```c
[...]

Issue #515, opened 2026-08-20
[perf] Build the release at -O3 — measured +5.5% on macOS arm64, suite green


JoeMatt:
Sub-issue of #509. The core builds every non-MSVC platform at a single global `-O2 -DNDEBUG` (`Makefile`, the `else` branch of the `DEBUG` block). The question is whether `-O3` is worth taking on the platforms we can test heavily (iOS, tvOS, macOS) and on the ones we suspect are slow (PS Vita, Switch).

**This issue exists because the obvious way to answer it gives a confidently wrong answer.**

## Measured 2026-08-20, macOS arm64 (M-series, 8P+4E) — inconclusive, and that is the finding

`test/tools/test_benchmark`, `yarc.j64`, 600 frames, 60 warmup, fast blitter. Two binaries built with a full `make clean` between them (mandatory — see the build-config note below), verified distinct by SHA-256.

| protocol | result |
|---|---|
| **Sequential** — all O2 runs, then all O3 runs (n=7 each) | O3 **+12.5% median, +40.5% best-run** |
| **Interleaved** — A/B/B/A × 8 (n=16 each) | O3 **−11.7% median**, Mann-Whitney **p = 0.55** |

Same binaries, same ROM, same machine, minutes apart. **The effect reversed sign when run order was controlled for.** In the sequential run O2 went first and its last three reps decayed from ~140 to ~75 fps, so the "+12.5%" was measuring drift, not the compiler.

Cause: the host was ~8x oversubscribed (load average **98** on 12 cores — an editor at 639% CPU, a concurrent Xcode build, a simulator runtime). Per-run spread reached **9x** (28–259 fps). No optimisation-level effect of a plausible size is resolvable through that.

**So: no evidence either way yet.** Not "O3 doesn't help" — the experiment could not see.

## Protocol for whoever redoes this

- **Interleave the arms.** A-then-B is invalid on any machine that drifts, and every laptop drifts. A/B/B/A cancels linear drift and pair-order bias.
- **Check the load average first** and abort above ~1 per core. This is cheap and would have saved the first two attempts.
- **Report a significance test, not a mean.** With this variance, a difference under ~20% is unresolvable at n=16.
- **Build each arm with a full `make clean`.** The `.build-config` stamp tracks `BUILD_AXES`, and the optimisation level is *not* one of them, so an incremental rebuild silently mixes objects — exactly the chimera-binary class #457 was filed for.
[...]

PR #539, opened 2026-08-20
perf: deterministic instruction-count A/B in CI (#534) + hoist GPU loop invariants (#532)

### LINKED ISSUE #532: [perf] Hoist loop-invariant globals out of the GPU interpreter loop
Split out of the #520 assessment, which measured the thing this depends on. Sub-issue of #509.

## Why this and not opcode inlining

#520 measured dispatch at **9-11% of interpreter time** across four build/workload combinations — so candidate #1 from that ticket (inlining opcode bodies into their threading labels) targets a ~10% bucket and is not worth the churn. This is candidate #2, which is narrower and better-founded.

## The finding

`vjs.gpuPipelineTiming` and `riscClockScalePct` are written **only** from `check_variables()`, never from emulation code, so they are loop-invariant across a GPU slice. But `executeOpcode` is an opaque (non-inlined) call, so the compiler must assume it may clobber them and reloads both every emulated instruction — visible in the disassembly as `ldrb w9, [x28, #0x4]` immediately after the `bl`.

Measured at **49.4% of GPU loop bookkeeping ≈ 8.8% of process time**. Treat that as an **upper bound, not a bookable number.**

**The DSP does not transfer.** Its hot loop lines are genuinely mutable state; only `dsp_exec_opcode_count++` (12.5% of its bookkeeping) is hoistable.

## Acceptance test must be disassembly-level, and must come first

`GPUExec` already saves x19-x28 with globals held in seven of them. Caching these in locals may simply **spill** rather than help. So:

1. Disassemble and confirm **both loads actually left the loop**. If they did not, stop — there is nothing to measure.
2. Only then run `test/tools/opt_ab.sh` (A/B/B/A, `make clean` between arms, Mann-Whitney). Note the load-average gate: the box must be under 1 load per core or the script refuses, and rightly so.

## Constraint

Any caching must stay correct across a `check_variables()` mid-run (option change while a game is loaded). Reload at slice entry, not once at init.

## Do not re-derive

#520's comment §4 has the reproduction details: the `noinline` measurement-build transform, and the self-time-vs-inclusive parsing rule for `/usr/bin/sample` output (misreading inclusive as self-time produces a wrong 'call boundary = 23-40%' conclusion). Also recorded there: **clang silently ignores `always_inline` on any function containing an address-of-label**, with no warning — so 'just inline the dispatch function' is dead on arrival.

JoeMatt:
Closes #534. Implements #532. Answers #533.

Three things, in the order they had to happen: an instrument, a check that the instrument works, and only then a measurement.

## 1. The instrument — `test/tools/ir_ab.sh` (#534)

Counts retired instructions (Ir) under `valgrind --tool=cachegrind` instead of timing anything. The emulator is deterministic, so a fixed ROM and frame count is a fixed amount of work, and cachegrind counts it in simulation — bit-identical run to run, on a busy box or a shared runner, with no privileges (unlike `perf stat`, which hosted runners lock down).

That matters because this project's wall-clock numbers have been wrong twice: #515 measured `-O3` at **+12.5% sequentially and −11.7% interleaved from the same two binaries**, and #520's timing phase was **blocked outright** at load 19.08/12 and never ran.

Four modes:

```bash
test/tools/ir_ab.sh 'OPT_LEVEL=-O2' 'OPT_LEVEL=-O3'   # flag flip (opt_ab.sh's convention)
test/tools/ir_ab.sh --ref libretro/develop HEAD        # source change / PR vs base
test/tools/ir_ab.sh --selftest                         # instrument check
IR_ROM=test/roms/yarc.j64 test/tools/ir_ab.sh --profile   # per-function Ir for one workload
```

Design points worth knowing:

- **No load-average gate, deliberately.** `opt_ab.sh` needs one; this is the tool that exists *because* that gate makes measurement unavailable exactly when the box is busy. The header says so, so nobody "fixes" it later.
- Each arm builds in its **own pristine source tree**, never the working tree — stronger than `make clean`, which the repo's own notes record as insufficient when a compile-affecting switch is missing from `BUILD_AXES`. Both libraries are SHA-256'd and the run **aborts** if they match, because a change that never reached the compiler cannot be A/B'd.
- The workload driver is `frame_hash_ab`, which has no wall-clock read in its output path (`test_benchmark`'s FPS `printf` would inject noise into Ir), and its per-frame hash CSV **doubles as an accuracy gate**.
- The runner binary is built **once, from arm A**, and used for both arms. The instrument must not vary.
- Determinism also needs identical argv and environment, so the measured process runs under `env -i` and the two arm paths are the same length by construction.
- On macOS (no valgrind on Apple Silicon) the build+measure half re-execs inside a Linux container automatically.

`.github/workflows/ir-ab.yml` runs it on PRs touching `gpu.c`, `blitter.c`, `op.c`, `dsp.c`, `src/m68000/**` or the Makefiles. Zero variance means a tight 0.5% threshold. **Advisory, not blocking** — see the caveats below.

## 2. Validating the instrument, in the order #534 specifies

| property | method | result |
|---|---|---|
| **Determinism** | same source both arms | **Ir identical** — 1,633,893,288 twice on one workload, 19,539,824,849 twice on another. Not close: identical. I1/D1/LL miss counts identical too. |
| **Sensitivity** | throwaway commit adding one `volatile` increment per emulated GPU opcode | **+1.815%** (+159,337,672 Ir), correct sign |

The sensitivity control is the one that matters — an instrument that has never been shown to *detect* anything reports "no change" forever and looks healthy. It also calibrates the tool: the waste is exactly 3 instructions per GPU opcode, so it says yarc@90 executes **53,112,557** GPU opcodes, which converts later deltas into per-instruction arithmetic instead of percentages. The throwaway commit was deleted, not merged.

### What Ir is not — stated here, not buried

Ir is **not wall time**. It models neither branch prediction nor ILP nor real cache behaviour. A change can legitimately *raise* Ir and still be faster (inlining, unrolling), and removing dependent loads helps latency more than instruction count — so this class of change is **understated** here. Absolute Ir is architecture-specific; only the within-run delta means anything. The denominator includes the harness (~4.4% of total on the default workload), so an `ir_ab` percentage is not directly comparable to a sampling profiler's "% of process time".

## 3. The ROM question (#533) — it is a frame-count question too

`--profile` gives per-function Ir with no source changes and no sampling bias:

| workload | GPU interp | DSP interp | blitter | 68K | total Ir |
|---|---|---|---|---|---|
| `jagniccc.j64` @ 90 | **0%** | 0% | 0% | ~38% | 1.6e9 |
| `jagniccc.j64` @ 300 | 38.6% | 22.3% | 7.1% | ~5.9% | 19.5e9 |
| `jagniccc.j64` @ 600 | 40.2% | 23.7% | 10.3% | ~2% | 52.3e9 |
| `yarc.j64` @ 90 | **71.2%** | 0.08% | 17.5% | ~0% | 8.8e9 |

`jagniccc.j64` **is** the Jaguar port of Leonard/Oxygene's NICCC 2000 (source in `test/vendor/JaguarDemos/jagniccc2000` — readme confirms the lineage, and the 68K sits in `stop #$2000` while a GPU program rasterises 75,340 precalculated polygons). So the hypothesis was right — but only after the boot window. **At 90 frames it executes zero GPU-interpreter instructions.** A CI gate on that window would have reported 0.0% forever and looked perfectly healthy, which is precisely the failure mode #534 asks to rule out.

300 frames is the cheapest count that reaches the steady-state mix and the only public workload covering every path in the workflow's `paths:` filter, so it is the default here and pinned explicitly in the workflow.

yarc stays documented as the GPU amplifier, with #533's caveat attached: its `jr`-heavy mix makes per-instruction loop *overhead* an unusually large share of its GPU time, so a loop-bookkeeping change scores high on it relative to a real game. Use it to make an effect visible, not to size it.

## 4. #532 — the disassembly gate first, then the measurement

#532 mandates a disassembly-level acceptance test before any benchmarking, because `GPUExec` already holds globals in x19–x28 and two more long-lived locals could simply **spill**. Apple clang, arm64, `-O3`, loop body measured from the back-edge:

| | instructions | loads | stack refs |
|---|---|---|---|
| before | 99 | 19 | 0 |
| after | 94 | 15 | 0 |

Both `ldrb w9, [x28, #0x4]` (`vjs.gpuPipelineTiming`) and the GOT-indirect `ldr x9,[x9]; ldr w9,[x9]` pair (`riscClockScalePct`) **left the loop**; the values live in w26/w27 for the whole slice. **No spill appeared** — zero `[sp,` references in either loop body. clang additionally collapses the two `pipeTiming` tests into a single branch, which the ticket did not anticipate.

Measured (`ir_ab.sh --ref`):

| workload | Ir delta | note |
|---|---|---|
| yarc @ 90 (GPU 71%, `jr`-heavy) | **−4.333%** | upper bound |
| jagniccc @ 300–600 (GPU ~40%, real rasteriser) | **−3.766%** | representative |

I1 misses drop 16.7% on yarc — the loop's instruction footprint shrank. D1/LL are flat, as expected: no data-access pattern changed.

The two instruments agree quantitatively. −397,589,757 Ir over the sensitivity-derived 53,112,557 GPU opcodes is **7.49 instructions saved per emulated GPU opcode**, against 8 predicted by hand-counting the clang loop body (the shortfall is iterations that take the non-default fetch/stall paths). Two independent measurements landing within 7% of each other is a much stronger result than either percentage alone.

Both are lower than #520's 8.8% upper bound, which is expected and consistent: that figure was sampled *time*, this is *work*, and a dependent-load elimination is exactly the case where Ir understates.

## Verification

- `make -j8` clean; `bash scripts/c89-lint.sh src/tom/gpu.c` passes.
- `make TEST_EXPORTS=1 test` exits 0, with the audio pair confirmed **by name** rather than by exit code: `Clipping check: Iron Soldier 2` → PASS (0.000% saturated, RMS 2598.3), `Presence check: Iron Soldier 1` → `PASS: audio is present and within expected envelope`, plus the `risc=2x` presence arm. Only 2 skips, both optional-asset (`no /tmp/fountain_vj.j64`, `no CHSE-capable chdman`).
- `test/tools/dram_scale_sweep.sh` — the mandatory gate for a `gpu.c` timing-model change — **PASS at every scale 1..16**, exit 0. A pure hoist should be timing-neutral; this is the evidence.
- `test/regression_test.sh` — 10 passed, 0 failed, screenshots unchanged, both before and after.
- **Accuracy closure.** The default-config CSV identity only proves the hoist didn't break the path it doesn't touch: with `gpuPipelineTiming` off and `riscClockScalePct == 100`, the modified branches are never taken. So both arms were also run with `gpu_pipeline_timing=enabled`, `risc_clock_scale` at 0.5x and 2x, and both together — **frame-hash CSVs byte-identical in all five configurations** (401 rows each), from two demonstrably different binaries.

## Not in this PR

`--profile` surfaced something bigger than #532 and out of scope here: `M68KInstructionHook()` runs on **every** emulated 68K instruction in shipped builds and calls `m68k_get_reg()` 17 times to fill a traceback ring. On the jagniccc boot window that is 22.9% + 14.7% = **~37.6% of total process Ir**. It is not pure debug (the CD HLE and NVM BIOS hooks ride on it), and the share collapses on GPU-heavy workloads where the 68K is parked — so it needs its own measurement before anyone touches it. Filed as #540 rather than bundled here.

### DIFF
--- a/.github/workflows/ir-ab.yml
+++ b/.github/workflows/ir-ab.yml
@@ -1,0 +1,131 @@
+name: Instruction-count A/B
+
+# Deterministic perf regression gate (issue #534).
+#
+# Counts retired instructions (Ir) under valgrind's cachegrind for the PR head
+# and for its merge base, and reports the delta.  cachegrind simulates, so the
+# count is bit-identical run to run and machine to machine for a given binary
+# and architecture -- which is exactly what a shared, noisy GitHub-hosted
+# runner cannot give a wall-clock benchmark.  It also needs no privileges,
+# unlike `perf stat`, which hosted runners keep locked down via
+# perf_event_paranoid.
+#
+# Because the variance is zero, the flagging threshold is tight: 0.5%.
+#
+# This job does NOT fail the PR.  Ir is a work count, not a clock: a change
+# that removes dependent loads shows a small Ir delta and a larger wall-clock
+# win, and inlining/unrolling can raise Ir while getting faster.  So the
+# summary is advisory -- a prompt to look, not a veto.  Set IR_GATE=1 below if
+# a future policy decision makes it blocking.
+#
+# Only within-run A/B deltas are meaningful.  Absolute Ir is architecture
+# specific; never cache a baseline count without keying it by runner arch.
+
+on:
+  pull_request:
+    branches: [develop, master]
+    paths:
+      # Hot paths only.  A docs or workflow-only PR should not pay for two
+      # full builds plus two cachegrind runs.
+      - 'src/tom/gpu.c'
+      - 'src/tom/blitter.c'
+      - 'src/tom/op.c'
+      - 'src/jerry/dsp.c'
+      - 'src/m68000/**'
+      - 'Makefile'
+      - 'Makefile.common'
+      - 'test/tools/ir_ab.sh'
+      - '.github/workflows/ir-ab.yml'
+  workflow_dispatch:
+    inputs:
+      frames:
+        description: 'Frames per arm'
+        required: false
+        default: '90'
+
+concurrency:
+  group: ir-ab-${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  ir-ab:
+    name: Ir A/B (linux x86_64)
+    runs-on: ubuntu-latest
+    timeout-minutes: 45
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v7
+        with:
+          # ir_ab.sh --ref snapshots both arms with `git archive`, so both
+          # commits must actually be present in the local object store.
+          fetch-depth: 0
+
+      - name: Install valgrind
+        run: |
+          set -euo pipefail
+          sudo apt-get update
+          sudo apt-get install -y --no-install-recommends valgrind
+          valgrind --version
+
+      - name: Resolve arms
+        id: arms
+        run: |
+          set -euo pipefail
+          if [ "${{ github.event_name }}" = pull_request ]; then
+            BASE='${{ github.event.pull_request.base.sha }}'
+            HEAD='${{ github.event.pull_request.head.sha }}'
+          else
+            BASE=$(git rev-parse HEAD~1)
+            HEAD=$(git rev-parse HEAD)
+          fi
+          # fetch-depth: 0 covers the PR ref's history; the base commit can
+          # still be missing on a force-pushed or cross-fork PR.
+          git cat-file -e "$BASE^{commit}" 2>/dev/null || git fetch --no-tags origin "$BASE"
+          git cat-file -e "$HEAD^{commit}" 2>/dev/null || git fetch --no-tags origin "$HEAD"
+          echo "base=$BASE" >> "$GITHUB_OUTPUT"
+          echo "head=$HEAD" >> "$GITHUB_OUTPUT"
+
+      - name: Measure
+        id: measure
+        env:
+          IR_FRAMES: ${{ github.event.inputs.frames || '90' }}
+          # Advisory only -- see the header.  Flip to 1 to make a >0.5%
+          # instruction-count regression block the PR.
+          IR_GATE: '0'
+          IR_THRESHOLD: '0.5'
+          # No load-average gate here, deliberately: Ir does not depend on how
+          # busy the runner is.  That is the entire reason this job exists.
+          IR_NO_CONTAINER: '1'
+        run: |
+          set -euo pipefail
+          bash test/tools/ir_ab.sh --ref \
+              "${{ steps.arms.outputs.base }}" "${{ steps.arms.outputs.head }}" \
+              2>&1 | tee ir-ab.log
+          grep '^IR_AB_RESULT' ir-ab.log >> "$GITHUB_OUTPUT" || true
+
+      - name: Job summary
+        if: always()
+        run: |
+          {
+            echo "## Instruction-count A/B"
+            echo
+            echo "arm A = merge base \`${{ steps.arms.outputs.base }}\`"
+            echo "arm B = PR head \`${{ steps.arms.outputs.head }}\`"
+            echo
+            echo '```'
+            sed -n '/metric/,$p' ir-ab.log 2>/dev/null || cat ir-ab.log 2>/dev/null || echo '(no output)'
+            echo '```'
+            echo
+            echo "Ir is a deterministic *work* count, not wall time. A change that removes"
+            echo "dependent loads understates here; inlining/unrolling can raise Ir and still"
+            echo "be faster. Treat a flagged delta as a prompt to look, not a veto."
+          } >> "$GITHUB_STEP_SUMMARY"
+
+      - name: Upload log
+        if: always()
+        uses: actions/upload-artifact@v7
+        with:
+          name: ir-ab-log
+          path: ir-ab.log
+          retention-days: 14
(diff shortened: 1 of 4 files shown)


JoeMatt:
Addressed all four Kimi findings — thanks, findings 1 and 2 were real bugs, not nits.

1. **Job went red on its own filter.** `ir_ab.sh` exits 2 on byte-identical arms, which is the *expected* outcome for a PR touching only `ir_ab.sh` or `ir-ab.yml` — both of which the `paths:` filter selects. Under `set -e` that failed the step, contradicting the header's "advisory, not a veto". Exit 2 is now a `::notice::` and a pass; every other exit code still propagates (via `PIPESTATUS`, since it runs through `tee`).
2. **macOS container leaked its work dir permanently.** Exactly as described — `exec` destroyed the host EXIT trap, and the container re-enters with `IR_AB_WORK` set so it skips trap setup too. Now runs as a child with its exit code propagated, so the trap fires.
3. **Dead summary fallback.** `sed -n` exits 0 on no-match, so `|| cat` was unreachable precisely when the log tail matters most. Now tests for the pattern explicitly and tails 40 lines on an aborted run.
4. **Docs arithmetic.** Correct — 159,337,672 / 3 is not an integer. Reworded as a calibration figure (~3/opcode, ~53.1M opcodes) and the non-integer called out, since the compiler can fold the increment differently per path.

No change to `src/tom/gpu.c`; the hoist itself was reported clean. Merging without re-review since these are confined to the harness and workflow, per maintainer instruction.

Issue #514, opened 2026-08-20
[feat] Pro Controller support — the six-button pad is not emulated


JoeMatt:
Companion to #513, from the same post-v3.4.0 doc audit: the website listed the **Pro Controller** among emulated peripherals. It is not emulated. The claim has been removed from the site; this issue is the actual work.

## Current state, verified in-repo

The core models the standard pad only. Grepping `src/jerry/joystick.c`, `src/jerry/inputdev.h` and `libretro.c` for any six-button / shoulder-button plumbing returns nothing: the device enum is

```
INPUTDEV_PAD, MOUSE_ST, MOUSE_AMIGA_ADAPTER, MOUSE_AMIGA_ON_ST,
ROTARY, ANALOG, DRIVING, LIGHTGUN
```

with no Pro Controller entry, and `joypadNButtons[21]` covers the standard pad's matrix. The RetroPad's `L`/`R`/`L2`/`R2`/`X`/`Y` ids appear in `libretro.c`'s remap table, but they feed the *existing* Jaguar buttons and keypad — they are not additional Jaguar inputs.

## What the work is

The Pro Controller is a superset of the standard pad — the extra buttons ride the same JOYSTICK row/column matrix rather than adding registers — so like #513 this is a decode change, not new hardware surface.

- [ ] **Settle the button-to-matrix mapping against TR10 first.** Which row/column intersections the extra buttons occupy, and how (or whether) a title distinguishes a Pro Controller from a standard pad. `docs/jtrm-jerry.md` and `docs/jtrm-register-map.md` document the base 4-to-16 mux and the standard pad's assignments; neither covers this controller. The repo rule applies: JTRM/TR10 and the jag_sim netlists, in that order — never a wiki page, never source-code comments.
- [ ] Extend the matrix decode, with the standard-pad path bit-identical when the Pro Controller is not selected. #446's `$F14000`/`$F14002` identity guardrail is the pattern to copy.
- [ ] Offer it as a per-port device type alongside the existing ones, default off.
- [ ] Map the extra buttons to sensible RetroPad ids and add input descriptors so the frontend's Controls menu labels them.
- [ ] Confirm whether a title that only knows the standard pad misreads a Pro Controller's extra rows. If so, that is the argument for keeping it opt-in rather than defaulting on.

## Titles

[...]

PR #547, opened 2026-08-20
feat(input): 6D controller, best-attempt (#538)

### LINKED ISSUE #538: [feat] 3D flight stick with 6 degrees of freedom
### What

See the Jaguar Console Hardware Release Notes on p. 20:

https://github.com/darthcloud/BlueRetro/wiki/04_Technical_Reference.pdf

### Why / use case

I don't know if any games support the specification, maybe not, it's an cancelled unreleased controller after all, but would be cool to have for completionism's sake.

### How (if you have ideas)

_No response_

### Pre-flight

- [x] This is libretro-fork-specific, not a general Virtual Jaguar emulation feature (those go upstream).

JoeMatt:
Closes #538.

Atari specified a six-degrees-of-freedom controller in the **Jaguar Technical Reference V10** and never shipped it. **No software anywhere is known to drive it.** This is a best attempt built from the manual alone — conformant to it, and unvalidated against any real program. The docs, the core-option help and the test header all say that plainly rather than implying support.

> **Only TR V10 carries the controller chapter.** `Technical Reference v8.pdf` and Atari's original `04 - Technical Reference.pdf` do not, so anyone verifying against the wrong revision concludes the spec does not exist — close to what happened in #514. Every table here was read as a **rendered page image** (`pdftoppm`), because `pdftotext` scrambles their column order into a vertical jumble.

## What it is

Six 8-bit DOF (X/Y/Z translation, TX/TY/TZ torque — TR10 names them Pitch=TZ, Yaw=TX, Roll=TY), seven buttons A–G and a Rezero control: 55 bits, which does not fit the 24 a standard controller returns, so the device answers the ordinary `$F14000` row scan with **three banks**.

It is the **same wire protocol as #437's analog / driving controller with three banks instead of two**, so it shares that code path rather than forking it. The row-3 → row-0 bank clock is generalised to a bank count; the engagement guardrail, the C-column identification field and TR10's interleave-immunity rule come along unchanged.

The J-line map (TR10 p.22) is a **literal twelve-cell table**, not a shift expression — row 3 of each bank carries a translation axis's *high* nibble while rows 0–2 carry a coherent triple:

| | Row 0 | Row 1 | Row 2 | Row 3 |
|---|---|---|---|---|
| Bank 0 | X lo | Y lo | Z lo | X hi |
| Bank 1 | TX lo | TY lo | TZ lo | Y hi |
| Bank 2 | TX hi | TY hi | TZ hi | Z hi |

Bank 2's B column is the **1,1,1,0 "6D Controller" identifier** from TR10 p.16's last-bank table, whose row-0 zero is a line held low with *no switch behind it* — #437's all-ones identifier needed no such case, which is exactly why it is easy to omit.

## Host interface

All six RetroPad analog signals, one-to-one, nothing doubled up:

| Host | DOF |
|---|---|
| Left stick X / Y | X / Y translation |
| R2 − L2 | Z (thrust) |
| Right stick X / Y | TX (yaw) / TZ (pitch) |
| R − L | TY (roll) |

Shoulder pairs are read as analog *buttons*, so a frontend with real pressure gives proportional roll and thrust and one without gives clean digital roll and thrust — the axis is never lost. Buttons: A/B/C/D on A/B/Y/X, E/F on the stick clicks, G and Rezero on Start/Select.

## Where TR V10 is ambiguous, and what I chose

1. **The X sign.** p.23's prose says *"X is positive right to left"*; **the figure on that same page draws its horizontal axis arrow pointing right.** They contradict each other and the page settles nothing. I followed the **prose**, because it is an unambiguous sentence where the figure carries only a bare `+` glyph. It is one negation in `InputDevFeed6D()` and one assertion in `sixd_decode_test` — a two-line flip if anyone ever proves otherwise.
2. **The three torque signs.** *"Counter-clockwise when facing the positive direction"* is a statement about a physical grip whose orientation relative to the player is undefined. They pass through un-negated. **A named guess, not a derivation.**
3. **Bank 1's button order.** Bank 0 runs A,B,C,D *up* the rows; bank 1 runs Rezero,G,F,E *down* them. Implemented **as printed**. This is the first thing a tester with silicon should check.
4. **Six DOF over two `axistune` slots.** Split by where the host value came from: X and TX take the X-axis tuning rows, everything else takes the Y-axis rows, so "X dead zone" still means "dead zone on horizontal stick motion".
5. **The DOF↔host pairing itself** is a choice, not a spec. TR10 says nothing about what a human holds.

## Four protocol rules deliberately not implemented

Each with its reason, rather than a faked detector:

- **"Output your last bank during controller identification"** (p.21) — an identification read is *indistinguishable on the bus* from a game read. TR10's own driver recipe (read every bank into a table, find bank 0 by its flag bit) works regardless.
- **The 100 µs Socket 0 / Row 0 validity rule** (p.21) — exists so a real microcontroller ignores Boot ROM row codes; inert in a model with no settling time.
- **The ≤50 mA / ≤10 mA current budget** (p.21) — electrical.
- **The DB15 passthrough** (p.21) — mechanical, *but it has a real consequence we inherit*: the 6D bank tables contain **no Pause bit and no Option bit anywhere**, so both are unreachable while the device is engaged. TR10's answer is a physical joypad relayed "as one of its banks", and the 6D layout has no spare bank. Recorded as a limitation; I did **not** invent a fourth bank.

Also inherited from #437: no settling delays (p.28's ~25 µs/row, ~200 µs/bank — a compliant driver's delay loop spins over already-valid data) and an ideal 128-centred range where real hardware drifts.

## Tests

New `test/tools/sixd_decode_test` (in `make test`). It drives **six pairwise-distinct nibble patterns** — X=0x12 Y=0x34 Z=0x56 TX=0x78 TY=0x9A TZ=0xBC — precisely so a transposition of two cells cannot pass; feeding a uniform value would make a bank-0 row-1/row-2 swap invisible. It also pins the three-bank cycle, other-socket interleave immunity, the B column in all three banks, the **port-dependent** C-row mapping (C2 is row 2 on port 1 but row 3 on port 2, per p.15's port-headed tables), TR10 p.23's sign conventions, the engagement guardrail and the savestate.

## Default-off bit identity — measured, not asserted

- `joymatrix_identity`: all four digests unchanged — `0xC24DDCDE` / `0xD0CD02E2` / `0xD0CD02E2` / `0x5E1858EA`. This is the stronger of the two, because it digests the matrix decode itself.
- `frame_hash_ab`, 900 frames, against a build of `libretro/develop` at the rebased tip: **byte-identical** on both in-tree ROMs — yarc `aac09cd4645b96e2798b3e1cd2283aa6`, jagniccc `c97207d60cbbb39d578ae4d937f7ead5`.
- `tuning_identity`, `analog_decode_test`, `paddle_decode_test`, `rotary_decode_test`, `mouse_decode_test` all pass.
- `make TEST_EXPORTS=1 test` exits 0, with the audio pair confirmed **by name**: `Clipping check: Skyhammer` / `Iron Soldier 2` (0.000% saturated, RMS 3083.3 / 2598.3) and two `audio is present and within expected envelope`.
- Both ABI directions build clean (plain → `TEST_EXPORTS=1` → plain). C89 lint passes on every touched `.c`.

## Savestate

The develop-only v12 chunk is extended **in place** to 65 bytes (six DOF bytes plus a widened switch mask per port); no `STATE_VERSION` bump, per the one-bump-per-release policy. The bank clamp on load is against the *currently attached* device's bank count, not a constant — the type is option-derived and deliberately not serialized, so a 6D state can be loaded into a session configured for the two-bank analog device.

## Docs

Extended `docs/input-devices-user-guide.md` rather than adding a competing file. The 6D section documents our interface, what we do not implement, both ambiguities, and ends with an explicit **call for testing** — what to try, what "working" looks like, and a request to file findings on #538, noting that a negative result is exactly as useful and rather more likely.


JoeMatt:
**Rebased onto `faab9f9`** (the Team Tap merge, #527) and re-verified end to end. Three conflicts, all resolved by keeping both features:

- `libretro_core_options.h` — Team Tap's port descriptions kept, the 6D paragraph appended to each.
- `libretro.c` — took develop's `ARRAY_SIZE(...)` controller-info counts over my hand-written literals. #527's own comment predicted this exact collision ("two independent branches each appended one entry while both carried the same hand-written literal"), so the ARRAY_SIZE form now also absorbs the 6D entry with no count to forget.
- `docs/input-devices-user-guide.md` — Team Tap and 6D sections both kept; the option table, the frontend device counts (eight on port 1, ten on port 2) and *Already shipped* now list every shipped device, which corrects two rows that were already stale before this branch.

`src/jerry/inputdev.{c,h}` did not conflict: Team Tap is a `joystick.c` decode change and adds no `InputDevType`, so 6D is still enum value 9 and `INPUTDEV_STATE_SIZE` still moves 51 → 65.

Re-measured on the rebased tip:

- `make TEST_EXPORTS=1 test` exits 0. Team Tap's own suite is green alongside (`teamtap_full_digest 0x480880C4`, `teamtap_detach_restores_identity`, `teamtap_state_roundtrip`), and `sixd_decode` passes.
- `joymatrix_identity` four digests unchanged: `0xC24DDCDE` / `0xD0CD02E2` / `0xD0CD02E2` / `0x5E1858EA`.
- `frame_hash_ab` 900 frames vs a fresh build of `libretro/develop@faab9f9`: byte-identical, yarc `aac09cd4645b96e2798b3e1cd2283aa6`, jagniccc `c97207d60cbbb39d578ae4d937f7ead5`.
- Audio pair confirmed by name (Skyhammer / Iron Soldier 2 clipping, two presence passes).

Also in this push, a second commit: **the two shoulder-derived DOF needed a digital fallback.** Z and TY are read through `RETRO_DEVICE_INDEX_ANALOG_BUTTON`; a frontend without analog-button support answers `0` for a shoulder that is being *held*, which would have left two of six DOF permanently at rest — dead, not merely non-proportional, and invisible to `sixd_decode_test` because that suite calls `InputDevFeed6D()` directly and never enters `update_input`. A zero analog read whose digital bit is down is now promoted to full scale, the same shape as the driving skin's existing `if (r2 || l2)` fallback. The eight switches were checked the same way and need no change: `ret[]` is either a `RETRO_DEVICE_ID_JOYPAD_MASK` read or a loop over `B..R3`, so L3, R3, Start and Select — E, F, G and Rezero — all arrive.

Issue #536, opened 2026-08-20
[perf][tests] Purpose-built microbenchmark ROMs (isolate GPU / DSP / blitter / 68K)


JoeMatt:
Split out of #533. Sub-issue candidate for #509.

## Why this is less speculative than it sounds

The repo **already has a Jaguar ROM build path**: `make jaguar-demos-build` drives lyxass/rmac against the pinned [42Bastian/JaguarDemos](https://codeberg.org/42Bastian/JaguarDemos) sources (`test/jaguar-demos/run.sh build`). So the assembler toolchain question — usually the blocker for 'let's write a test ROM' — is already answered.

## What real titles cannot give us

Commercial and demo ROMs mix all five engines, so a measurement on one is always a blend. #520 hit this directly: the corpus split is GPU 47.2% / DSP 34.9%, and the DSP figure was initially misread as a 75% idle-poll loop purely because the census was taken during menus rather than gameplay. A workload you *wrote* has none of that ambiguity.

A microbenchmark ROM can guarantee properties no real title offers:

- **Single-engine** — a GPU-only loop with the DSP halted, and the converse. Isolates which interpreter a change actually helps.
- **Known opcode mix** — e.g. a loop that is 90% `move`/`add` vs one that is branch-heavy, to test dispatch-vs-body hypotheses (#532) directly instead of inferring them from a census.
- **Terminates deterministically** — runs a fixed instruction count then halts, so Ir counts under cachegrind (#534) are exact and comparable, and there is no 'how many frames is enough' question.
- **No input dependence** — no menus, no attract-mode divergence, no `--press` scripting needed.

## Scope sketch

- A small set (4-6) of `.s` sources in-tree with a build rule, not vendored.
- One per engine: GPU arithmetic loop, GPU branch-heavy loop, DSP loop, blitter fill/copy throughput, 68K loop.
- Each prints or parks a known completion marker the harness can assert on.
- Committed `.j64` alongside source so CI needs no assembler (the JaguarDemos job already takes this approach — 'CI only runs .j64 files already committed').

## Sequencing

**Do #533's measurement of the ROMs we already have first.** `jagniccc.j64` and the JaguarDemos GPU-only carts may cover enough of this for free. This ticket is for the gap that remains after that — which is most likely the *isolation* and *exact-termination* properties, not raw coverage.

Parked in Icebox pending that answer; promote once #533 says what is actually missing.

JoeMatt:
[...]

PR #581, opened 2026-08-25
feat(toolchain): on-demand Jaguar assembler toolchain (rmac/rln, lyxass, pc_jagcrypt, new_bjl)

JoeMatt:
## Summary

Unblocks issue #536 (purpose-built microbenchmark ROMs to isolate GPU/DSP/blitter/68K for perf work) and the previously-no-op `make jaguar-demos-build` target, neither of which could move without a Jaguar assembler toolchain that no session so far had available — `test/jaguar-demos/run.sh`'s `have_jag_toolchain()` only ever checked whether `lyxass`/`rmac`/`rln` happened to already be on `PATH`.

- **`tools/jaguar-toolchain/`**: `PIN` file + `setup.sh` (Linux/macOS) with real, tested `fetch`/`build`/`env` subcommands that clone and build `rmac`/`rln` (68000 asm+link), `lyxass` (GPU/DSP asm), and `pc_jagcrypt` (BootIntro encrypt) from source, matching `test/jaguar-demos/`'s existing on-demand-clone convention — nothing vendored into git history.
- **`have_jag_toolchain()`** in `test/jaguar-demos/run.sh` now also checks the local toolchain build, additively — `make jaguar-demos-build` goes from a permanent skip to actually building (verified end-to-end: confirmed via `make -n jaguar-demos-build`. 38 of 54 JaguarDemos Makefiles still fail downstream on a `jagcrypt -tursi` flag the pinned `pc_jagcrypt` has never supported — a real upstream gap, documented in the README, not something this PR works around).
- **`tools/jaguar-toolchain/setup.ps1`**: Windows port of the same fetch/build/env logic, reading the same `PIN`. Honestly caveated as unverified for a real Windows build (no Windows/MinGW host in this session) — verified for syntax (real `pwsh` AST parse), logic parity, and a live fetch against all 5 real upstreams.
- **`docker/jaguar-toolchain.Dockerfile`** + **`.github/workflows/jaguar-toolchain.yml`**: multi-stage build, now actually validated in CI (`docker build` + smoke-check), plus a real assembly smoke test — two minimal `.s` sources (68K via rmac/rln, GPU/DSP via lyxass) that only assemble successfully after real iteration against the tools' actual syntax; the CI check specifically catches a `lyxass` silent-success failure mode (exits 0, writes a near-empty file when `.run` isn't positioned correctly) that a naive `test -s` would have missed.
- **`.github/workflows/jaguar-toolchain-image.yml`**: publishes the built image to `ghcr.io/libretro/virtualjaguar-libretro-jaguar-toolchain` on `develop` pushes touching the PIN/Dockerfile/setup.sh, using the repo's own `GITHUB_TOKEN` (no new secrets).

Built via subagent-driven development: 9 tasks + one final-review fix wave, every task independently reviewed (spec compliance + quality), several reviewers independently reproduced the implementer's claims against the real tools rather than trusting reports (e.g. the final reviewer built and executed the actual committed CI workflow steps, and confirmed GNU-coreutils portability for the smoke checks in a Debian container).

**Known, deliberately-accepted trade-off**: `rmac`/`rln`/`pc_jagcrypt` have no detectable upstream LICENSE file. This toolchain never vendors their source (clone-on-demand only), but the publish workflow does build and push compiled binaries of them to a public registry. Discussed and accepted — the code is open, sourced in good faith, and Docker publishing is a convenience layer with the from-source path always available as the documented fallback.

## Test plan

- [x] `make -j$(getconf _NPROCESSORS_ONLN)` builds clean (this PR touches zero core `.c` files)
- [x] `make -n jaguar-toolchain-build` / `make -n jaguar-demos-build` — correct command sequences
- [x] Real `setup.sh fetch`/`build`/`env` run end-to-end on macOS arm64: all 4 buildable tools (rmac, rln, lyxass, pc_jagcrypt) produce working binaries; `new_bjl`'s prebuilt binaries fetch correctly
- [x] Real `make jaguar-demos-build` confirmed to move past its previous permanent skip and actually invoke per-demo builds
- [x] Both new CI workflows parse as valid YAML; every shell block passes `bash -n`
- [x] Smoke `.s` sources verified to assemble/link with the real tools, including a negative-control proof that the CI check isn't vacuous
- [ ] Windows `setup.ps1` — unverified on a real Windows/MinGW host (no such host in this session); syntax/logic-parity verified via `pwsh`
- [ ] Docker image build — unverified in this session (no Docker daemon available); now gated by a real `docker build` + smoke-check step in CI, which will exercise it for the first time on this PR


JoeMatt:
Both Kimi findings fixed in 1c310a5:

1. \`.dockerignore\`'s negation chain now uses the correct per-level exclude/re-include pattern (\`tools/*\` → \`!tools/jaguar-toolchain\` → \`tools/jaguar-toolchain/*\` → the two leaf-file negations) instead of a single \`!tools\` that never actually scoped anything.
2. Both workflows' \`paths:\` filters now cover \`docker/jaguar-toolchain.Dockerfile.dockerignore\`, and \`jaguar-toolchain-image.yml\` now covers its own filename too.

Issue #601, opened 2026-08-25
[perf] Capture the on-device A10X per-subsystem breakdown (blocked on hardware access)


JoeMatt:
## Summary

Split out of #509 (performance epic for older/slower platforms), which is now closed — all ten of its sub-issues shipped. This is the one remaining item, and it is **not engineering work**: it is a measurement that requires physical access to an Apple TV 4K 1st-gen (A10X).

## What's needed

A per-subsystem millisecond breakdown captured **on the device**, using the `RETRO_ENVIRONMENT_GET_PERF_INTERFACE` instrumentation that shipped in #510 and the capture tooling from #563.

Desktop profiling says GPU RISC interpretation is ~74% of instrumented time against the blitter's ~17%. That is the current best guess at where an optimization pass should start — but it is a Mac number, and #509's whole premise was that desktop profiling does not predict A10X behavior.

## Why it's blocked

Needs the maintainer's own hardware and a hands-on RetroArch session. Nobody else can run it.

## Two traps already documented (don't rediscover these)

- `perfcnt_enable` off means every counter reads **zero** — indistinguishable from broken instrumentation rather than "disabled".
- RetroArch reports **Avg**, not total, per counter.

## Not milestoned on purpose

Dating this would stall whatever release it's attached to — exactly what it did to v3.5.0. It lands when the capture happens.

## References

- #509 (parent epic, closed as delivered)
- #510 (perf interface), #563 (capture tooling)
- `docs/perf-audit-2026-08.md`

JoeMatt:
Milestoned to **Maintenance** (2026-08-25 reconcile).

This was deliberately left unmilestoned when it was split out of #509, on the rule that undatable work must not hold a version number hostage. That rule is right, but "unmilestoned" was the wrong expression of it — Maintenance and Icebox are both non-versioned, so the rule is satisfied without leaving the issue floating outside the tracker. Maintenance already holds two other `blocked` items (#463, #568) waiting on artifacts we do not control, so this is the consistent home rather than a new precedent.

No change to the substance: still blocked on physical access to an A10X, still not engineering work.

Issue #595, opened 2026-08-25
[3.5.1] Surface or programmatically flag setting combinations that compound badly (overclock disables idle-skip)


JoeMatt:
## Summary

Split out of the v3.5.0 retro on performance feedback (Apple TV 4K 1st-gen / A10X, RetroArch, default settings mostly fine, some titles noticeably worse under overclock). Some core-option combinations are known to compound badly, and right now nothing surfaces that to the user — they just experience "this made it slower than expected" with no explanation.

## Known compounding case (concrete example to design against)

`virtualjaguar_risc_idle_skip` (RISC idle-loop fast-forward, the single biggest perf lever shipped in #569 — 66-87% cut in DSP opcodes/frame on measured titles) **turns itself off entirely** the moment any of these are active:
- `risc_clock_scale` / `m68k_clock_scale` away from stock
- any timing model (`dram_timing`, `gpu_pipeline_timing`, `blitterTiming`)
- `virtualjaguar_blit_memo`
- a single-step or an armed trace/watchpoint

So a user who cranks the overclock slider on a title that's borderline-slow doesn't just pay the overclock's own cost — they also silently forfeit the biggest available speedup, compounding two costs into one confusing "it got much slower" experience with zero indication why.

## Scope

Design and implement one (or both, if they turn out complementary rather than redundant) of:

1. **Surface it in the option text/UI** — e.g. `risc_idle_skip`'s description (or the overclock options' own description) states plainly that overclock/timing-model/blit-memo/tracing disable it. Cheapest, no logic change, but passive — a user has to read it.
2. **Programmatic warning** — when `check_variables()` detects a combination known to disable idle-skip (or any other future known-bad combination) while the user has explicitly enabled the thing that gets disabled, log a `[perf]`-tagged warning the same way `titledb`'s known-bad refusal already does (`get_variable_pertitle()` in `libretro.c`) — same established UX pattern from #464, reused for a different problem (perf conflict, not correctness).

## Open question

Is `risc_idle_skip` vs. overclock the ONLY such conflict worth covering in 3.5.1, or should this issue also inventory other option pairs that are quietly redundant/counterproductive together (e.g. does `virtualjaguar_true_color` interact with anything, does hi-res interact with texture replacement in a costly way)? Scope that inventory as part of this issue's Phase 1 before implementing Phase 2 (the actual warning/UI mechanism).

## References

- #569 (perf audit — idle-loop fast-forward's disable conditions)
[...]

PR #603, opened 2026-08-25
docs: settings and performance tuning guide, plus a site page

JoeMatt:
Closes the "nothing explains which of the ~40 options matter" gap. Adds `docs/settings-and-performance-guide.md`, a condensed `site/pages/performance.html`, and two README pointers.

## Why it's worth reviewing rather than skimming

The performance section is sourced, not folklore, and it corrects two claims that were circulating in our own issue text:

- **DSP idle-skip does NOT disable itself under `m68k_clock_scale` or `blitter_timing`** — only under RISC clock scale, `dram_timing`, `gpu_pipeline_timing`, and `blit_memo`. Verified against the gate in `src/jerry/dsp.c`, not taken from #595's body (which had it wrong; corrected there now).
- **The "GPU RISC ~74% / blitter ~17%" split is not in any committed doc.** It traces to notes about an uncaptured device run. The guide uses only doc-backed figures from `docs/perf-audit-2026-08.md` / `docs/op-perf-profile.md`, and labels them explicitly as host arm64 profiles rather than device measurements — no on-device A10X capture exists yet (#601).

## Content

- Defaults are the right answer; a "safe to ignore forever" table
- Per-title enhancement defaults, and that a user-set value always wins
- Enhancements: presentation-only guarantee stated once, not per option
- **Performance**: where frame time goes, an ordered what-to-try list that *ends* at overclocking rather than starting there, the compounding trap, low-end platform notes
- Timing models, Jaguar CD boot mode, restart-required table, symptom→setting troubleshooting table

## Verification

`python3 scripts/build_site.py` builds clean — the new page lands in the nav and the sitemap (5 URLs).

## Note

The guide states "since v3.5.1" for the `[perf]` warning, so it wants #602 to land first. Merging this to `develop` before then is harmless (the doc is ahead of the branch it describes for a few hours), but sequencing after is tidier.

### DIFF
--- a/README.md
+++ b/README.md
@@ -312,6 +312,8 @@ System clock 26.590906 MHz NTSC / 26.593900 MHz PAL. Full source map: [`docs/sou
 
 **Known limitations:** the blitter is not fully cycle-accurate (a few games need fast mode) and GPU/DSP pipeline hazards are not yet modelled ([#313](https://github.com/libretro/virtualjaguar-libretro/issues/313)), which shows up as some titles running fast.
 
+**Running slow?** The defaults are tuned for compatibility, not speed, and the biggest available speedup ships off by default. See the [settings & performance tuning guide](docs/settings-and-performance-guide.md) — it covers which options actually matter, and which combinations quietly cancel each other out.
+
 ---
 
 ## Documentation and support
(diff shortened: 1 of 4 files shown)


JoeMatt:
Thanks — three of four were real and are fixed in the push above. One rests on a wrong premise, detailed below.

**1. `bios` vs `bios_type` — not a contradiction; they are two different options.** Verified against `libretro_core_options.h`:

- `virtualjaguar_bios` — *"BIOS (Cartridges)"*, HLE vs Real
- `virtualjaguar_bios_type` — *"Cart BIOS Type (Restart)"*, which boot ROM (Series K / Model M / Custom) once the above is Real

Both rows were already correct: the restart-required table lists `bios_type` because that is the one needing a restart, and the quick-reference lists `bios` because that is the one you actually toggle. Nobody pasting either key gets a non-existent option. **That said, the confusion is the reader's too** — if a reviewer reading closely merged them, a user skimming will. Both rows now say what distinguishes them.

**2. Field/sample rates — figures correct, citation weak. Fixed.** They are not folklore: `src/core/jaguar.h` defines `VJ_HALFLINES_PER_FIELD_NTSC/PAL` as 524/624 and derives the rate as `1e6 / (halflines × halfline_µs)` = 60.05445 / 50.08013, and `retro_get_system_av_info()` derives sample rate from it because the core submits a fixed batch of pairs exactly once per field. But you are right that citing `frontend-pacing.md` was the same weak sourcing this PR rejected elsewhere — the guide now cites the code and explains the derivation, including why a flat 48000 would over-deliver by ~0.09% and drift the frontend's audio buffer.

**3. §4.1 self-contradiction — real, fixed.** The table bolded 50–67% as the DSP range and then cited Skyhammer at ~33% two rows down. Now 33–67% overall with 50–67% scoped to the DSP-heavy subset. The site page carried the same table and got the same fix.

**4. `/blob/master/` link — real bug, mine, fixed.** Every other site page uses `/blob/develop/`; the new one did not. `master` does exist here (release-only, tagged commits), so it would not have 404'd — but it would have pointed at a tag-frozen copy that goes stale between releases, which is worse in a subtler way. Now consistent with the rest of the site.

On the note about the idle-skip gate: correct that you could not confirm it from the diff. It is `src/jerry/dsp.c`, `DSPExec()` — `riscClockScalePct == 100 && !busArbiter.enabled && !vjs.gpuPipelineTiming && !blitMemoMode && !blitMemoRecording && !(dsp_control & 0x18)`. Two agents were told to verify against that gate rather than the issue text, and both independently found the issue text wrong about `m68k_clock_scale` and `blitter_timing`; #595 has been corrected.

Issue #463, opened 2026-08-15
Cybermorph: Codex-level crash + erratic ship movement under overclock — not reproducible headlessly, needs artifacts


JoeMatt:
User report (via JoeMatt), on the **Codex** level of Cybermorph:

> On the latest develop, Cybermorph still has that same crash, with the same requirements to reproduce it. Now, though, it seems like the game is completely broken, with ship movement and altitude gone nuts. Can't really make a move without randomly crashing into a suddenly-appearing wall.

They also say the crash **predates this release** ("has for the last version too and maybe older releases"). Those are two claims and they should be tracked as two: a long-standing level-specific crash, and a newer movement problem.

## Artifact in hand

A savestate captured **just before** the crash. It loads clean on `6e3d2b3` and lands on `CODEX BRIEFING — "find the pods, grab them and get out quick."` Kept in the private tree (savestates are never committed — see `test/fixtures/README.md`).

- ROM: `Cybermorph (1993).jag`, **md5 `1c2e3d172151dcae3b3fc9a23084dcf2`**, 2 MB World rev. The 1 MB `Cybermorph_(1994).jag` renders **black** from this state — wrong ROM, worth knowing before anyone debugs a phantom.
- Savestate version **11** (also what v3.3.0 writes).
- Briefing is dismissed with **A, B or C** (`pause`/`option`/keypad do nothing). A press before ~field 60 is ignored.

## Not reproduced headlessly — what is already excluded

All from that state, scripted digital input, on `6e3d2b3`:

| hypothesis | test | result |
|---|---|---|
| Regression since v3.3.0 | `6e3d2b3` vs `385c4d4`, identical input, 2400 fields | **byte-identical fb hashes, every field** |
| `dram_timing` / `blitter_timing` / all three timing models | vs default | **identical** |
| `gpu_pipeline_timing` (the suspected user config) | vs default, 14 400 fields | **identical output** — though the knob is **live**: +8% instructions retired (233.5 G -> 252.3 G), so it changes cycle accounting, not what the game computes |
| Emulator-level fault while flying | 14 400 fields (~4 min emulated) varied flight, `crash_detect=verbose`, both timing configs | **no `gpu_pc_escape`, no wedge, no stall, not frozen** (256 distinct hashes in the last 600 fields) |

So at **stock clocks** the emulation path is unchanged between v3.3.0 and develop, and none of the timing models alter this title's behaviour.

[...]

PR #610, opened 2026-08-26
feat(options): split Timing into Speed and Hardware Timing; overclock gate-1 tooling

JoeMatt:
Two things, both from trying to answer "which games can overclocking actually speed up".

## The options were confusing because one category held three intents

`Timing` contained, in one flat list:

- **two overclocks** (`m68k_clock_scale`, `risc_clock_scale`) that change *what the game computes* and can break it;
- **two fast-forwards** (`risc_idle_skip`, `blit_memo`) that are bit-exact — identical output, less host work, and by construction they *cannot* change the frame rate;
- **three accuracy models** (`dram_timing`, `gpu_pipeline_timing`, `blitter_timing`) that deliberately make things **slower**.

A user had no way to tell "makes it faster" from "makes it slower on purpose".

Split into **Speed** and **Hardware Timing (Experimental)**, with Speed ordered by what to try first: the two free fast-forwards, then the two overclocks. `blit_memo` moves in from **Video** — same free-speedup class, and it silently disables idle-skip, which was undiscoverable from another category.

**Option keys are unchanged**, so no user config breaks. `retro_api_version` untouched.

`risc_clock_scale` now leads with the trap instead of burying it: turning it up switches idle-skip off, so on a DSP-bound title it makes you *slower*.

## One text fix that was actively wrong

`m68k_clock_scale` cited **AvP and Checkered Flag** as games overclocking can smooth. Both were measured under #378 and both failed — AvP is locked to one frame per 5 fields, Checkered Flag caps itself in software at ROM `$8026EE`. Replaced with the general rule and the two counterexamples.

## Tooling + the first title to clear gate 1

- `scripts/ocgrid.sh` — the {1x,1.5x,2x} RISC x {1x,1.5x} 68K grid, with saturation detection.
- `scripts/rzip_extract.py` — RetroArch writes rzip-compressed states when compression is on, and `retro_unserialize` rejects them on size. Raw files pass through unchanged.
- `docs/overclock-gate1-checklist.md` — traps, savestate recipe, per-title checklist.

**Cybermorph clears gate 1** — the first title to do so. 2400-frame window:

| cell | transitions | vs stock |
|---|---:|---:|
| RISC 0.5x (control) | 697/2400 | **-48.7%** |
| RISC 1x, 68K 1x (stock) | 1359/2400 | — |
| RISC 1.5x | 1967/2400 | **+44.7%** |
| RISC 2x | 2349/2400 | **+72.8%** |
| RISC 1x, 68K 1.5x | 1362/2400 | +0.2% |
| RISC 1.5x, 68K 1.5x | 1994/2400 | +46.7% |
| RISC 2x, 68K 1.5x | 2311/2400 | +70.1% |

All seven cells come from one 2400-frame window; nothing is spliced across window lengths.

The 0.5x arm is the control that makes the rest trustworthy: it proves the metric was live. Iron Soldier reads 115 at 0.5x against 116 at 1x — a genuine null, not a dead measurement, and consistent with it being the extreme DSP-idle case (needs idle-skip, not overclock).

Not proposing a DB preset: Cybermorph is also the subject of the open overclock crash report #463, and gate 2 (play-test) still applies.

`make TEST_EXPORTS=1 test` exits 0, C89 lint clean.

Refs #378

### DIFF
--- a/docs/overclock-gate1-checklist.md
+++ b/docs/overclock-gate1-checklist.md
@@ -82,19 +82,28 @@ ROMs for all eight candidates are present in `jaguar-roms-private/ROMS/`.
 | Iron Soldier | `iron_soldier_v104_f2400.state` | Null — clock-insensitive, confirmed by underclock control |
 | Val D'Isere | `... (1994).state.auto` | **UNMEASURED** — near-static scene, needs a real gameplay state |
 
-**Cybermorph is the first title to clear gate 1.** 2400-frame window (ceiling 2399):
+**Cybermorph is the first title to clear gate 1.** All seven cells, one 2400-frame window
+(saturation ceiling 2399):
 
-| RISC scale | transitions | ≈ fps | vs stock |
+| cell | transitions | ≈ fps | vs stock |
 |---|---:|---:|---:|
-| 0.5x | 258/900 | — | −49.7% |
-| **1x (stock)** | **1359/2399** | **~34** | — |
-| 1.5x | 1967/2399 | ~49 | **+44.7%** |
-| 2x | 2349/2399 | ~59 | **+72.8%** |
+| RISC 0.5x (control) | 697/2400 | ~17 | **−48.7%** |
+| **RISC 1x, 68K 1x (stock)** | **1359/2400** | **~34** | — |
+| RISC 1.5x | 1967/2400 | ~49 | **+44.7%** |
+| RISC 2x | 2349/2400 | ~59 | **+72.8%** |
+| RISC 1x, 68K 1.5x | 1362/2400 | ~34 | +0.2% |
+| RISC 1.5x, 68K 1.5x | 1994/2400 | ~50 | +46.7% |
+| RISC 2x, 68K 1.5x | 2311/2400 | ~58 | +70.1% |
 
-The 68K axis is inert here (+0.6% at 1.5x): this is a RISC-bound title, so only
-`risc_clock_scale` matters.
+The 68K axis is inert (+0.2% on its own): this is a RISC-bound title, so only
+`risc_clock_scale` matters, and stacking the 68K scale on top adds nothing.
 
-At 2x, 2349 of 2399 fields carry a new frame — Cybermorph is now presenting at essentially
+**The control is what makes the rest trustworthy.** RISC 0.5x costs 48.7% — near-linear
+against the +44.7% that 1.5x gains — so the metric tracks clock speed in both directions on
+this title. Every figure above comes from the same 2400-frame window; nothing is spliced
+across window lengths.
+
+At 2x, 2349 of 2400 fields carry a new frame — Cybermorph is now presenting at essentially
 the display rate. That is a real ceiling rather than the trap-2 counter artefact, but it does
 mean **+72.8% is a floor on the 2x gain, not the exact figure**; the engine may have headroom
 the display cannot show.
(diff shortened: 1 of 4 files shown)


JoeMatt:
All four findings were real and are fixed in `ac7a8e0`. The intl question resolves to "ignore", with evidence.

**1. `ocgrid.sh` discarded its output argument — fixed.** `OUT=$1; shift` consumed the filename and nothing ever referenced it, so anyone following §4 got stdout and a silently dropped file. It now writes `$OUT` *and* echoes, via a `tee`-based `emit`.

**2. The table spliced two window lengths — fixed by re-measuring, not by re-labelling.** You were right about the arithmetic and right about the cause: a 900-frame control run had been folded into a table headed "2400-frame window", quoting a 1x baseline of 513 that appears nowhere. Rather than add the explanatory sentence, all seven cells were re-run in one 2400-frame window:

| cell | transitions | vs stock |
|---|---:|---:|
| RISC 0.5x (control) | 697/2400 | **−48.7%** |
| RISC 1x, 68K 1x (stock) | 1359/2400 | — |
| RISC 1.5x | 1967/2400 | **+44.7%** |
| RISC 2x | 2349/2400 | **+72.8%** |
| RISC 1x, 68K 1.5x | 1362/2400 | +0.2% |
| RISC 1.5x, 68K 1.5x | 1994/2400 | +46.7% |
| RISC 2x, 68K 1.5x | 2311/2400 | +70.1% |

The conclusion is unchanged and now better supported: −48.7% at half clock against +44.7% at 1.5x is near-linear in both directions. Every number comes from one window.

**3. Bare `rm -rf` — fixed.** `trap ... EXIT INT TERM` + `command rm -rf`, so it also stops leaking the mktemp dir when a run fails or is interrupted.

**4. Hardcoded `.dylib` — fixed.** `CORE=${CORE:-./virtualjaguar_libretro.dylib}`, documented in §4.

**Your question about intl arrays.** Checked: `libretro_core_options_intl.h` is a 91-line stub — no `option_cats_*`, no `option_defs_*`, and zero references to `"timing"` or any other category key. Nothing renders uncategorized in any locale. Good question to ask, though; if translations had existed this would have been a silent per-locale regression that no test here would catch.

**One change beyond the findings.** Finding 2 was only possible because the control arm was run separately and by hand. The 0.5x control is now a fixed cell in the grid rather than something the operator remembers — a flat 1x/1.5x/2x row cannot distinguish a genuine frame cap from a measurement that never responded, and that ambiguity is the direct cause of both retracted result tables on #378. Reporting is two-phase so the stock cell is known before anything is quoted against it.

JoeMatt:
Both findings fixed in `2b180cb`, and one of them was validated by real data in the same run.

**1. `trap ... EXIT INT TERM` never stopped the sweep — fixed.** Exactly as described: the handler cleaned up and execution continued, so Ctrl-C killed one cell (recorded FAILED) and six more multi-minute cells followed. Now `trap ... EXIT` for cleanup plus `trap "exit 130" INT TERM` to actually stop, so the temp dir is still removed.

**2. A saturated stock cell could become the baseline — fixed.** Only a measured stock cell sets `base` now; otherwise every row reports `NO USABLE STOCK BASELINE` instead of falling through to a percentage against a ceiling-pinned number.

Worth noting this was not hypothetical. **Super Burnout hit it in the very run that prompted the fix** — 1799/1800 in all seven cells, stock included. Every other cell happened to saturate too, so nothing bogus printed, but a single non-saturated arm would have shown an authoritative `+x.y% vs stock` against a meaningless baseline.

**3. PR description table — fixed.** Replaced with the shipped 2400-frame table (697/2400 control). You were right that it undercut the doc's "one window, nothing spliced" claim.

**4. intl tables — checked, nothing stale.** `libretro_core_options_intl.h` is a 91-line stub: no `option_cats_*`, no `option_defs_*`, and zero references to `"timing"` or any category key. No locale regresses.

---

Gate 1 is now **complete across all eight candidates**, which is also in the doc:

| Title | control 0.5x | 1.5x | 2x | verdict |
|---|---:|---:|---:|---|
| **Cybermorph** | −48.7% | **+44.7%** | **+72.8%** | **PASSES** |
| I-War | −18.2% | +6.3% | +1.1% | noise, capped at stock |
| Missile Command 3D | −0.2% | **−14.3%** | −11.2% | overclock actively harmful |
| Doom | ±0.0% | ±0.0% | ±0.0% | flat lock |
| Club Drive | ±0.0% | ±0.0% | ±0.0% | flat lock |
| Iron Soldier | −0.9% | ±0.0% | ±0.0% | flat lock |
| Super Burnout | SATURATED | — | — | already at the display ceiling |
| Val D'Isere | ±0.0% | ±0.0% | ±0.0% | unmeasured, static savestate |

One pass in eight. The structural-null hypothesis holds as a general rule but not as a law — Cybermorph disproves the strong form.

Two incidental findings recorded in the doc rather than dropped: Missile Command 3D renders **less** when given more cycles (timing-sensitive path, worth its own look), and Doom reads 593/1800 = **19.8 fps in gameplay** against real hardware's ~20 renders/s — a lead for #401, since the 2x-too-fast figure came from the attract demo, a different code path.

PR #613, opened 2026-08-26
perf(blitter): opt-in arch autodetect so SPM builds stop inlining scalar

### LINKED ISSUE #612: [perf][build] SPM/Xcode builds inline the SCALAR blitter on arm64 — NEON is never selected
## Summary

Same silent-scalar class as #560, different build system.

`src/tom/blitter_simd.h` selects the inline implementation from
`-DBLITTER_SIMD_<ARCH>`, which `Makefile.common` passes alongside the matching
`blitter_simd_<arch>.c`. Any build system that selects the `.c` by *arch guard*
instead of by Makefile selection cannot pass that `-D`, and falls through to:

```c
#else
#  include "blitter_simd_scalar.h"
#endif
```

The Provenance frontend's Swift Package Manager build
(`Provenance-Emu/Core-VirtualJaguar`, `Package.swift`) is exactly that: it hands
every `blitter_simd_<arch>.c` to the compiler at once and lets each file's
top-of-file arch guard blank out the ones that don't apply.

## Impact

On arm64 (every iOS/tvOS/visionOS device and Apple-silicon Mac), the SPM build:

- inlines the **scalar** ops into `BlitterMidsummer2` (the hot path), while
- linking a **NEON** `blitter_simd_ops` vtable.

The vtable is only consumed by `test/test_blitter_simd.c`, so the test suite has
been validating NEON while the shipped binary ran scalar. Nothing caught it.

Per `blitter_simd.h`'s own header comment, reaching these ops out-of-line cost
~21% of accurate-blitter runtime; the inline path is the whole point of the
`blitter_simd_<arch>.h` split.

## Reproduction

Preprocessor probe against `blitter_simd.h`:

```c
#include "blitter_simd.h"
#if defined(BLITTER_SIMD_NEON)
#error "SELECTED=NEON"
#elif defined(BLITTER_SIMD_SSE2)
#error "SELECTED=SSE2"
#else
#error "SELECTED=SCALAR"
#endif
```

```
cc -fsyntax-only -I./src -I./src/core -I./src/tom -I./libretro-common/include probe.c
  -> SELECTED=SCALAR        # what SPM gets today, on arm64
cc -fsyntax-only -DBLITTER_SIMD_NEON ... probe.c
  -> SELECTED=NEON          # what the Makefile gets
```

## Fix

Opt-in `BLITTER_SIMD_AUTODETECT` in `blitter_simd.h`: when no explicit
`BLITTER_SIMD_*` is defined *and* the consumer asks for detection, derive the
arch from the compiler's own predefined macros.

Detection belongs in the header, not in `Package.swift`: the SPM manifest is
compiled for the **host**, so a manifest-side check picks arm64 for the x86_64
simulator slice. The header is evaluated by the real target compiler.

Deliberately opt-in so ndk-build and every MSVC target — which pass no `-D` and
select scalar on purpose — stay byte-identical.

JoeMatt:
Fixes the silent-scalar blitter in SPM/Xcode builds reported in #612.

## What

`blitter_simd.h` gains an **opt-in** `BLITTER_SIMD_AUTODETECT` gate. When the
consumer defines it and no explicit `BLITTER_SIMD_*` is set, the arch is derived
from the compiler's own predefined macros.

## Why the header and not `Package.swift`

The SPM manifest is compiled for the **host**. A manifest-side check would pick
arm64 for the x86_64 simulator slice. The header is evaluated by the real target
compiler, so it gets every slice right.

## Why opt-in

`ndk-build` and every MSVC target pass no `-D` and select scalar deliberately —
the existing header comment says so. Gating on `BLITTER_SIMD_AUTODETECT` leaves
them byte-identical. Every Makefile target is unaffected because Makefile.common
already passes an explicit `-D`.

## Verification

Preprocessor probe (see #612 for the probe source), this tree:

| flags | selected |
|---|---|
| *(bare, no `-D`)* — MSVC / ndk-build path | `SCALAR` (unchanged) |
| `-DBLITTER_SIMD_AUTODETECT`, arm64 | `NEON` |
| `-DBLITTER_SIMD_AUTODETECT -target x86_64-apple-macos14` | `SSE2` |
| `-DBLITTER_SIMD_SCALAR -DBLITTER_SIMD_AUTODETECT` | `SCALAR` (explicit wins) |
| `-DBLITTER_SIMD_NEON` — Makefile path | `NEON` (unchanged) |

Row 3 is the one that proves the fix is at the right layer.

```
make -j10                    # clean
make TEST_EXPORTS=1 test     # OK: 0 test(s) failed.
```

Header-only change; `scripts/c89-lint.sh` has no `.c` to check here, and the
sibling `blitter_simd_*.c` files are untouched.

## Follow-up (separate repo, not this PR)

`Provenance-Emu/Core-VirtualJaguar`'s `Package.swift` needs
`.define("BLITTER_SIMD_AUTODETECT", to: "1")` on the `libjaguar` target to
actually consume this.

### DIFF
--- a/src/tom/blitter_simd.h
+++ b/src/tom/blitter_simd.h
@@ -89,7 +89,35 @@ extern const blitter_simd_ops_t blitter_simd_ops;
  * added to SOURCES_C.  Builds that don't go through it (ndk-build, and
  * every MSVC target) get scalar, which is what those already selected.
  * A mismatch between the -D and the compiled .c is a hard compile error
- * rather than a silent slow path -- see blitter_simd_<arch>.c. */
+ * rather than a silent slow path -- see blitter_simd_<arch>.c.
+ *
+ * BLITTER_SIMD_AUTODETECT is the opt-in for build systems that compile
+ * the arch .c files by guard rather than by selection, and so cannot
+ * pass a matching -D.  The Swift Package Manager manifest used by the
+ * Provenance frontend is the case this exists for: it hands every
+ * blitter_simd_<arch>.c to the compiler at once and lets each file's
+ * own top-of-file arch guard blank out the ones that don't apply.
+ * Without a -D those builds fell through to the scalar branch below and
+ * inlined the scalar ops into BlitterMidsummer2 on hardware that has
+ * NEON -- the vtable was NEON, but the vtable is only used by
+ * test/test_blitter_simd.c, so nothing caught it.  That is a silent
+ * slow path of exactly the kind the paragraph above says cannot happen,
+ * and it is why the detection is spelled here, where the real target
+ * compiler evaluates it, rather than in the manifest, which is compiled
+ * for the host and would get the x86_64 simulator slice wrong.
+ *
+ * Deliberately opt-in: leaving it undefined keeps ndk-build and every
+ * MSVC target on the scalar path they select today, byte for byte. */
+#if !defined(BLITTER_SIMD_NEON) && !defined(BLITTER_SIMD_SSE2) \
+ && !defined(BLITTER_SIMD_SCALAR) && defined(BLITTER_SIMD_AUTODETECT)
+#  if defined(__ARM_NEON) || defined(__ARM_NEON__) || defined(__aarch64__)
+#    define BLITTER_SIMD_NEON 1
+#  elif defined(__x86_64__) || defined(__i386__) \
+     || defined(_M_X64) || defined(_M_IX86)
+#    define BLITTER_SIMD_SSE2 1
+#  endif
+#endif
+
 #if defined(BLITTER_SIMD_NEON)
 #  include "blitter_simd_neon.h"
 #elif defined(BLITTER_SIMD_SSE2)
--- a/src/tom/blitter_simd.h
+++ b/src/tom/blitter_simd.h
@@ -110,10 +110,24 @@ extern const blitter_simd_ops_t blitter_simd_ops;
  * MSVC target on the scalar path they select today, byte for byte. */
 #if !defined(BLITTER_SIMD_NEON) && !defined(BLITTER_SIMD_SSE2) \
  && !defined(BLITTER_SIMD_SCALAR) && defined(BLITTER_SIMD_AUTODETECT)
-#  if defined(__ARM_NEON) || defined(__ARM_NEON__) || defined(__aarch64__)
+   /* NEON: mandatory on AArch64, opt-in on ARMv7-A.  __ARM_NEON is the
+    * ACLE spelling every GCC/Clang emits when NEON codegen is on;
+    * _M_ARM64 is MSVC's, where NEON is likewise mandatory. */
+#  if defined(__ARM_NEON) || defined(__ARM_NEON__) || defined(__aarch64__) \
+   || defined(_M_ARM64) || defined(_M_ARM64EC)
 #    define BLITTER_SIMD_NEON 1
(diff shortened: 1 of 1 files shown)


JoeMatt:
Thanks — the 32-bit x86 finding is correct and is now fixed in 596f02f. Taking the three points in turn.

### 1. SSE2 inferred from `__i386__`/`_M_IX86` — real, fixed

The asymmetry that makes this a genuine defect rather than a theoretical one: `Makefile.common` can **add** the capability, autodetect can only **observe** it. Its SSE2 rule appends the flag explicitly —

> `# Add -msse2 flag for all GCC/Clang SSE2 builds. No-op on x86_64 (baseline), required on i686 / gcc -m32.`

— so on the Makefile path `-msse2` is a consequence of the selection. In the header there is no such lever, so inferring SSE2 from the arch hands `blitter_simd_sse2.c`'s intrinsics to targets that cannot compile them. Exactly the class the surrounding comment claims cannot happen.

Now gated on `__SSE2__` / `_M_IX86_FP == 2` for the 32-bit arms; the x86-64 spellings keep no extra test because SSE2 really is baseline there. A plain i386 build falls through to scalar, which is the safe answer.

### 2. `_M_ARM64` — folded in

Agreed on the asymmetry. Added `_M_ARM64` and `_M_ARM64EC`.

### 3. Guard-vs-include ordering — verified, and you're right that this diff can't show it

Good catch that it's the one link not evidenced here. The reason is that the arch guards don't exist on `develop` at all — the Makefile selects exactly one `.c`, so no guard is needed. They live on the consumer branch (`Provenance-Emu/virtualjaguar-libretro@provenance-spm-fixes`), where the ordering is:

```
16: #if defined(__ARM_NEON) || defined(__ARM_NEON__) || defined(__aarch64__)
24: #ifndef BLITTER_SIMD_NEON
25: #define BLITTER_SIMD_NEON 1
26: #endif
28: #include "blitter_simd.h"
```

The guard tests the **compiler's** arch macro, not `BLITTER_SIMD_NEON`, so the inversion you describe can't occur. Confirmed empirically rather than by reading: `swift build --target libjaguar` links clean, and `blitter.c.o` (arm64, `-O3`) goes from **2 to 24** NEON vector instructions with the define on — so `blitter_simd_ops` resolved *and* the hot path actually moved.

### Selection matrix

Impl headers stubbed via their include guards so only the selection is under test:

| flags | selected |
|---|---|
| *(bare, no `-D`)* | `SCALAR` (unchanged) |
| `AUTODETECT` arm64 | `NEON` |
| `AUTODETECT` x86_64 | `SSE2` |
| `AUTODETECT` armv7 `-mfpu=neon` | `NEON` |
| `AUTODETECT` armv7 `-mfpu=vfpv3-d16` | `SCALAR` |
| `AUTODETECT` i386 `-mno-sse2` | `SCALAR` ← **was `SSE2`, the finding** |
| `AUTODETECT` i386 `-msse2` | `SSE2` |
| `AUTODETECT` i386 `-march=pentium4` | `SSE2` |
| `BLITTER_SIMD_SCALAR` + `AUTODETECT` | `SCALAR` (explicit wins) |
| `BLITTER_SIMD_NEON` (Makefile) | `NEON` (unchanged) |

`make TEST_EXPORTS=1 test`: 0 test(s) failed.

The broader question of 32-bit x86 SSE2 handling elsewhere in the tree is filed separately rather than widened into this PR.

Issue #617, opened 2026-08-26
[build] Arch files can't self-select, so an all-variants build duplicates or loses blitter_simd_ops


JoeMatt:
## Problem

`blitter_simd_{neon,sse2,scalar}.c` each define `blitter_simd_ops` unconditionally.
That is correct for `Makefile.common` and `ndk-build`, which compile exactly **one** of
them. It does not work for a build system with no per-target source selection, which
must hand the compiler all three: it gets three definitions of the same symbol.

The workaround downstream (`Provenance-Emu/Core-VirtualJaguar`) is to add a
top-of-file arch guard to `neon.c`/`sse2.c` on a fork branch and `exclude:` the scalar
file in `Package.swift`. Three problems with that:

1. It lives on a fork branch that has to be rebased forever (#614).
2. Its guards test the **architecture**, so a 32-bit x86 target without `-msse2` matches
   the SSE2 guard and compiles intrinsics it cannot build — the same defect found in
   review on #613.
3. Excluding scalar means a target that is neither ARM nor x86 defines `blitter_simd_ops`
   **nowhere** and fails to link, rather than falling back.

There is also a subtler hazard: the guard and #613's `BLITTER_SIMD_AUTODETECT` block are
two separate copies of the same condition. If they ever drift, the build either defines
the vtable twice or inlines one implementation while linking another — which is exactly
the silent-mismatch class #612 was about.

## Proposal

Add `src/tom/blitter_simd_arch.h` as the single source of truth for "what does this
compilation actually have", and have both consumers read it:

- `blitter_simd.h`'s autodetect block — picks the **inline** set `blitter.c` gets.
- each arch `.c`'s guard — picks which file contributes the **vtable**.

The `.c` files cannot get this from `blitter_simd.h` itself: they must select their
implementation *before* including it, since it dispatches on `BLITTER_SIMD_<ARCH>` and
including it first pulls in the scalar inline set, which then collides with the arch one.

Guard scalar as well, so under autodetect exactly one of the three compiles on **every**
target and the set is total — no hand-maintained `exclude:` downstream, no link failure
on an exotic target.

Gate the whole thing on `BLITTER_SIMD_AUTODETECT` so `BLITTER_SIMD_BUILD_*` is
unconditionally true for Makefile and ndk-build. Their chosen file can never be blanked
out.

## Context

Second of the four SPM-compat fixes tracked in #614. Siblings: #612/PR #613,
#615/PR #616.

PR #625, opened 2026-08-26
build: ship a Package.swift so Apple frontends stop vendoring this repo

### LINKED ISSUE #614: [build] Ship a Package.swift in this repo so Apple consumers stop needing a fork
## Problem

Apple/SwiftPM consumption of this core currently needs a **fork branch that must be
rebased forever**: `Provenance-Emu/virtualjaguar-libretro@provenance-spm-fixes`. The
Provenance wrapper (`Provenance-Emu/Core-VirtualJaguar`) vendors that branch as a
submodule and re-declares every upstream C source in its own `Package.swift`.

Two failure modes fall directly out of that topology, both already observed:

1. **Source-list drift.** The wrapper hand-mirrors `Makefile.common`. It drifted by
   27 files over seven releases before the last reconciliation.
2. **Build-flag drift.** #612 — SPM passed no `-DBLITTER_SIMD_<ARCH>`, so `blitter.c`
   inlined the **scalar** blitter on every arm64 device while the linked vtable was
   NEON. The vtable is only consumed by `test/test_blitter_simd.c`, so this repo's own
   test suite validated NEON while the shipped binary ran scalar. Nothing upstream
   could have caught it, because nothing upstream describes the SPM build.

The fork branch also strands consumers on old trees — it was pinned to a v3.5.1-era
tree while `develop` had moved on.

## Proposal

Ship a **minimal `Package.swift` in this repo** exposing only the C library targets
(`libjaguar`, `libretro-common`, `libchdr`). Frontend-specific Swift/ObjC targets stay
in the consumer, which then depends on this package by URL instead of vendoring a
submodule.

That makes the source list reviewable *next to the sources it mirrors*, and lets CI
assert it matches `Makefile.common` — turning both drift classes into a red build here
rather than a silent regression downstream.

## Prerequisite: land the SPM-compat source fixes first

Most of the benefit does not require the manifest at all. Four fixes currently live
only on the fork branch, and all four look upstreamable:

- [x] `BLITTER_SIMD_AUTODETECT` gate in `blitter_simd.h` — #612 / PR #613
- [ ] Top-of-file arch guards in `blitter_simd_{neon,sse2}.c`, so a build that compiles
      every variant blanks out the inapplicable ones. Inert for Makefile builds, which
      compile exactly one.
- [ ] `vjag_memory.h` `DEBUG` macro push/pop — Xcode passes `-DDEBUG=1` for Debug
      configurations, which expands the `DEBUG` **enumerator** in
      `enum { UNKNOWN, ..., DEBUG }` to `1` and breaks the parse. Affects any Xcode
      consumer, not just SPM.
- [ ] `src/core/version.h` — gitignored and generated by `scripts/gen-version-h.sh`,
      which SPM has no hook to run. Needs a committed fallback or a
      tolerate-missing-header path in `libretro.c`.

**Once those land, the consumer's submodule can track plain `develop` and the fork
branch dies.** That is the bulk of the win and needs no decision about a Swift manifest.
The manifest is the follow-up that closes the source-list gap.

## Known technical caveat

`libjaguar` needs `path: "."` (because `libretro.c` sits at the repo root), which
overlaps `libretro-common/` and `deps/libchdr/`. SPM tolerates this when each target
lists its `sources:` explicitly — the Provenance wrapper already does exactly this
overlap against the submodule root and builds clean, so it is proven rather than
theoretical.

Non-Apple builds are unaffected: a root `Package.swift` is inert to every Makefile
target. The main cost is repo-level (GitHub will label the repo Swift, and SwiftPM will
treat the repo as a package), so this is worth an explicit maintainer OK before
implementing.

JoeMatt:
Closes the last item on #614.

> **Stacked.** Based on `fix/duplicate-boolean-h` (#624), which is based on
> `fix/drop-dangling-vfs-cdrom-header` (#622). Merge **#622 → #624 → this**.
>
> **The `src/core/boolean.h` deletion visible in this diff is #624's, not this change's.**
> GitHub shows the diff against `develop`, so the parent's commit rides along in the view.
> This PR adds only `Package.swift`, `scripts/check-package-sources.sh`, and two CI jobs.

## Why

Apple consumption of this core needs a fork branch plus a downstream manifest that
re-declares every source file. **That second copy of the list is the problem.** The
Provenance wrapper kept one and drifted by 27 files over seven releases, and a missing
`-DBLITTER_SIMD_<ARCH>` in it meant every arm64 build inlined the **scalar** blitter while
the linked vtable was NEON (#612) — invisible because only the core's own tests read that
vtable.

This manifest builds the C core only. A frontend depends on it as a package instead of
vendoring it, and the list lives next to the sources it mirrors.

Inert everywhere else: `make`, `jni/Android.mk` and the MSVC projects never read it, and
nothing here changes a flag they use.

## Three guards, because a hand-copied list rots

**`scripts/check-package-sources.sh`** diffs the manifest's arrays against
`Makefile.common`, which stays the single source of truth. Fails loudly and names the
files; never auto-fixes. Verified it catches **both** directions:

```
$ sh scripts/check-package-sources.sh          # a file dropped from Package.swift
ERROR: in Makefile.common but NOT in Package.swift (coreSources):
    src/tom/blit_memo.c

$ sh scripts/check-package-sources.sh          # a stale entry Makefile.common dropped
ERROR: in Package.swift but NOT in Makefile.common (coreSources):
    src/tom/ghost_file.c
```

**`swift build` in CI.** Nothing else in CI reads this file, so a manifest that does not
parse would merge unnoticed. That job also asserts **exactly one** `blitter_simd_ops`
survives when all three SIMD variants compile together — the invariant #617's arch guards
exist to hold, which no Makefile target can check because it compiles only one variant.

**A versioned consumer builds against it.** Added after review: SwiftPM refuses to build a
versioned dependency on a package whose targets carry `unsafeFlags`, and building in-repo
cannot detect that because the **root** package is exempt. This job makes a throwaway
tagged clone and a tiny consumer that depends on it by version, which is the only way to
see the restriction. It must use `swift build`, not `swift package resolve` — resolve
succeeds even with unsafe flags present.

## Details worth a look

- **Target names are prefixed** (`virtualjaguar-libretro-common`, `virtualjaguar-libchdr`).
  SPM requires uniqueness across the consuming app's *whole* package graph, and generic
  names collide — a frontend carrying two cores that each vendor CHD cannot build if both
  call the target `libchdr`. Not hypothetical; it broke the Provenance build this week.
- **`path: "."`** because `libretro.c` sits at the repo root. The explicit `sources` list
  means only named files compile, so `docs/`, `test/`, `scripts/` are ignored; the two
  subdirectories owned by other targets are excluded so SPM does not reject overlapping
  paths.
- **`HAVE_COCOATOUCH` / `__GCCUNIX__` are deliberately absent** — the downstream manifest
  defines both, but neither appears anywhere in the core sources.

## Cost

A root `Package.swift` makes GitHub label the repo "Swift" and makes SwiftPM treat it as a
package. That is the trade-off flagged in #614 and worth an explicit maintainer OK.

## Verification

| check | result |
|---|---|
| `swift build` (root) | clean, 80 files |
| `swift build` from a **versioned** consumer | links; fails as expected when `unsafeFlags` is reintroduced |
| `scripts/check-package-sources.sh` | 66 core + 13 libretro-common agree |
| `make -j` | 0 errors (unaffected) |
| `make TEST_EXPORTS=1 test` | 0 test(s) failed |
| SIMD vtables in the SPM objects (`nm`) | neon 1, sse2 0, scalar 0 |
| blitter hot path | 47 NEON vector instructions |

### DIFF
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -503,6 +503,47 @@ jobs:
     - name: swift build
       run: swift build -v 2>&1 | tail -40
 
+    # The blind spot that let an `unsafeFlags` slip in: SwiftPM REFUSES to
+    # resolve a VERSIONED dependency on a package whose targets carry unsafe
+    # flags, but a package built as the ROOT (which `swift build` above does)
+    # is exempt -- so the restriction is invisible from inside the repo. This
+    # consumes the package the way a frontend actually would, by version tag,
+    # which is the only way to see it.
+    - name: a versioned consumer can build against this package
+      run: |
+        set -euo pipefail
+        work=$(mktemp -d)
+        # A tag on a throwaway clone; the real repo is untouched.
+        # Clone under the repo name: SwiftPM derives package identity from the
+        # URL's last path component, not from `name:` in the manifest.
+        git clone --quiet --no-hardlinks . "$work/virtualjaguar-libretro"
+        git -C "$work/virtualjaguar-libretro" -c user.email=ci@example.com -c user.name=CI tag -f 0.0.1
+        mkdir -p "$work/app/Sources/app"
+        echo 'print("ok")' > "$work/app/Sources/app/main.swift"
+        cat > "$work/app/Package.swift" <<PKG
+        // swift-tools-version:5.9
+        import PackageDescription
+        let package = Package(
+            name: "app",
+            dependencies: [.package(url: "$work/virtualjaguar-libretro", from: "0.0.1")],
+            targets: [.executableTarget(name: "app", dependencies: [
+                .product(name: "VirtualJaguar", package: "virtualjaguar-libretro")])]
+        )
+        PKG
+        cd "$work/app"
+        # `swift build`, NOT `swift package resolve`: resolve succeeds even with
+        # unsafe flags present -- the restriction is enforced when the dependency
+        # graph is loaded for a build. Verified both ways before relying on it.
+        if ! out=$(swift build 2>&1); then
+          echo "$out"
+          echo "::error::a versioned dependency on this package does not resolve."
+          echo "::error::If this says 'contains unsafe build flags', remove the .unsafeFlags"
+          echo "::error::from Package.swift -- they make the package unconsumable by any"
+          echo "::error::frontend pinning a release tag, which is the point of shipping it."
+          exit 1
+        fi
+        echo "OK: builds as a versioned dependency"
+
     - name: exactly one SIMD vtable, and it is not the scalar one
       run: |
         set -euo pipefail
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -445,8 +445,8 @@ jobs:
       if: ${{ !matrix.config.emscripten && !matrix.config.android && !matrix.config.cross && runner.os != 'Windows' }}
       run: make -j4 CC="${{ matrix.config.cc }}" CXX="${{ matrix.config.cxx }}" ${{ matrix.config.make_extra || '' }}
 
-  version-fallback-sync:
-    name: version fallback in sync
+  duplicated-lists-in-sync:
+    name: duplicated lists in sync
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v7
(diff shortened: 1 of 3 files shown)


JoeMatt:
All three addressed in 2afaf99. The first one is a real defect and the review is right about why my own verification could not have caught it.

### 1. `unsafeFlags` makes the package unconsumable by version — fixed

Reproduced before changing anything, with a throwaway tagged clone and a minimal consumer:

```
error: 'app': the target 'virtualjaguar-libchdr' in product 'VirtualJaguar'
       contains unsafe build flags
```

The diagnosis is exactly right, including why the verification table was blind to it: `swift build` inside the repo builds this as the **root** package, and the root is exempt. The `spm-build` job had the same blind spot. That is the second time on this branch that a test structurally could not observe the failure it was cited as evidence against — the `boolean.h` correction earlier was the same shape.

The flags only silenced unused-function/variable warnings in libchdr's unity TU, so they are simply gone.

**The more useful part of the fix is a CI job that consumes the package the way a frontend actually would** — throwaway tagged clone, tiny consumer depending on it by version — so the restriction is tested, not remembered. Two things about that job were established by experiment rather than assumed, and both would have made it useless:

- **It must run `swift build`, not `swift package resolve`.** Resolve **succeeds** with unsafe flags present; the graph is only validated for a build. My first draft used `resolve` and passed against a deliberately regressed clone.
- **The clone must sit in a directory named after the repo**, because SwiftPM derives package identity from the URL's last path component, not from `name:` in the manifest. My first draft failed with `unknown package 'VirtualJaguar'`.

Verified both directions after committing:

```
$ swift build          # versioned consumer, unsafeFlags removed
[86/88] Linking app
Build complete! (5.09s)

$ swift build          # same consumer, unsafeFlags reintroduced in the clone
error: ... contains unsafe build flags
```

### 2. `set -u` and the bare `rm` in the trap — fixed

Agreed, and the "precedent other guards get copied from" framing is the right reason. Now `set -eu`, and `command rm -rf` on `EXIT INT TERM`. An unset variable expanding to empty is precisely how a drift guard starts quietly passing on nothing, and this repo's shells alias `rm` to `rm -i`, which blocks forever with no tty.

### 3. The `boolean.h` hunk in the diff — belongs to the parent

Correct, and worth being explicit rather than relying on merge order. That hunk is #624 (`fix/duplicate-boolean-h`), which this branch is stacked on; it is not part of this change. The PR description now says so directly instead of only saying "stacked".

On the risk noted: nothing in the tree still includes `core/boolean.h` after #624 — verified that `boolean.h` was the only header basename shared between `src/core/` (24 headers) and `libretro-common/include/` (23), and the intersection is now empty, so `<boolean.h>` resolves to libretro-common's copy everywhere. But the ordering constraint stands and #624 should land first regardless.

### On the items checked and cleared

Thanks in particular for confirming the `nm` assertion cannot pass spuriously (`total=0` on a wrong path still fails `-eq 1`) and that `tail -40` is safe under Actions' `-eo pipefail`. Both were deliberate and it is useful to have them independently checked rather than assumed.

Issue #623, opened 2026-08-26
[build] Duplicate src/core/boolean.h shadows libretro-common's and cycles any modules build


JoeMatt:
## Problem

`src/core/boolean.h` is a 2015 copy of `libretro-common/include/boolean.h` — the only
header basename shared between those two directories. `INCFLAGS` lists `src/core` before
`libretro-common/include`, so every `#include <boolean.h>` in the tree has been resolving
to the stale copy.

Harmless for the Makefile build. Fatal for any modules-based one: when the core is a Clang
module rooted at `src/core` (SwiftPM, and Xcode via the Provenance SPM wrapper),
`libretro-common/include/encodings/utf.h` does `#include <boolean.h>`, which binds back
into the core's own module:

```
Cyclic dependency in module 'libjaguar':
    libjaguar -> libretro_common -> libjaguar
```

## Why this is easy to miss

A plain `swift build` of a C target does **not** reproduce it. Those compile as ordinary
translation units with no `-fmodules`, so the module map is never built and the cycle never
forms. It only appears once something actually imports the module — the wrapper's bridge
target, which passes `-fmodules`.

Worth recording because it makes "I built it and it was fine" not evidence here.

## Fix

Delete `src/core/boolean.h` and let `<boolean.h>` resolve to libretro-common's copy.

The two differ only in the MSVC guard, and libretro-common's is newer and more correct:

| | condition for the `#define bool unsigned char` hack |
|---|---|
| `src/core/boolean.h` (2015) | `defined(_MSC_VER) && !defined(SN_TARGET_PS3)` |
| `libretro-common/include/boolean.h` (2020) | `defined(_MSC_VER) && _MSC_VER < 1800 && ...` |

For msvc05/msvc10 (`_MSC_VER` 1400/1600) both take the same branch, so the legacy GitLab
buildbot targets are unaffected. Only MSVC ≥ 1800 changes, and it changes to real
`<stdbool.h>` instead of the macro hack — the more correct behaviour, and what every
non-MSVC target already gets.

## Scope check

`boolean.h` was the only shared basename: `src/core/` has 24 headers,
`libretro-common/include/` has 23, and after this the intersection is empty. So no other
cycle of this shape remains possible.

## Context

Related to the SPM consolidation tracked in #614; sibling to #621/PR #622, which fixes a
separate dangling-header defect in the same area.

PR #629, opened 2026-08-27
Fix build issues by removing unused and duplicate headers

JoeMatt:
This pull request removes two header files from the codebase: `libretro-common/include/vfs/vfs_implementation_cdrom.h` and `src/core/boolean.h`. These files are no longer needed, likely due to refactoring or cleanup of unused or redundant code.

Removed header files:

* Removed the `vfs_implementation_cdrom.h` header, which contained CD-ROM related VFS function declarations.
* Removed the `boolean.h` header, which provided a compatibility layer for boolean types in C.

## Test plan

- [ ] `make -j$(getconf _NPROCESSORS_ONLN)` builds clean on host
- [ ] `make test` passes (or N/A — explain why)
- [ ] `bash scripts/c89-lint.sh` passes (no mid-block declarations)

## Branch base

- [ ] **Base is `develop`** (the integration branch)
- [ ] OR base is `master` *and* this is a `hotfix/*` or `release/*` branch


JoeMatt:
Superseded — closing in favour of #624, which does the same thing with the rationale and the fix that makes CI pass.

**These are duplicates now.** This PR originally removed two headers, but `vfs_implementation_cdrom.h` landed separately via #622, so its remaining diff is exactly one file: `src/core/boolean.h`. That is #624.

Keeping #624 rather than this one because it carries three things this branch doesn't:

1. **The reasoning**, which is not obvious from the diff — the two `boolean.h` copies differ in their MSVC guard (`_MSC_VER` unconditional vs `< 1800`), and `src/core` precedes `libretro-common/include` on the include path, so the stale copy was winning everywhere. msvc05/msvc10 are unaffected either way; only MSVC ≥ 1800 changes, and to the more correct real `<stdbool.h>`.
2. **A linked issue** (#623), which `pr-issue-link.yml` requires.
3. **The CI fix this change needs and neither branch had.**

On that last point — this deletion breaks four jobs on its own:

```
test/../src/tom/blitter_simd.h:29:10: fatal error: 'boolean.h' file not found
```

`build-macOS arm64`, `build-Linux aarch64`, and both Windows MSYS2 builds fail at *Run SIMD blitter tests*. Those two by-hand compiles in `c-cpp.yml` pass only `-I src -I src/core -I src/tom`, and were resolving `<boolean.h>` **through the duplicate we are deleting**. `Makefile.common` has always passed `-I libretro-common/include`, so no normal build was affected — which is precisely how a shadowing header sits in a tree for years unnoticed.

Fixed in `ef2b5db` on #624 by adding the include path those compiles were always missing. Verified with the exact CI command: `56103 tests, 56103 passed, 0 failed`, both the arch build and the scalar cross-check.

If this branch is merged instead, the same four jobs will fail.

(`spm-zero` was a scratch branch I pushed to stage the downstream SPM work — it was not meant as a PR source, sorry for the confusion.)

PR #658, opened 2026-08-27
fix(op): clamp the object window to the visible window, not VBB (fixes #641 regression)

### LINKED ISSUE #632: [bug] Super Burnout: HUD's first text row loses its top scanline (LAP renders as LHP)
## Summary

Super Burnout's in-race HUD loses the **topmost scanline of its first text row**. `LAP` renders
as `LHP`, `POSITION` as `POSTTTON`. Every glyph on that row is missing exactly one pixel row off
the top; the rows below it are intact.

Reported by a user; reproduced headlessly.

## Evidence

Build `59e37e9`, private ROM `Super Burnout (1995).jag`, gameplay savestate, frame 10.
Framebuffer is 326x240. Pixels matching the white HUD text colour, `x=14..52`:

```
     ...LL......AAAAA..PPPPP...
 32  ......................................   <- MISSING
 33  ...##......##...##.##...##............
 34  ...##......##...##.##...##............
 35  ...##......#######.######.............
 36  ...##......##...##.##.................
 37  ...##......##...##.##.................
 38  ...#######.##...##.##.................
```

`A` has no apex, so it reads as `H`. `P`'s bowl is open at the top.

The decisive glyph is the `O` in `POSITION` (`x=88..150`, same rows): it keeps its **bottom**
closing row at 38 but has no top row. So the glyph is clipped at the top by one line — it is not
shifted, and it is not a different 6-row font.

The row is genuinely absent, not overdrawn: no text-coloured pixel exists anywhere on scanline 32.

## What this is NOT

- **Not a blitter-engine bug.** Accurate and fast blitters produce a **byte-identical** framebuffer
  here (`nonuniform=11294 pct=57.7403` for both). Note this does not exonerate the blitter
  *entirely* — both engines share the register decode in `src/tom/blitter_mmio.c`, so a shared
  A1/A2/count off-by-one would also hit both. It only rules out the two inner loops.
- **Not every text object on screen.** The blue lap-time line directly below (`0'05"07`, rows
  40-46) is a full **7 rows** and is correct. `SPEED:` at rows 80-86 is a full 7 rows and correct.
  Only the first HUD text row loses a line.

That asymmetry is the useful clue: whatever draws the `LAP`/`POSITION` row starts one line lower,
or is one line shorter, than what draws the two rows that are fine.

## Repro

```bash
python3 scripts/rzip_extract.py "Super Burnout (1995).state1" /tmp/sb.state
./test/tools/hires_shot ./virtualjaguar_libretro.dylib "Super Burnout (1995).jag" \
    --load-state /tmp/sb.state --frames 20 --shot 10 --out-prefix /tmp/sb
```

Then inspect scanlines 30-42 of `/tmp/sb_f00010.ppm`.

## Next step

Establish whether that row is drawn by the blitter or composited by the Object Processor, and if
by the OP, whether the object's `YPOS`/height accounting drops its first line. `vjtrace`
(`make TEST_EXPORTS=1`) emits `OP_OBJECT` events and can name the object covering `y=32`.

JoeMatt:
**#641 introduced a regression on roughly a third of the corpus. This fixes it and keeps #632 fixed.**

## What went wrong

#641 clamped `endingHalfline` to `VBB` when `VDE` is past the end of the field. On NTSC, blanking
begins at **VBB=500** while the visible window runs to **511** — so the clamp stopped object
processing five halflines (about three scanlines) *before the display actually ends*.

Measured against the pre-#641 core across the private corpus at 300 frames: about a third diverged.
Every divergence I sampled had the same signature — the last rows of the frame, and fewer non-black
pixels:

| title | differing rows | non-black, base → #641 |
|---|---|---:|
| Hover Strike (652×480, 2x) | **474–479** | 57196 → 56286 |
| JagFest Demo (326×240) | **237–239** | 19082 → 18842 |
| JSSDemo (326×240) | **237–239** | 7209 → 6973 |

The bottom three scanlines stopped being drawn.

## The fix

Clamp to `TOMGetBottomVisible()` instead — the same value the line-copy loop further down that
function uses to decide when to stop reading.

That makes it correct by construction rather than by measurement: a pass at or past
`bottomVisible` fills a line buffer nobody ever copies out, so declining to run it **cannot** change
the picture, while running it *can* corrupt the next field. There is no gap to tune.

Super Burnout's `VDE=$7FF` resolves to the NTSC fallback of 511 — still well below the halfline-522
pass that was eating its HUD object's first line, so #632 stays fixed.

## Verification

| check | result |
|---|---|
| Super Burnout HUD | all seven rows — `LAP`, not `LHP` |
| Hover Strike | byte-identical to pre-#641 |
| JagFest Demo | byte-identical to pre-#641 |
| JSSDemo | byte-identical to pre-#641 |
| Iron Soldier | byte-identical to pre-#641 |

## What I got wrong in #641, and why

#641 states "Super Burnout is the only title in the corpus that sets `VDE` past the end of the
field." That was measured on **five ROMs**, and it did not survive 160. `docs/jtrm-register-map.md`
calls `VDE=$FFFF` *standard practice* — it is common, not rare. I had the sentence that contradicted
me in the tree already and read it as background instead of as a population estimate.

The corpus sweep that would have caught this before merge was running the whole time and produced
nothing: it wedged on `Chroma-Luma Color Pick (PD).jag`, which hangs `frame_hash_ab` indefinitely,
and sat there for 2h23m with no timeout. I should have treated "still running" as a result worth
checking rather than as progress. The sweep now runs every case under `timeout` and prints progress
— worth folding into a reusable script separately, along with an issue for the ROM that hangs the
core.

### DIFF
--- a/src/tom/tom.c
+++ b/src/tom/tom.c
@@ -1552,22 +1552,28 @@ void TOMExecHalfline(uint16_t halfline, bool render)
        * Super Burnout (#632) is the case that exposed this.  It sets
        * VDB=25, VDE=$7FF, VP=523, VBB=500, VI=521, and rebuilds the
        * display list in its vertical-interrupt handler.  The OP then ran
-       * once more at halfline 522 -- 22 halflines into blanking, one
-       * halfline after VI -- consuming line 0 of the freshly written HUD
-       * object.  The visible result was the top scanline missing from the
-       * whole first HUD text row, so LAP rendered as LHP and POSITION as
-       * POSTTTON, identically on both blitters because the blitter was
-       * never involved.
+       * once more at halfline 522 -- one halfline after VI -- consuming
+       * line 0 of the freshly written HUD object.  The visible result was
+       * the top scanline missing from the whole first HUD text row, so LAP
+       * rendered as LHP and POSITION as POSTTTON, identically on both
+       * blitters because the blitter was never involved.
        *
-       * Clamping to VBB is the narrow reading: blanking begins there, so
-       * that is where the object list stops being displayed.  It applies
-       * only on this branch -- a title with a real in-range VDE keeps
-       * exactly the window it asked for, including one that deliberately
-       * extends past VBB. */
+       * Clamp to the VISIBLE window, which is where the line-copy loop
+       * below stops reading: a pass at or past bottomVisible produces a
+       * line buffer nobody ever copies out, so declining to run it cannot
+       * change the picture, while running it can corrupt the next field.
+       *
+       * Not VBB.  Clamping to VBB was the first attempt and it was wrong:
+       * blanking begins before the display window ends on this hardware
+       * (NTSC VBB=500 against a visible window running to 511), so it cut
+       * the bottom three scanlines off every title using the VDE=$FFFF
+       * idiom -- about a third of the private corpus, including Hover
+       * Strike, JSSDemo and JagFest Demo.  "Only Super Burnout does this"
+       * was measured on five ROMs and did not survive 160. */
       {
-         uint16_t vbb = GET16(tomRam8, VBB);
-         if (vbb > 0 && vbb < endingHalfline)
-            endingHalfline = vbb;
+         uint16_t visible_end = TOMGetBottomVisible();
+         if (visible_end > 0 && visible_end < endingHalfline)
+            endingHalfline = visible_end;
       }
    }
 


JoeMatt:
## Full corpus sweep: clean

Pre-#641 core vs this branch, every `.jag`/`.j64`/`.rom` in the private library, 300 frames from
cold boot, complete framebuffer-hash stream compared.

```
---- identical: 158   differing: 1   skipped: 1   (of 160) ----

DIFF  [138/160] Super Burnout (1995).jag        base=143 fix=143
SKIP  [ 40/160] Chroma-Luma Color Pick (PD)     (base timeout)
```

**One title differs, and it is the one this is meant to change.** Super Burnout's transition count
is unchanged at 143/300 — same motion, different pixels — which is the expected shape for "the HUD
object's first line is now drawn."

Compare with the same sweep against #641 as merged, which diverged on **roughly a third of the
corpus** with the bottom three scanlines missing.

The skip is a separate pre-existing problem: `Chroma-Luma Color Pick by Matthias Domin (bin) (1996)
(PD).jag` hangs `frame_hash_ab` indefinitely — the first version of this sweep sat on it for 2h23m
producing nothing, which is why #641's regression reached develop unverified. Filed separately;
the sweep now runs every case under `timeout`.

PR #663, opened 2026-08-27
fix(blitter): refuse a nested blit dispatched from inside a running one

### LINKED ISSUE #659: [bug] Chroma-Luma Color Pick (PD) hangs retro_run forever in blitter_generic — watchdog cannot see it
## Summary

`Chroma-Luma Color Pick by Matthias Domin (bin) (1996) (PD).jag` makes `retro_run()` **never
return**. Not a black screen, not a wedge the watchdog reports — the call does not come back, so a
frontend calling it is frozen with no way out but a kill.

Found because it silently destroyed a regression sweep: the first corpus A/B run for #632 sat on
this one ROM for **2 hours 23 minutes** producing no output, and the #641 regression reached
`develop` unverified as a direct result.

## Where it hangs

`sample` on the stuck process, 3 s, main thread — 2435 of 2451 samples:

```
main  (in frame_hash_ab)
  blitter_generic  (in virtualjaguar_libretro.dylib) + 14840,14836,...
  JaguarReadWord   (in virtualjaguar_libretro.dylib) + 220,412,...
```

An unbounded loop inside the blitter, reading memory. Not the GPU, not the 68K, not the OP.

## Characteristics

| | |
|---|---|
| first hangs between | frame **31 and 35** (30 frames complete, 35 does not) |
| accurate blitter | hangs |
| **fast blitter** | **also hangs** |
| pre-#641 core | hangs — **pre-existing**, not caused by the OP window change |
| `virtualjaguar_crash_detect=enabled` | **prints nothing** |

The watchdog silence is the important part. `crash_detect` reports `gpu_wedge`, `dsp_wedge`,
`video_stall` and friends from *between* frames, so a loop that never exits `retro_run` is exactly
the shape it cannot see. Every existing wedge signature assumes the frame loop still turns.

## Why this matters beyond one PD ROM

- In RetroArch this is a **hard freeze of the frontend**, which is worse than any of the crash
  classes the watchdog was built for.
- It means "the harness is still running" is not evidence of progress. Any long sweep over the
  corpus needs a per-run `timeout`, and any corpus tooling that lacks one can report a clean result
  by simply never reaching the interesting ROMs.

## Repro

```bash
gtimeout 20 ./test/tools/frame_hash_ab ./virtualjaguar_libretro.dylib \
  "Chroma-Luma Color Pick by Matthias Domin (bin) (1996) (PD).jag" \
  --csv /tmp/cl.csv --frames 60          # times out; --frames 30 completes
```

## Suggested direction

Two independent pieces of work:

1. **The blitter loop needs a bound.** Whatever inner/outer count this ROM programs, the engine
   should refuse it rather than spin — the OP already has `OP_RUNAWAY_GUARD_OBJECTS` for the
   equivalent malformed-list case, and the blitter has no counterpart. A guard here also gives
   `crash_detect` something to report.
2. **Corpus tooling should bound every run.** A shared sweep runner with a per-ROM timeout, so a
   hang shows up as one `SKIP` line rather than as a silent absence of results.

JoeMatt:
Fixes the `retro_run` freeze in #659.

## Not an infinite loop — the blitter re-enters itself

`blitter_write_word()` / `_long()` route any destination at or above `$200000` through
`JaguarWrite*()`, which lands **back in `BlitterWriteWord()`**. A blit whose destination runs into
`$F022xx` therefore writes the blitter's own register file as pixel data — and a pixel landing on
B_CMD's low word (offset `$3A`, the dispatch trigger) starts a **second blit from inside the first**.

Traced, with the 68K's setup and the blitter's own writes distinguished by `who`:

```
BW off=$38 data=$2704 who=6(M68K)
BW off=$3A data=$2704 who=6(M68K)
  TRIGGER depth=0 cmd=$27042704 count=$2704A7E4 (px=42980 lines=9988)
BW off=$38 data=$0000 who=7(BLITTER)      <- blit destination lands on B_CMD
BW off=$3A data=$0000 who=7(BLITTER)
  TRIGGER depth=1 cmd=$00000000 count=$2704A7E4    <- nested, same count latched
```

429M pixels per blit, re-entered, with nothing bounding the nesting.

## The guard

Hardware cannot do this. The blitter is **one state machine and one bus master** — it is not
re-entrant, and a write to B_CMD while it is running cannot start a nested blit on top of the
running one.

So the guard refuses only the *nested dispatch*. The register write itself is real and still
happens. The warning fires once per session, never per pixel.

## Why not a cap on blit size

That was the obvious first idea and it is the wrong one:

- **The blit is finite** — 429M pixels ≈ 16 s of Jaguar time. Hardware would complete it. Truncating
  it would be a visible deviation from hardware, not a fix.
- **Abandoning a blit part-way leaves A1/A2 and the counters in a state hardware never reaches**, and
  those are savestated — precisely the determinism that run-ahead (#400) and netplay depend on.
- This guard is transient within a single dispatch, always zero between frames, and therefore has
  **no savestate surface at all**.

The OP's `OP_RUNAWAY_GUARD_OBJECTS` precedent doesn't transfer either: that bounds a structurally
*infinite* walk, which is a different class from finite-but-enormous.

## Results

| check | before | after |
|---|---|---|
| 600 frames of the ROM | 2h23m+ hang | **5.95 s**, full speed |
| corpus A/B, 300 frames, all 160 ROMs | — | posted below |
| `test_audio_presence` (Iron Soldier) | RMS 1174.2 | RMS 1174.2 |
| `test_audio_clipping` | PASS | PASS |

Both audio tests run because the change sits in the shared dispatch site, which is audio-adjacent.

## Scope

The ROM still renders garbage — a 1024-pixel-wide frame of green scanline noise — and is very
likely a bad dump (it is a raw `(bin)` rip). Rendering it correctly is not in scope. **Not hanging
the frontend on a bad dump is**, and that is what this fixes: in RetroArch the current behaviour is
an unrecoverable freeze that `crash_detect` cannot even report, because every watchdog signature is
checked between frames and this never reaches one.

### DIFF
--- a/src/tom/blitter_mmio.c
+++ b/src/tom/blitter_mmio.c
@@ -322,6 +322,49 @@ void BlitterWriteWord(uint32_t offset, uint16_t data, uint32_t who/*=UNKNOWN*/)
        * shadow registers as they run. */
       uint32_t busClks = 0;
       int trBlit = 0;
+
+      /* Re-entrancy guard (issue #659).
+       *
+       * A blit whose destination lands in $F022xx writes the blitter's own
+       * registers as pixel data -- blitter_write_word()/_long() route
+       * anything at or above $200000 through JaguarWrite*(), which comes
+       * straight back here.  A pixel landing on B_CMD's low word (offset
+       * $3A) then dispatches a SECOND blit from inside the first.
+       *
+       * Hardware cannot do that.  The blitter is one state machine and one
+       * bus master; it is not re-entrant, and a write to B_CMD while it is
+       * running cannot start a nested blit on top of the running one.  The
+       * register write itself is real and still happens -- only the nested
+       * dispatch is refused.
+       *
+       * Chroma-Luma Color Pick (PD) is the case that exposed it: the 68K
+       * programs B_COUNT = $2704A7E4 (42980 x 9988 = 429M pixels), the blit
+       * runs away into the register file, writes $0000 over B_CMD, and
+       * re-enters with the same count still latched. Traced at depth=1, but
+       * nothing bounded the nesting, and retro_run never returned -- a hard
+       * freeze of the frontend, invisible to crash_detect because every
+       * watchdog signature is checked between frames.
+       *
+       * Deliberately NOT a cap on blit size. That blit is finite (~16s of
+       * Jaguar time) and hardware would complete it, so truncating it would
+       * be a visible deviation; worse, abandoning a blit part-way leaves
+       * A1/A2 and the counters in a state hardware never reaches, and those
+       * are savestated (the determinism run-ahead and netplay depend on).
+       * This guard is transient within a single dispatch, always zero
+       * between frames, and so has no savestate surface at all. */
+      static int blit_in_progress = 0;
+
+      /* Silent on purpose.  A warning here would be genuinely useful --
+       * a title that reaches this is misprogramming its destination --
+       * but this file is compiled STANDALONE by test/test_blitter_mmio,
+       * outside CFLAGS and without linking the core.  src/core/log.h
+       * needs -DINLINE and resolves to vj_log_cb, so including it turns
+       * that target into a link error.  Keeping blitter_mmio.c free of
+       * core dependencies is the existing design; do not add logging
+       * here without giving that test a stub. */
+      if (blit_in_progress)
+         return;
+
       if (vjs.blitterTiming)
          busClks = BlitDurationSysclks();
 
@@ -340,6 +383,7 @@ void BlitterWriteWord(uint32_t offset, uint16_t data, uint32_t who/*=UNKNOWN*/)
       if (texReplaceEnabled)
          trBlit = TexReplacePreBlit();
 
+      blit_in_progress = 1;
       if (BlitterCompareIsEnabled())
          BlitterRunComparison();
       else if (!BlitMemoLaunch())
@@ -350,6 +394,7 @@ void BlitterWriteWord(uint32_t offset, uint16_t data, uint32_t who/*=UNKNOWN*/)
          else
             BlitterMidsummer2();
       }
+      blit_in_progress = 0;
 
       if (trBlit)
          TexReplacePostBlit();


JoeMatt:
## Corpus sweep: no other title changes

`develop` vs this branch, all 160 ROMs, 300 frames, full framebuffer-hash stream compared:

```
---- identical: 159   differing: 0   skipped: 1   (of 160) ----

SKIP  [ 40/160] Chroma-Luma Color Pick by Matthias Domin (bin) (1996) (PD)  (base timeout/fail)
```

**Zero differences.** The guard only ever fires on a blit that writes its own command register, so
nothing else in the corpus can reach it — which is what the sweep confirms rather than assumes.

The one SKIP is the subject ROM itself, and the reason is worth stating plainly: it is skipped
because the **base** core cannot complete it. `corpus_ab_sweep.sh` needs both sides to finish in
order to compare, and on `develop` this ROM never returns from `retro_run`. Measured separately:

| | 600 frames |
|---|---|
| `develop` | hangs (2h23m+, killed) |
| this branch | **5.95 s** |

So the sweep's own SKIP line is a restatement of the bug.

JoeMatt:
## CI fixed — and the failure was mine, in the part I added last

Every build job reported `fail`, but no build broke. The failing *step* inside each was
"Run cheat engine unit tests" (`make test`):

```
src/tom/../core/log.h:14:8: error: unknown type name 'INLINE'
```

`test/test_blitter_mmio` compiles `src/tom/blitter_mmio.c` **standalone** — outside `CFLAGS`,
without the core's `-DINLINE="inline"`, and without linking the core. The once-per-session
`LOG_WRN` I had added pulled in `src/core/log.h` and broke that target on every platform.

I tried supplying `-DINLINE=inline` to the test rule first. It compiled, then failed to link:

```
Undefined symbols: "_vj_log_cb", referenced from: _BlitterWriteWord
```

Which settles it: `blitter_mmio.c` being free of core dependencies is deliberate, not incidental.
So the log line is gone, the Makefile change is reverted, and the reason is recorded in the code so
the next person to want a warning there knows it needs a `vj_log_cb` stub in that test first.

**The guard itself is unchanged.** Only the diagnostic I had bolted onto it.

## Now verified

| check | result |
|---|---|
| `make test` | **0 failed**, 2 skipped (both optional, missing private assets) |
| `test/test_blitter_mmio` standalone | builds, `All tests passed.` |
| corpus A/B, 300 frames, 160 ROMs | 159 identical, 0 differing, 1 skipped |
| 600 frames of the subject ROM | 2h23m+ hang → **5.95 s** |
| `test_audio_presence` / `test_audio_clipping` | unchanged |

## Known tradeoff

The guard is now **silent**. A user hitting a runaway blit gets garbage rendering with no log line
explaining it. Restoring that needs a `vj_log_cb` stub in `test/test_blitter_mmio.c` — a separate
change rather than something to widen this PR with.

## What I should have done

I ran the 160-ROM corpus sweep and both audio tests, and treated that as sufficient. I never ran
`make test`, which `CLAUDE.md` says plainly to run after changes and which would have caught this in
90 seconds. The expensive bespoke check felt thorough enough to skip the cheap standard one.

PR #674, opened 2026-08-27
perf(libretro): hot-path hygiene (Memory Track pack, serialize slack, netlink clock, LOG_DBG)

### LINKED ISSUE #673: [perf] Hot-path hygiene: Memory Track pack, serialize slack, netlink clock, LOG_DBG
Four small hot-path hygiene fixes from `docs/perf-audit/hotpath-followups-2026-08.md` (F2–F5). F1 (software framebuffer) needs a device A/B first; F6 (CDDA residue) is still an open investigation; F7 (fast-forward audio skip) is not worth a standalone PR.

- **F2** — Memory Track `mt_dirty_cb` was memcpying 128 KB on every AT29C010 byte write. Latch a dirty flag and pack at most once per frame.
- **F3** — `retro_serialize` zero-fills ~350–450 KB of slack on every call, including rewind/run-ahead. Skip the memset for rollback/run-ahead savestate contexts; keep it for on-disk states.
- **F4** — The netlink peer-picker block called `JLinkNowMs()` every frame even with netlink disabled (a second `clock_gettime` on top of `JLinkFrameTick`). Gate it.
- **F5** — `LOG_DBG` had no compile-out. Under `NDEBUG` it becomes a no-op unless `VJ_DEBUG_LOG` is set; INF/WRN/ERR stay live.

JoeMatt:
Closes #673

## Summary

Four small hot-path hygiene fixes from [`docs/perf-audit/hotpath-followups-2026-08.md`](https://github.com/libretro/virtualjaguar-libretro/blob/develop/docs/perf-audit/hotpath-followups-2026-08.md) (F2–F5). One commit per finding. F1 / F6 / F7 are intentionally not in this PR.

- **F2** — `mt_dirty_cb` was memcpying 128 KB on every AT29C010 byte write. It now latches a dirty flag; `mt_flush_save_buf` packs at most once per frame and on `retro_get_memory_data`. `retro_serialize` / unserialize / deinit read `mtMem` via `MTStateSave`, not `eeprom_save_buf`, so they do not need a flush.
- **F3** — `retro_serialize` zero-fills ~350–450 KB of slack per call. Query `RETRO_ENVIRONMENT_GET_SAVESTATE_CONTEXT` per serialize (context is not session-stable: run-ahead and a user save interleave) and skip the memset only for `RUNAHEAD_SAME_INSTANCE` / `RUNAHEAD_SAME_BINARY`. Keep it for `NORMAL` (on-disk), `ROLLBACK_NETPLAY` (crosses the network; peers may CRC-compare the blob), unknown contexts, and when the frontend does not implement the env call.
- **F4** — The netlink peer-picker block called `JLinkNowMs()` every frame even with netlink disabled. Guard on `JLinkMode() != JLINK_MODE_DISABLED || netlink_peers_dirty` so a latched peer-set change is delayed, never lost. Enabled-netlink behavior is unchanged.
- **F5** — Under `NDEBUG` (already set by the Makefile for release builds), `LOG_DBG` expands to `do { } while (0)` unless `VJ_DEBUG_LOG` is defined. INF/WRN/ERR stay live (crash-detect signatures are `LOG_ERR`/`LOG_WRN`). Call-site args are format strings and register reads — no required side effects.

## Test plan

- [x] `bash scripts/c89-lint.sh libretro.c src/core/log.h`
- [x] `DEVELOPER_DIR=/Library/Developer/CommandLineTools make -j8` (clean)
- [x] `DEVELOPER_DIR=/Library/Developer/CommandLineTools make TEST_EXPORTS=1 test` — 0 failed, 2 skipped (Fountain live abort: no `/tmp/fountain_vj.j64`; jagcd CUE→CHD: no CHSE-capable chdman)
- [x] `./test/test_memtrack` — 28 checks, 0 failed
- [x] `./test/test_nvmbios` — 40 checks, 0 failed
- [x] `./test/regression_test.sh ./virtualjaguar_libretro.dylib` — 10 passed, 0 failed (screenshot, determinism, frameskip, savestate round-trip, rewind)
- [x] Follow-up `b7ea33c`: re-ran lint + `make TEST_EXPORTS=1 test` after dropping `ROLLBACK_NETPLAY` from the skip — 0 failed, 2 skipped

### DIFF
--- a/libretro.c
+++ b/libretro.c
@@ -4157,6 +4157,7 @@ bool retro_serialize(void *data, size_t size)
    uint8_t *buf, *start;
    size_t written;
    uint32_t magic, version, flags, reserved;
+   enum retro_savestate_context ss_ctx;
    extern uint8_t jerry_ram_8[];
    extern bool lowerField;
 
@@ -4283,9 +4284,22 @@ bool retro_serialize(void *data, size_t size)
       }
    }
 
-   /* Zero-fill remaining bytes for deterministic save states */
+   /* Zero-fill remaining bytes for deterministic on-disk save states.
+    * Rewind / run-ahead / rollback-netplay serialize every frame and
+    * never persist or compare slack, so skip the 350-450 KB memset
+    * there.  Query per call: RetroArch's savestate context is not
+    * session-stable (run-ahead and a user save interleave).  A frontend
+    * that does not implement GET_SAVESTATE_CONTEXT leaves ss_ctx at
+    * NORMAL and keeps the zero-fill. */
    if (written < STATE_SIZE)
-      memset(buf, 0, STATE_SIZE - written);
+   {
+      ss_ctx = RETRO_SAVESTATE_CONTEXT_NORMAL;
+      if (!(environ_cb(RETRO_ENVIRONMENT_GET_SAVESTATE_CONTEXT, &ss_ctx)
+            && (ss_ctx == RETRO_SAVESTATE_CONTEXT_ROLLBACK_NETPLAY
+                || ss_ctx == RETRO_SAVESTATE_CONTEXT_RUNAHEAD_SAME_INSTANCE
+                || ss_ctx == RETRO_SAVESTATE_CONTEXT_RUNAHEAD_SAME_BINARY)))
+         memset(buf, 0, STATE_SIZE - written);
+   }
 
    return true;
 }
--- a/libretro.c
+++ b/libretro.c
@@ -4284,10 +4284,14 @@ bool retro_serialize(void *data, size_t size)
       }
    }
 
-   /* Zero-fill remaining bytes for deterministic on-disk save states.
-    * Rewind / run-ahead / rollback-netplay serialize every frame and
-    * never persist or compare slack, so skip the 350-450 KB memset
-    * there.  Query per call: RetroArch's savestate context is not
+   /* Zero-fill remaining bytes for deterministic save states.  Skip
+    * only for same-session run-ahead (SAME_INSTANCE / SAME_BINARY):
+    * those buffers never leave the process, are never compared, and
+    * serialize every frame.  Keep the memset for NORMAL (on-disk) and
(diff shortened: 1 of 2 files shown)


JoeMatt:
Review finding (blocking): the zero-fill skip must NOT include `RETRO_SAVESTATE_CONTEXT_ROLLBACK_NETPLAY`.

Per libretro.h (5718-5730), a rollback-netplay state "will almost certainly be loaded by a separate binary, device, and address space" — i.e. it crosses the network, and peers may CRC-compare serialized buffers for desync detection. Skipping the slack memset leaves whatever bytes the frontend's reused buffer held, which differs per machine → false desync signals.

The skip is only safe for `RUNAHEAD_SAME_INSTANCE` and `RUNAHEAD_SAME_BINARY` (guaranteed same-session, never written to disk or network). Please drop `ROLLBACK_NETPLAY` from the condition.

Everything else in this PR reviews clean: NDEBUG is confirmed defined in release builds (Makefile:850-853, jni/Android.mk:53) so the LOG_DBG gate is live; the Memory Track dirty-flag flush points (retro_run post-emulation, retro_get_memory_data, eeprom_pack_save_buf) cover all consumers; the netlink clock gate preserves enabled-path behavior.

JoeMatt:
Addressed in b7ea33c (`fix(libretro): keep serialize slack zero-fill for rollback netplay`).

`RETRO_SAVESTATE_CONTEXT_ROLLBACK_NETPLAY` is no longer in the skip. Slack memset is skipped only for `RUNAHEAD_SAME_INSTANCE` / `RUNAHEAD_SAME_BINARY`. `NORMAL` and rollback-netplay still zero-fill, matching libretro.h 5718–5730 (rollback states cross the network and peers may CRC-compare the blob).

Re-ran `c89-lint` on `libretro.c` and `make TEST_EXPORTS=1 test`: 0 failed, 2 skipped (same optional skips as before).

Issue #611, opened 2026-08-26
[bug] Club Drive freezes in gameplay at STOCK settings — GPU runs away into main RAM ($00041E3x)


JoeMatt:
Reported by the user playing in RetroArch: **Club Drive freezes shortly into gameplay at stock settings.** Reproduced headlessly from their savestates.

This is **not** #565 (Club Drive aborting under `risc_clock_scale=2x`, fixed in v3.5.0). That needed an overclock; this needs nothing — every run below is stock.

## Reproduction

Two of the three user savestates captured the machine **already wedged**:

| savestate | framebuffer transitions / 1800 frames |
|---|---|
| `Club Drive (1994).state1` | **0** |
| `Club Drive (1994).state.auto` | **0** |
| `Club Drive (1994).state3` | 1471 (healthy) |

Zero transitions means the picture is genuinely frozen — this is not a watchdog false positive.

```
frame_hash_ab <core> "Club Drive (1994).jag" --csv out.csv \
  --load-state "Club Drive (1994).state1" --frames 1800 \
  --option virtualjaguar_crash_detect=verbose
```

```
[CRASH-DETECT] gpu_runaway  frame=1   pc=$00041E4E (not local RAM, not a GPU-GO page)
[CRASH-DETECT] video_stall  frame=301 fb_hash=$659217D0 unchanged for 300 frames
                            gpu_pc=$00041E3E gpu_run=1 dsp_pc=$00F1B124 dsp_run=1
```

The GPU is **running** (`gpu_run=1`), not halted, spinning somewhere around `$00041E38-$00041E4E` — main RAM, not GPU local RAM at `$F03000`. The DSP is also still running. So the machine is live and the GPU is executing a region it has no business executing.

## What is exonerated

**The three experimental timing models are not the cause.** With `dram_timing`, `gpu_pipeline_timing` and `blitter_timing` all enabled the trace is identical — same wedge, same PC range, same stall frame:

```
[CRASH-DETECT] gpu_runaway  frame=1   pc=$00041E38
[CRASH-DETECT] video_stall  frame=301 ... gpu_pc=$00041E38 gpu_run=1
```

**It is not progressive / time-based.** From the healthy `state3`, **12,000 frames** (~200 emulated seconds) run completely clean: no watchdog events, 9756 transitions, and the rate is linear throughout (1632 per 2000 frames at every checkpoint). So the game does not simply rot after N frames — something specific between `state3` and `state1` triggers it.

[...]

Issue #565, opened 2026-08-22
[bug] Club Drive aborts under virtualjaguar_risc_clock_scale=2x (dsp_pc_escape, PC=$00FFFFFF)


JoeMatt:
Found while running #378's gate-1 measurement grid.

## Symptom

Club Drive with the shipped `virtualjaguar_risc_clock_scale` option set to **2x** drives the DSP program counter to `$00FFFFFF` and the process aborts (SIGABRT, exit 134).

```
[CRASH-DETECT] dsp_pc_escape frame=142 pc=$00FFFFFF (valid: $0-$E3FFFF or $F1B000-$F1CFFF)
```

## Reproduce

```bash
make TEST_EXPORTS=1 -j$(getconf _NPROCESSORS_ONLN)
VJ_EXPECT_BUILD=$(./scripts/build-id.sh) ./test/tools/vjtrace_smoke \
  ./virtualjaguar_libretro.dylib "Club Drive (1994).jag" \
  --frames 900 --option virtualjaguar_risc_clock_scale=2x --quiet
```

## Scope — it is specific, not general

| | result |
|---|---|
| Club Drive @ **1x** | exit 0, clean |
| Club Drive @ **1.5x** | exit 0, clean |
| Club Drive @ **2x** | **exit 134**, `dsp_pc_escape` |
| Cybermorph @ 2x | exit 0 |
| I-War @ 2x | exit 0 |
| Missile Command 3D @ 2x | exit 0 |

So it is not "2x breaks everything" — it is one title, and only at the top scale.

Deterministic: 6/6 runs aborted at `frame=142`, same PC. (One earlier pair of runs in the same session reported `frame=742`; I could not account for the difference, so treat the exact frame as **not fully pinned** even though the crash itself is completely reproducible.)

## Why this matters beyond one title

`virtualjaguar_risc_clock_scale` is a **shipped user-facing option**, and 2x is one of its four selectable values. The option's own description says "May break titles that depend on stock RISC timing… Report bugs only at 1x" — which sets the expectation that a title may misbehave, not that the core aborts the process.

This is also the second member of the same family: `m68k_clock_scale` at 3x drives a GPU runaway in Defender 2000 (#460/#461).

## Relationship to #378

#378 proposed shipping per-title overclock presets. That issue is being closed as a measured null — none of its candidates clear gate 1 — so this crash is **not** blocking a feature. It is a standalone defect in an option users can already set today.

[...]

Issue #460, opened 2026-08-15
Defender 2000 hangs at level start with m68k_clock_scale=3x — GPU runaway into a data buffer


JoeMatt:
## Minimal reproduction: `virtualjaguar_m68k_clock_scale=3x`, on its own

Nothing else is required. Reported by a user on iOS with overclocks + all enhancements on and all timing models off; the enhancements turn out to be irrelevant.

| configuration | result |
|---|---|
| stock | clean, 4000 frames, two input schedules |
| m68k 0.5x/1x/1.5x/2x x risc 0.5x/1x/1.5x/2x | clean, all pairings |
| **m68k 3x x risc 0.5x/1x/1.5x/2x** | **hangs, all four** |
| 2x / true_color / 2x+true_color | bit-identical freeze profile to overclock alone |
| dram / pipeline / both / blitter timing | all still hang |

**The enhancement path is exonerated.** Overclock-alone and overclock+both enhancements freeze at the *same* frame 751 for the *same* 1649 frames — hi-res and true colour are inert to the emulated machine, exactly as designed. Defender 2000 (CRC `0x27594C6A`) has no `titledb` row, so `pertitle_defaults` substituted nothing.

Worth noting: **no timing model rescues it.** The option help's advice to "try enabling the timing models" is measurably wrong for this title.

## Mechanism

A GPU task dispatcher in local RAM executes `JUMP T,(R7)` at `$F03366` with **R7 = `$001CB000`** — zeroed main RAM in *both* good and bad runs, a data buffer rather than code. At 3x the 68K outruns the game's own 68K→GPU task handshake, and the GPU consumes a descriptor slot before it has been filled.

The GPU then walks forward through zeros (`$0000` decodes as `ADD R0,R0`) until it is captured by an 8-instruction fragment of ordinary game **data** at `$012784-$012792`: seven `STORE Rn,(R14+off)` followed by `$D714` = `JR NN,-8` back to the start. It never leaves. Rendering never completes, so the framebuffer freezes.

At stock clocks the GPU never leaves local RAM at all — 0 escapes across 4000 frames on two input schedules.

Corroboration that this is the user's bug: the frozen frame is the wave-start banner **"TIME TO KICK IT!"** — the hang lands exactly at level start, which is what they described.

## What did NOT reproduce

The user reported "it glitches then locks **or reboots**". Only the lock reproduced:

[...]

PR #717, opened 2026-08-28
feat(perf): enable RISC idle-loop fast-forward by default (#608)

### LINKED ISSUE #608: [perf] Flip DSP idle-loop fast-forward on by default — the audio + corpus verification #569 left owed
Split out of #569 (perf audit 2026-08), which closed 2026-08-25 with all eight of its tasks merged. P1 — the DSP idle-loop fast-forward, merged via #575 — shipped **disabled by default**, with verification explicitly still owed. That owed work had no tracking issue, so it closed with the epic. This is that issue.

## Current state

`virtualjaguar_risc_idle_skip` exists in `libretro_core_options.h` and is off by default. Its own description says why:

> New in this release and off by default while the compatibility corpus grows — if a title looks or sounds wrong with it on, turn it off and please report it.

So the measured win — **66–87% fewer DSP opcodes per frame across 7 titles** — currently reaches nobody who does not go looking for the toggle. On the RPi 4 / A10X class of device that #509 and #569 both exist to serve, that is the single largest unclaimed speedup we have merged.

## What is owed before the default flips

Both were named in #569's P1 line and neither has been done:

1. **A human audio listening pass.** The fast-forward is bit-exact by construction and a committed regression test falsifies it in both directions, but the soundness gap found in review (the probe verified the *decoded* loop body, not the *executed* path) is exactly the class of bug a headless suite cannot hear. Per the standing rule, `test_audio_clipping` and `test_audio_presence` must both pass, and then a person has to actually listen — headless cannot tell music from structured noise.
2. **Broader corpus and on-hardware FPS numbers.** The 7-title measurement is a sample, not the corpus. Wanted: a pace-neutrality sweep across the cart corpus proving no title changes speed with it on, plus real FPS deltas on a slow target rather than desktop instruction counts.

## Acceptance

- Both audio tests green, plus a recorded human listening pass on at least the titles where the DSP share is largest (Iron Soldier, AvP, Doom).
- Corpus sweep showing bit-identical output with the option on, or an explicit list of titles that diverge.
- FPS measured on real slow hardware, not extrapolated.
- Then, and only then, flip the default to enabled and update the option text to drop the "new in this release, off by default" clause.

If the corpus sweep turns up divergent titles, the fallback is a known-bad row in the per-title DB (#464's negative class) rather than keeping the option off for everyone.

## Notes

The option is already ignored automatically when DRAM timing, GPU pipeline timing, a non-stock RISC clock scale, or blit memoization is in use, so the flip does not interact with the accuracy toggles.

Related: #569 (epic, closed) · #575 (the merge) · #509 (why it matters) · #601 (the A10X capture that would give the hardware FPS half of this)

JoeMatt:
Closes #608

## Summary

#575 shipped the RISC idle-loop fast-forward **disabled**, with verification explicitly owed. This discharges that verification and flips the default.

The stated reason for the opt-in period was that the feature had been verified against **nine titles**, which is not a ~200-title library. That gap is now closed.

## Acceptance (#608 listed four items)

| # | Item | State |
|---|---|---|
| 1a | Both audio tests green | ✅ PASS on presence **and** clipping — Doom, AvP, Iron Soldier |
| 1b | Human listening pass | ✅ maintainer tested; **framerate visibly improved**, no audio problem |
| 2 | Corpus sweep, bit-identical | ✅ #708 — **148 cart images + 6 CD spot-checks**, all byte-identical |
| 3 | FPS on real slow hardware | ⚠️ **not done** — see below |

**On item 3.** No slow target is available to test right now (#601 is still blocked on hardware access). Merging on the strength of the other three plus green CI is a deliberate maintainer decision, not an oversight. The substance of what item 3 was asking — *does this actually make games faster in practice* — was answered by the hands-on test: it does, measurably, on real hardware. What is missing is specifically an FPS number on a **slow** device.

## Changes

- `libretro_core_options.h`: default `disabled` → `enabled`; the description no longer says "off by default while the compatibility corpus grows" and now cites the 148-image sweep.
- `docs/settings-and-performance-guide.md`: the guide said "ships off by default", "you have to turn this on yourself", and listed it as **off** in the quick-reference table.

**One stale claim corrected in passing.** The guide said the feature "is DSP-only, despite the option key" — the GPU port made that wrong, and it contradicted the option's own text ("GPU + DSP"). Flagged because it is not part of the flip; it was already wrong on `develop`.

## A note on how the audio tests were run

`test_audio_clipping` predates the shared harness and has **no `--option`**, so the flag could not be passed as an override. The only honest way to test the flip is to *make* the flip and run against the shipped configuration, which is what this branch does.

## Iron Soldier overclock lockup — filed separately as #718

Found during the hands-on test: Iron Soldier locks up with m68k or GPU overclock. **Not this PR**, on two independent grounds:

- GPU/RISC overclock **suppresses idle-skip outright** (`riscClockScalePct != 100`, `libretro.c:2394`) — it is inert in that arm.
- m68k overclock does **not** suppress it, so that arm is live, and it still changes nothing: `frame_hash_ab`, 1800 frames, `m68k_clock_scale=2x`, idle-skip on vs off → **byte-identical, 1801 rows**.

Control, so the null is not vacuous: stock vs `m68k_clock_scale=2x` **do** differ. Limit: those frames are attract/title, not the gameplay point where the lockup occurs — strong evidence, not proof for the in-game path. #718 records the mid-game savestate that would settle it.

## Verification

`make TEST_EXPORTS=1 test` → exit 0, zero FAILs, with the default flipped. Full CI green.

### DIFF
--- a/docs/settings-and-performance-guide.md
+++ b/docs/settings-and-performance-guide.md
@@ -222,8 +222,9 @@ In order. Stop as soon as it is fast enough.
    step one, not step three.
 2. **Set Internal Resolution to 1x and True Color off** — the manual version of
    step 1, if you want to keep per-title defaults for other games.
-3. **Turn on DSP Idle-Loop Fast-Forward** (`risc_idle_skip`). See §4.3 — this is
-   the biggest lever and it is **off by default**, so nobody gets it by accident.
+3. **Check DSP Idle-Loop Fast-Forward is on** (`risc_idle_skip`). See §4.3 —
+   this is the biggest lever, and it is **on by default** since the corpus
+   sweep. If you turned it off during the opt-in period, turn it back on.
 4. **Switch Blitter to Fast.** Measured 1.21× faster typical (Iron Soldier), up
    to 1.58× on the title that gains most (Skyhammer). It breaks some games —
    wireframe artifacts and missing geometry are the usual symptom — so switch
@@ -246,17 +247,18 @@ unaffected. The saving reported with the feature in the v3.5.0 release notes is
 figure comes from the shipped A/B, not from the profiling documents cited in
 §4.1 — those establish the 50–67% and 90–99.7% numbers, not this one.)
 
-**It ships off by default.** It was verified byte-identical on framebuffer,
-audio and periodic save-state digest across a nine-title corpus including two CD
-games — but nine titles is not a ~200-title library, and a silent audio or video
-divergence nobody can attribute is a worse failure than an opt-in toggle. So it
-waits one release cycle. **If you are chasing performance, you have to turn this
-on yourself.** If a title looks or sounds wrong with it on, turn it off and
-please report it.
+**It now ships on by default.** It shipped off in v3.5.0 against a nine-title
+verification, which was not a ~200-title library — and a silent audio or video
+divergence nobody can attribute is worse than an opt-in toggle. That gap has
+since been closed: the corpus sweep behind #708 ran **148 cart images plus 6 CD
+spot-checks**, off versus on, and every single one was byte-identical across
+framebuffer, audio and save-state hash streams. If a title still looks or sounds
+wrong with it on, turn it off and please report it — that is a finding worth
+having.
 
-**It is DSP-only, despite the option key.** The key is `risc_idle_skip`, but only
-the DSP has the fast-forward path. The GPU spins in idle loops too and this
-option does nothing about that.
+**It covers both RISC processors.** It began DSP-only, which is why the option
+key says `risc_idle_skip` and older text here called it DSP-only; the GPU path
+landed afterwards and the option now drives both.
 
 #### The compounding trap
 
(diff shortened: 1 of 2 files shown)


JoeMatt:
Tested by the maintainer: **the flip works and improves framerate** on real hardware. Human listening pass is therefore discharged (acceptance item 1b).

One finding, and it is **not** this PR: **Iron Soldier locks up with m68k or GPU overclock**, while stock + idle-skip is fine.

Idle-skip is exonerated, on two independent grounds:

- **GPU/RISC overclock suppresses idle-skip outright.** `riscClockScalePct != 100` is one of the four suppression conditions in `perf_warn_idle_skip_suppressed()` (`libretro.c:2394`), so the fast-forward is doing nothing at all in that arm.
- **m68k overclock does NOT suppress it** — so that combination is live — and it still changes nothing. `frame_hash_ab`, 1800 frames, `m68k_clock_scale=2x` with `risc_idle_skip` enabled vs disabled: **byte-identical, 1801 rows**, framebuffer + audio + savestate hash streams.

Control, so the null is not vacuous: stock vs `m68k_clock_scale=2x` (both idle-skip on) **do** differ, confirming the arms are real binaries with real behaviour.

Limit worth stating: those 1800 frames are attract/title, not the gameplay point where the lockup was seen. The byte-identity is strong evidence, not proof for the in-game path; a mid-game savestate would settle it.

This puts Iron Soldier in the existing overclock family — #460 (Defender 2000, m68k 3x), #463 (Cybermorph), #565 / #611 (Club Drive at 2x) — same signature: stock fine, overclock wedges. Worth its own issue rather than blocking this one.

Remaining before merge: **FPS on real slow hardware**, still blocked on #601.

PR #721, opened 2026-08-31
ci(build): guard the MSVC job's source list against Makefile.common drift (#679)

### LINKED ISSUE #679: [ci] Guard the MSVC job's hand-maintained source list against Makefile.common drift
## The problem

`Makefile.common` holds the real source list. Several CI paths keep their **own hand-maintained copies** of it, and they silently fall behind whenever a new source directory appears. The failure mode is always the same: everything is green except the one path with the stale list, and the error is a missing header rather than anything pointing at the real cause.

This bit us **three times in a single day** (2026-08-26):

1. **PR #624** — the SIMD blitter test compiles in `.github/workflows/c-cpp.yml` lacked `-I libretro-common/include`. Four platform jobs failed. The Makefile had always carried the flag; only the by-hand CI compiles didn't.
2. **`scripts/c89-lint.sh`** — no `-Isrc/debug`, so linting any new file under `src/debug/` failed on its own header.
3. **PR #662** — the `msvc-check` job had neither `/Isrc\debug` nor the `src\debug\*.c` files. `libretro.c` had gained `#include "gdbstub.h"`, so `cl.exe` failed on **x64 and x86 only**, while all ~35 other checks passed.

Case 3 is the sharpest illustration: a brand-new source directory built fine everywhere except Windows, and nothing in the repo could have told us before pushing.

## Why the existing guard does not cover this

`scripts/check-package-sources.sh` already solves exactly this problem — for SPM. It diffs `Package.swift`'s `coreSources` against `Makefile.common` and fails on drift. It has caught real bugs (including one in #662).

**Nothing equivalent guards the MSVC job.** Its list lives inline in `.github/workflows/c-cpp.yml`, is maintained by hand, and is only validated by pushing and waiting for Windows.

## Proposal

A `scripts/check-msvc-sources.sh` in the same spirit:

- Extract the `.c` files and `/I` include flags from the `msvc-check` job's **first** `cl.exe` invocation in `.github/workflows/c-cpp.yml` (the second compiles only `blitter.c` + `blitter_simd_sse2.c` for the SSE2 shape and is deliberately narrow — the check must not flag it).
- Compare against `SOURCES_C` / `INCFLAGS` in `Makefile.common`.
- Fail on any source present in `Makefile.common` but absent from the MSVC list, and on any include directory the Makefile has that the MSVC job lacks.
- **Allow a documented exclusion list**, since the MSVC job legitimately omits things (`jlink*.c`, netcode, anything platform-gated). An exclusion must carry a one-line reason inline, so "it's excluded" and "we forgot it" stay distinguishable — that distinction is the whole value of the check.

Wire it into `make lint` alongside the existing checks so it runs locally before a push, not only in CI.

## Worth considering while in there

The same drift risk applies to any other hand-maintained list. Worth grepping for others (the test-target compile rules in `Makefile` are the obvious candidate) and deciding whether one generalised checker beats three specific ones. A single script that takes "which list to check" as an argument may be the better shape.

## Acceptance

- Removing a `src/debug/*.c` entry from the MSVC job's list makes `make lint` fail with a message naming the missing file.
- Adding a new source directory to `Makefile.common` without updating the MSVC job fails locally, before anyone pushes.
- The SSE2-only second `cl.exe` invocation is not flagged.

Related: #624, #662, `scripts/check-package-sources.sh`.

JoeMatt:
Closes #679

## Summary

`scripts/check-package-sources.sh` already guards `Package.swift` against `Makefile.common` drift. **Nothing guarded the `msvc-check` job**, whose source list lives inline in the workflow, is maintained by hand, and is validated only by pushing and waiting for Windows.

That gap cost real time three times in one day (#624; `c89-lint.sh`'s missing `-Isrc/debug`; and **#662**, where a brand-new `src/debug/` directory built everywhere except MSVC x64/x86 while ~35 other checks stayed green).

New `scripts/check-msvc-sources.sh`, wired into **`make lint`** (so drift fails before a push) and into the "duplicated lists in sync" CI job beside its sibling.

## Two lists, deliberately kept apart

The issue is explicit that "excluded on purpose" and "we forgot it" must stay distinguishable. Running the check for the first time found **30 sources the job has never compiled**, despite its step being named "Compile *all* sources" — so folding them into "excluded" would launder exactly what #679 warns about.

- **EXCLUDED** — design, each with a reason. Only the two arch-gated `blitter_simd_*` variants qualify.
- **KNOWN GAP** — debt, tracked, with the rule that it **may shrink, never grow**. A newly added source lands in *neither* list and therefore fails the check — which is the #662 case and the entire point.

**Why not close the 30-file gap here.** It means adding sources to a Windows-only job that cannot be tested from this machine, days before a release. Doing both at once would make a red MSVC job ambiguous: guard first, then the gap, so a failure has one possible cause.

## Acceptance — verified by actually breaking things

| Criterion | Result |
|---|---|
| Removing `src\debug\gdbstub.c` from the job | **FAILS**, naming the file |
| Adding a source to `Makefile.common` only | **FAILS** (exit 1), naming it |
| SSE2-only second `cl.exe` step | **not flagged** |

## A bug this script had, found before it shipped

The include-dir comparison was **vacuous on first write**. `Makefile.common` spells them `-I$(CORE_DIR)/src/...`; the regex matched a bare path, `mk_inc` came out empty, and `comm` against an empty list passes on nothing — precisely the failure class this file exists to prevent. It printed a cheerful `0 include dirs agree`.

Fixed two ways: `mk_inc` is now in the emptiness guard alongside the other lists, and the parse is scoped to the unconditional `INCFLAGS :=` block, because the later `+=` is `msvc2003`-gated and must **not** be demanded of the modern job. **14 include dirs are now really compared.**

## Verification

- `make lint` clean (all three guards).
- `make TEST_EXPORTS=1 test` → exit 0, zero FAILs.

The `yamllint` line-length warning on line 307 is pre-existing — this diff starts at line 484.

### DIFF
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -484,6 +484,13 @@ jobs:
     - name: Package.swift sources match Makefile.common
       run: sh scripts/check-package-sources.sh
 
+    # Same posture for the msvc-check job's inline source list, which rots the
+    # same way and worse: it is Windows-only, so drift shows up as a missing
+    # header on two jobs while ~35 others stay green (issue #679, and PR #662
+    # is the case in point).
+    - name: MSVC job sources match Makefile.common
+      run: sh scripts/check-msvc-sources.sh
+
   spm-build:
     name: SwiftPM build (macOS arm64)
     runs-on: macos-latest
--- a/Makefile
+++ b/Makefile
@@ -2836,6 +2836,11 @@ tools/jagcd/jagcd-chd-check: tools/jagcd/jagcd-chd-check.c $(SOURCES_LIBCHDR)
 
 lint:
 	@scripts/c89-lint.sh
+	@# Hand-maintained source lists that repeat Makefile.common and rot
+	@# invisibly.  Run locally so drift fails BEFORE a push, not after a
+	@# Windows-only CI job fails on a missing header (issue #679).
+	@sh scripts/check-package-sources.sh
+	@sh scripts/check-msvc-sources.sh
 
 # `make coverage` -- builds with gcov instrumentation, runs the full
 # test suite, and produces a Cobertura XML report at coverage.xml plus
--- a/scripts/check-msvc-sources.sh
+++ b/scripts/check-msvc-sources.sh
@@ -1,0 +1,196 @@
+#!/usr/bin/env sh
+# Assert the msvc-check job's hand-maintained source list still matches
+# Makefile.common.
+#
+# Makefile.common is the single source of truth for what makes up the core.
+# The MSVC job in .github/workflows/c-cpp.yml has to repeat it, because
+# cl.exe is driven from a workflow step and cannot read a Makefile.  A second
+# hand-maintained copy of a list that changes every release is exactly the
+# thing that rots, and it rots INVISIBLY here: everything stays green except
+# Windows, and the error is a missing header rather than anything naming the
+# real cause.
+#
+# It has already cost real time three times (issue #679): PR #624 (missing
+# -I libretro-common/include in the by-hand CI compiles), scripts/c89-lint.sh
+# (missing -Isrc/debug), and PR #662, where a brand-new src/debug/ directory
+# built everywhere except MSVC x64 and x86 while ~35 other checks passed.
+#
+# Same posture as scripts/check-package-sources.sh, the version_fallback.h
+# check and the docs/cd-boot-matrix.md drift guard: fail loudly, name the
+# offending files, never edit the workflow.  A human decides what the right
+# list is -- which matters more here than in the SPM case, because the MSVC
+# job legitimately omits things and "excluded on purpose" must stay
+# distinguishable from "forgotten".  That is what EXCLUDED below is for, and
+# why every entry carries a reason.
+#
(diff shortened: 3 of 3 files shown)


JoeMatt:
Kimi's finding was correct and is fixed in `538f4a0`.

The `sed -n '/start/,/end/p'` range whose **end** pattern never matches runs silently to EOF, so renaming only the second step would have folded it into the parse and broken the stated "ONLY the first cl.exe step" invariant — with no error, because the `[ -s ]` guard only catches the *start* pattern failing. The message already said "the step names this parser keys on", plural, while checking one of them. Benign today purely by luck (`blitter_simd_sse2.c` is in `Makefile.common` so it never trips the stale branch, and excluded so it never trips the missing branch) — luck is not an invariant.

Now symmetric: both markers must be present, **and** the extracted range must terminate on the end marker rather than run off the end of the file.

Verified by breaking each one separately rather than by inspection:

| Change | Result |
|---|---|
| rename the END marker only | exit 1, names the missing marker |
| rename the START marker only | exit 1, names the missing marker |
| clean tree | exit 0 |

Shell-only change; `make lint` clean across all three guards, and the three acceptance criteria from #679 re-verified.

Issue #652, opened 2026-08-27
[feat] GDB remote serial protocol stub — debug the 68K, GPU and DSP from a real debugger


JoeMatt:
## Summary

Expose the emulated Jaguar over GDB's **Remote Serial Protocol**, so homebrew developers can debug the 68000, GPU and DSP from `gdb`, `lldb`, VS Code, or any RSP client they already use — breakpoints, watchpoints, single-step, register and memory access.

**Design doc: `docs/gdb-stub-design.md`** (see the PR linked below). This issue tracks the work; the doc carries the reasoning.

## Why this, and why not a scripting VM

This is also the settled answer to "should we have a scripting system". GDB ships a full Python scripting environment on top of breakpoints and memory access — which is the entire feature list BigPEmu's bespoke C-to-bytecode VM sells. Implementing a documented wire protocol gets that capability permanently, with the scripting engine maintained by someone else.

#370 rejected a VM twice on scope grounds; #637 re-confirmed it. A C89 core built for ~20 targets (including msvc05/msvc10 on the GitLab buildbot) has no business maintaining an interpreter, a bytecode format and a sandbox.

Nothing comparable exists for the Jaguar. BigPEmu integrates the Noesis debugger — closed, Windows-centric, one client. An RSP stub works with every debugger, on every platform we build for.

## Why it is smaller than it sounds

Most of the hard parts are already in the tree:

| Need | Already present |
|---|---|
| Per-instruction 68K hook | `M68KInstructionHook()` (`src/core/jaguar.c:364`) — already called every instruction in **shipped** builds for #542's traceback ring; `M68K_HOOK_FUNCTION` is already defined (`m68kinterface.h:101`) |
| 68K halt/resume | `M68KDebugHalt()`/`M68KDebugResume()` (`m68kinterface.c:90`), already wired to six memory-breakpoint sites |
| Memory watchpoints | Exported memory-access breakpoint vars (`jaguar.c:356`), checked at six sites |
| RISC disassembly | `test/tools/gpu_disasm_dump`, `disasm_gpu_isr.py`, `bios_disasm.py` |
| Cross-platform TCP | `src/jerry/jlink_tcp.c`, already built on every target |

The genuinely new code is the RSP protocol layer plus target glue, in a new `src/debug/`.

## Settled decisions

[...]

Issue #651, opened 2026-08-27
[feat] Disk control interface — the missing half of no-content CD boot and the audio-CD path


JoeMatt:
## Summary

The core implements **no** `RETRO_ENVIRONMENT_SET_DISK_CONTROL_INTERFACE` or `..._EXT` — confirmed by grepping the whole of `libretro.c`, nothing. This turns out to be the shared blocker behind two separate features, which is why it is worth filing on its own rather than as a footnote to either.

## Why it matters now

**1. No-content CD boot is structurally unreachable.** #646 / PR #649 added `SET_SUPPORT_NO_GAME`, so the core can now launch with no content and boot the bare console. But with no disk-control interface, a frontend has **no mechanism to hand the core a disc afterwards** — so "start the core, then insert a CD" cannot work, and the CD half of no-content boot was correctly left out of scope there.

**2. It is half of the audio-CD story.** The Virtual Light Machine — the audio-CD visualiser inside the retail CD BIOS — already works (shipped v3.1.0, #291/#300/#325). Combined with no-content boot plus disc insertion, "launch the core, put in a music CD, get the VLM" becomes reachable, which is the BigPEmu behaviour that prompted this. Note the VLM is BIOS code and only runs in **real-BIOS** mode, while our CD default is `hle` — see the separate audio-disc auto-detect work.

## Scope

- Register `SET_DISK_CONTROL_EXT_INTERFACE` where available, falling back to `SET_DISK_CONTROL_INTERFACE`, and report the interface version via `GET_DISK_CONTROL_INTERFACE_VERSION`.
- Implement eject/insert, image index get/set, and the label/path accessors the ext interface adds.
- Decide and document what "eject" means to a running Jaguar CD — the real unit's behaviour on a disc swap mid-session is the accuracy question here, and it should be answered from the JTRM / Jaguar CD documentation (`docs/atari-jaguar-1999/06 - Jaguar CD-ROM.pdf`), **not** from source comments.
- M3U playlist handling, if the frontend expects it for multi-disc sets.

## Open questions to settle first

- **Do any Jaguar CD titles ship on more than one disc?** If none do, multi-disc swapping is not the motivation and the interface exists purely for the insert-after-launch case — which changes how much of the ext interface is worth implementing.
[...]

Issue #637, opened 2026-08-27
[feat] Patching a game: one routing doc for cheats/soft-patches/hooks, plus a user-supplied hook file


JoeMatt:
## The problem

We ship **three** overlapping ways to modify a game, with no single door and no doc that routes between them. The question that prompted this — *"we have the patch stuff but it's not clear how to use it, and I'm not sure users can put their own patch files in yet"* — came from someone who works on this core, which is the strongest possible signal that the discoverability is the bug.

| Mechanism | User-supplied? | Where | State |
|---|---|---|---|
| **Cheats** (#111) | **Yes** — PAR / GameShark hex, entered through RetroArch's cheat UI or a `.cht` file | `src/core/cheat.c`, `retro_cheat_set` | Ships |
| **Soft patching** (#409) | **Yes** — IPS / BPS / UPS placed beside the ROM, applied by the frontend before the core sees the image | `docs/rom-patches.md` | Ships (carts) |
| **Enhancement hooks** (#370) | **No** — static in-core table keyed on CRC32 | `src/core/titledb.c`, `src/core/titlehook.c`, `docs/enhancement-hooks.md` | Mechanism complete, gated off, **zero rows** by design |

So the answer to "can users put their own patch files in?" is **yes, two ways, today** — and nowhere do we say so in one place. The one thing a user genuinely *cannot* do is author a hook.

### 1. One routing doc (`no-code`, the larger half of the value)

A single "how do I modify a game" page that answers, in order:

- **I have a hex code from a cheat site** → cheats. Format, where to paste it, the multi-code `+` separator.
- **I have an IPS/BPS/UPS patch (a translation, a romhack)** → soft patching. Exact filename and directory rules, what the frontend does vs. what the core does, and the cart-only caveat.
- **I want to change bytes in one specific game myself** → enhancement hooks, and the honest statement that this is currently authoring-by-PR.
- **I want to change how the emulator behaves for one game** → per-title defaults (`pairs[]`), not a patch at all. `docs/enhancement-hooks.md`'s "Is my idea a hook?" table already draws this line well; link to it rather than restating it.

Cross-link from the README, the site, and the top of each of the three existing docs, so any entry point reaches the router.

### 2. User-supplied hook file (`size: M`)

Let a user drop a hook definition in the system directory (e.g. `<system>/vj_hooks.json` or a per-title file) and have `TitleHookApplyROM()` consider it alongside the built-in table.

[...]

Issue #635, opened 2026-08-27
[bug] White Men Can't Jump: logo/menu instability under every BIOS and timing combination (68K/TOM IRQ phase)


JoeMatt:
## Report

> "white men can't jump has a hard time booting at all, drawing the logo screens and not crashing
> with pretty much any set of settings matrix, real or hle bios or timing, depending it gets
> further but always has issues"

## Headless boot matrix (build `59e37e9`, 1800 frames, cold boot)

| config | transitions/1800 |
|---|---:|
| HLE BIOS (default) | 351 |
| real BIOS | 372 |
| HLE + all 3 Hardware Timing options | 421 |
| real BIOS + all 3 Hardware Timing options | 372 |
| HLE + fast blitter | 351 |

It does not hard-hang in any of these — the framebuffer keeps changing for the full 1800 frames
everywhere. ~20 % of frames differing on what should be logo/attract screens is consistent with
the reported flickering rather than with a freeze.

Headless transitions cannot tell "flickering badly" from "animating", so this matrix confirms
*liveness*, not correctness.

## This is already root-caused — see `docs/emulation-bug-hunt-todos.md` (High Priority)

Prior private-ROM tracing established:

- The visible jumps happen with a **stable `OLP` and stable object-list contents**.
- The jumping object is a fixed bitmap consumed by OP write-back and restored outside direct 68K
  stores, through the 68K VBI path.
- Bad frames occur when the video interrupt lands **while the 68K SR masks level-2 IRQs**, delaying
  the object restore into active display (`VC=$0866` in the captured repro).
- **TOM timer IRQs are the immediate collision source**: disabling TOM PIT interrupts removes the
  boot-logo jump.

Things already tried that did **not** work, and should not be retried blind:

- Suppressing OP write-back — hides the jump but contradicts hardware behaviour.
- Alternate PIT reload formulas — made the frame sequence worse.
- Line-buffer clearing probes — made the frame sequence worse.

A real fix did land for `INT1` byte reads/writes exposing and clearing pending IRQ sources
consistently with word accesses. WMCJ still needs **68K/TOM interrupt phase accuracy**, not another
bitmap-level workaround.

## Related

NBA Jam T.E. shows the same "occasional screen jumping" and is the same engine (both High Voltage
Software) — likely one shared IRQ-phase fix.

Filed so the report has a tracking issue; the analysis above is the starting point, not a fresh
investigation.

JoeMatt:
[...]

PR #724, opened 2026-09-02
release: v3.6.0 — Debugger & libretro surface

JoeMatt:
Release PR: `release/3.6.0` → `master`, per `docs/release-process.md`.

**91 commits since v3.5.1.** Full notes: `docs/RELEASE_NOTES_v3.6.0.md` (this becomes the GitHub release body).

## Headline

- **GDB remote stub** (#652) — attach `m68k-elf-gdb`/lldb over TCP; GPU and DSP appear as GDB threads.
- **Disk control interface** (#651) — boot with no content, insert a disc after launch, reach the VLM with an audio CD.
- **User-authorable hooks + one patching router doc** (#637).
- **RISC idle-loop fast-forward on by default** (#608), after a 148-image corpus sweep came back byte-identical.
- **White Men Can't Jump fixed** (#635) — froze at its menus with zero audio in both BIOS modes.
- Broad **NEON/SSE2** pass over OP, TOM, blitter, shadowfb and voice chat.

## Version bumps

`CORE_BASE_VERSION` v3.5.1 → v3.6.0 in **both** `Makefile` and `src/core/version_fallback.h` (the deliberate copy for build systems that can't run `gen-version-h.sh`; CI fails if they disagree — it confirmed `CORE_BASE_VERSION = display_version = v3.6.0` on commit).

`dist/info`: `display_version` updated, and **`disk_control` "false" → "true"** since #651 shipped the interface.

## Pre-tag checklist

| Item | Result |
|---|---|
| Clean `make` build | OK |
| `make TEST_EXPORTS=1 test` | **exit 0, zero FAILs** |
| Build identity | `Virtual Jaguar v3.6.0 646d82f` |
| `scripts/c89-lint.sh` | passed (incl. NEON pass) |
| **Audio pair by hand** | presence **and** clipping PASS on Iron Soldier, Doom, AvP, Tempest 2000 |
| Acid gate | `check-baseline.py` → **"OK: no regressions"** |

The audio pair was run **by hand against the private ROMs** — the checklist calls this out specifically because `make test` prints SKIP and still exits 0 without them, so a silencing regression would otherwise pass unnoticed. Iron Soldier 2 is not newly silent (the Songbird rip passes presence; the `.cdi` rip still shows the expected clipping FAIL, i.e. loud-broken not silent). Skyhammer is not in this corpus.

## Noted, deliberately not done

The acid run reports **two timing tests now passing** that `BASELINE.txt` doesn't record as passes (`gpu_store_ext_loop_rate`, `render_bound_loop_rate`) — an improvement, most likely from the idle-skip default. Ratcheting the baseline belongs on `develop` after the release, not inside a release branch.

`docs/WHATSNEW` had **no sections for v3.1.0 through v3.5.1**. Rather than backfill five releases I didn't cut, the new section says so and points at `docs/RELEASE_NOTES_v*.md`, which are complete for those.

`regression_test.sh` could not run on this host (`clang++` is a ccache shim pointing at a stale Swift toolchain — pre-existing, unrelated). CI runs it.

## After merge

Per the process: tag `v3.6.0` and push it (triggers `release.yml` across 16 platforms), then **back-merge `master` → `develop`**, then the `libretro-super` info PR.


JoeMatt:
Both review findings checked. **One fixed, one verified as not applicable to this build** — detail below rather than a blanket "addressed".

## VJ_RESTRICT redefinition — fixed (`a1dc157`)

`src/core/vjag_memory.h` defined `VJ_RESTRICT` unguarded while `tom_scan_simd_sse2.h`, `tom_scan_simd_neon.h` and `shadowfb_simd_neon.h` all define it under `#ifndef`. Now guarded to match.

**Correcting the stated consequence, though:** the review says this "can trigger a macro redefinition warning on some compilers". It cannot today — all four spellings are token-identical (`#if defined(__GNUC__) || defined(__clang__) || defined(_MSC_VER)` → `__restrict`, else empty), and C explicitly permits *identical* macro redefinition. So there is no warning to fix.

The guard is still worth having, for a different reason than the one given: it removes the dependency on four separate copies staying character-for-character in step. That is the real hazard, and it is latent rather than live.

## Vendored CDROM VFS header — verified, not actionable here

The header `libretro-common/include/vfs/vfs_implementation_cdrom.h` is indeed absent, and `libretro-common/vfs/vfs_implementation.c` (which **is** in our build — `Makefile.common:274`, `Package.swift:116`) does include it. But the include is guarded:

```c
#ifdef HAVE_CDROM
#include <vfs/vfs_implementation_cdrom.h>
#endif
```

and `HAVE_CDROM` is **never defined by this build** — not in `Makefile`, not in `Makefile.common`, not in `Package.swift`. Confirmed empirically as well as by inspection: this PR is green across all 16 release platforms plus the SwiftPM and MSVC jobs.

So it is a latent hazard for a *downstream* consumer that defines `HAVE_CDROM`, not a break in ours. Deliberately not fixing it in a release branch: the options are restoring vendored code or rewrapping the include, and both deserve consideration on `develop` rather than being decided during a release cut. Worth its own issue.

## Verification after the change

`make TEST_EXPORTS=1 test` → exit 0, zero FAILs. `scripts/c89-lint.sh` clean. The master conflict (one line: `actions/checkout@v4` vs `@v7` in `pr-issue-link.yml`) is also resolved, and the suite was re-run after that merge too.

JoeMatt:
Both findings were real. One I could resolve in the opposite direction to the guess, since `libretro.c` was outside the reviewed diff. Fixed in `eb4821e`.

## 1. Watchpoints silently never fire — real defect, fixed

Confirmed exactly as described. All six `GDBMemWatchHit()` call sites in `src/core/jaguar.c` sit inside `#ifdef ALPINE_FUNCTIONS` (jaguar.c:537-545 and its five twins), and **no build file in this repo defines it** — checked `Makefile`, `Makefile.common`, `Package.swift` and `.gitlab-ci.yml`. `GDBInsertWatch()` set `bpmActive` and returned success regardless.

So on every shipped target GDB reported the watchpoint as set and then never stopped. Agreed that is the worst failure mode available to a debugger: the user concludes the address is never written.

`GDBInsertWatch()` now returns `-1` when the hit sites are not compiled in, so GDB reports "not supported" rather than lying. Making watchpoints actually *work* means moving those calls out of the ALPINE gate, which puts six branches in the hottest memory path — that wants its own perf measurement, not a release-eve edit. Follow-up, not a release blocker.

Worth noting the release notes never claimed watchpoints; they say "breakpoints, single-step, register and memory access". So nothing shipped is now overclaimed.

## 2. Idle-skip default — the CODE COMMENT was stale, not the notes

You flagged the contradiction correctly and correctly declined to guess. The answer:

`libretro_core_options.h:251` is **`"enabled"`** — #608/#717 flipped the global default. So the release notes are right, and `titledb.c`'s "the global default stays disabled" predates that change (it was written for #708).

The consequence is the interesting part, and I've recorded it rather than quietly leaving it: **the ~30 per-title `risc_idle_skip` rows are now no-ops.** A per-title default only substitutes when the option sits at its registered default, and that value is now the same one the rows carry — exactly the reasoning you used to argue they'd be meaningless under an "enabled" default.

They are kept rather than deleted, for two reasons now stated in the comment: they document *which* titles were actually swept (fires=0 titles still have no row), and they become load-bearing again the moment the global default moves back. The comment explicitly says not to read them as the reason idle-skip is on.

## On the "not flagged" items

- `JGD_AUTO_THRESHOLD` in `none_boot` — leaving as-is; unchanged in this release.
- The duplicated `gdbSession.noAckMode = 0;` — agreed, harmless, not holding a release for it.

## Verification

`make TEST_EXPORTS=1 test` exit 0, zero FAILs. `c89-lint` clean. `test_titledb` and `test_gdbstub_proto` (18 checks) both pass.

JoeMatt:
Two red checks, neither of them this diff:

- **`PR is linked to an issue`** — a release PR has no single issue to link; applied the documented `no-issue` label.
- **`cppcheck static analysis`** — failed at its **`Install cppcheck`** step with exit 100 (package install), not on any finding in the code. Infrastructure, not a result.

Everything else is green: **67 passed**, 1 skipped, including all 16 release build targets, MSVC x64/x86, the SwiftPM build, regression tests and the acid suite.

Merging, then tagging v3.6.0.

PR #725, opened 2026-09-03
feat(debug): opt-in LAN binding for the GDB stub, with a peer check (#652)

JoeMatt:
Refs #652

## Why

The stub bound `INADDR_LOOPBACK` unconditionally. Right as a default, but it makes one real case impossible: **debugging a game running on a phone, tablet or TV.** SSH port-forwarding covers desktop-to-desktop; a stock iOS/tvOS device has no shell to forward *from*, so there was no path at all.

## What

New core option **`virtualjaguar_gdb_bind`** — `loopback` (default) / `lan`. The plumbing is four lines; the security posture is the actual design:

- **Fails closed.** Absent, unset or unrecognised values resolve to loopback inside `GDBSockSetBindMode()`, *and* the module's initial value is loopback — so a caller that never sets it (a test, or a frontend that has never heard of the option) still gets the safe bind.
- **Read raw via `environ_cb`**, never `get_variable_pertitle()` — same rule as the enhancement-hooks gate. A per-title DB row must not be able to open a debug socket on the user's behalf.
- **Latched once at content load**, so changing the setting mid-session can't widen an already-open socket.
- **Peer check at `accept()`.** In `lan` mode only RFC1918, CGNAT (100.64/10), link-local and loopback peers are accepted; anything routable is closed and logged with its address.
- **Loud.** `LOG_WRN` (not INF) naming the bound address and port, plus the on-screen banner, whenever `lan` is live.

## What the peer check does and doesn't do

It stops the **silent accidental** case — a public address via carrier NAT, a misconfigured hotspot — which is the failure a user cannot notice. It does **not** make a hostile LAN safe, and it will refuse a legitimate VPN peer on a public range (visibly, in the log). That trade is stated rather than implied.

**The GDB remote protocol has no authentication of any kind**, and nothing here adds any. While the stub is open on the LAN, anyone who can reach the port can read/write emulated memory and control execution. That's in the option text, the user guide, and the header comment.

## Two things corrected in passing

- The banner previously printed a hardcoded `127.0.0.1`. It now reports the **actual** bound address — otherwise it would have become a lie in `lan` mode.
- The `gdb_stub` option's own description asserted the server *"only ever listens on 127.0.0.1 — it is not reachable from another machine."* No longer unconditionally true; reworded.

## Tests

Four fail-closed assertions in `test_gdbstub_proto` (**22 total, was 18**): default is loopback with no setter call, explicit LAN is honoured, an unrecognised value falls back to loopback rather than LAN, and LAN can be turned back off. Those are what stop the default drifting in a later refactor.

Linking `gdbsock.c` into that test needed `-DINLINE=inline` (for `log.h`) and the `vj_log_cb` stub that `test_blitter_mmio.c` and `test_titlehook.c` already use.

## Not verified

Whether **iOS/tvOS** permits RetroArch to accept inbound LAN connections without a local-network entitlement prompt. Listening is generally allowed, but I have not tested it on a device and this PR does not claim it works there — that's the next thing to try, and the reason the option exists.

## Verification

`make TEST_EXPORTS=1 test` → exit 0, zero FAILs. `scripts/c89-lint.sh` clean on every touched `.c`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2155,9 +2155,14 @@ test/test_titlehook: test/tools/test_titlehook.c src/core/titlehook.c \
 # GDB Remote Serial Protocol engine (issue #652, Phase 1).  Pure protocol
 # unit tests: no emulator, no sockets. src/debug/gdbstub.c never calls
 # socket() and never dereferences a Jaguar global, so it links alone here.
-test/tools/test_gdbstub_proto: test/tools/test_gdbstub_proto.c src/debug/gdbstub.c
-	$(CC) -O2 -Wall -std=c99 $(INCFLAGS) \
-		-o $@ test/tools/test_gdbstub_proto.c src/debug/gdbstub.c
+test/tools/test_gdbstub_proto: test/tools/test_gdbstub_proto.c src/debug/gdbstub.c \
+		src/debug/gdbsock.c
+	@# -DINLINE: gdbsock.c pulls in src/core/log.h for the refused-peer
+	@# warning, and log.h uses INLINE, which the core build supplies via
+	@# -DINLINE="inline". Test rules get no such default.
+	$(CC) -O2 -Wall -std=c99 -DINLINE=inline $(INCFLAGS) \
+		-o $@ test/tools/test_gdbstub_proto.c src/debug/gdbstub.c \
+		src/debug/gdbsock.c
 
 test/test_event_queue: test/test_event_queue.c src/core/event.c src/core/event.h
 	$(CC) -O2 -Wall -std=c99 $(INCFLAGS) \
--- a/Makefile
+++ b/Makefile
@@ -2152,9 +2152,15 @@ test/test_titlehook: test/tools/test_titlehook.c src/core/titlehook.c \
 		-o $@ test/tools/test_titlehook.c src/core/titlehook.c \
 		src/core/titledb.c src/core/crc32.c src/core/hookfile.c
 
-# GDB Remote Serial Protocol engine (issue #652, Phase 1).  Pure protocol
-# unit tests: no emulator, no sockets. src/debug/gdbstub.c never calls
-# socket() and never dereferences a Jaguar global, so it links alone here.
+# GDB Remote Serial Protocol engine (issue #652).  No emulator:
+# src/debug/gdbstub.c never dereferences a Jaguar global.
+#
+# gdbsock.c is linked too, for the bind-mode fail-closed tests (#652 LAN
+# option).  It DOES contain socket code, but none of it runs here -- the
+# tests only exercise GDBSockSetBindMode()/GDBSockGetBindMode(), which
+# touch a single static and never open a socket.  Linking the real file
+# rather than a stub is the point: it is the shipped default that must
+# be proven safe, not a copy of it.
 test/tools/test_gdbstub_proto: test/tools/test_gdbstub_proto.c src/debug/gdbstub.c \
 		src/debug/gdbsock.c
 	@# -DINLINE: gdbsock.c pulls in src/core/log.h for the refused-peer
--- a/docs/gdb-stub-guide.md
+++ b/docs/gdb-stub-guide.md
@@ -197,8 +197,23 @@ implementation was actually verified against; please file an issue with what you
 
 ## Security notes (short version — see the design doc for the full reasoning)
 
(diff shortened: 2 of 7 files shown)


JoeMatt:
Retargeted to `release/3.6.0` (clean — the branch is exactly one commit ahead of it, so nothing from `develop` rides along). Both findings addressed below; one is a non-issue and I'd rather show the evidence than quietly "fix" it.

## 1. `socklen_t` on MSVC — verified, no change needed

You correctly flagged that you couldn't see the include block. It is:

```c
#ifdef _WIN32
#include <winsock2.h>
#include <ws2tcpip.h>      /* gdbsock.c:23-24 */
```

That is **byte-identical to `src/jerry/jlink_tcp.c:32-33`**, which uses bare `socklen_t` at line 533 — as does `jlink_discover.c:359`. Both files are in `SOURCES_C` and compile on the msvc05 i686 / msvc10 i686 / msvc10 x64 buildbot targets today. So `socklen_t` is already reaching those compilers through `ws2tcpip.h`, and this file pulls in exactly the same header.

I deliberately did **not** add the defensive `typedef int socklen_t` — on an SDK that *does* define it (which the evidence above says these are), a bare typedef is a redefinition, so the "cheap insurance" would be the thing that breaks the Windows build rather than saving it.

Worth noting the buildbot has already had one genuine MSVC catch on this release (`gdbtarget.o : unresolved external symbol snprintf`, fixed in `7f0e12f`), so the house rule is being applied — this one just isn't a hit.

## 2. Header summary missing CGNAT — real, fixed in `19bc253`

Correct. `gdb_peer_allowed()` accepts 100.64/10 and the user guide says so, but `gdbstub.h`'s summary listed only RFC1918 / link-local / loopback. Since the header is what a future reader trusts, it now names all six ranges and points at `gdb_peer_allowed()` as the single authority rather than restating the list a third place that can drift.

## Verification after the change

`test_gdbstub_proto` 22 passed / 0 failed; `c89-lint` clean.

Thanks for checking the mask arithmetic range by range and confirming `ntohl` ordering — that and the fail-closed/latch semantics were the parts I most wanted a second pair of eyes on.

JoeMatt:
All three inline findings were real and are fixed in `b1c3a27`.

**1. `libretro.c` — LAN banner dropped the freeze warning.** Correct, and the reasoning is the part worth keeping: binding scope changes *who can reach* the stub, not what a halt does to the frontend, so that warning applies in both modes. My ternary had made the banner say **less** exactly when more is at stake. Both warnings now appear in LAN mode; buffer grew 96 → 160 to fit them.

**2. `Makefile` — stale comment.** Right: "no sockets ... links alone here" stopped being true the moment the rule gained `src/debug/gdbsock.c`. It now states that gdbsock.c is linked, that none of its socket code actually runs (the tests only touch `GDBSockSetBindMode`/`GetBindMode`, which move a single static), and why the real file is linked rather than a stub — the shipped default is what has to be proven safe, not a copy of it.

**3. `gdbsock.c` — log tag and option name.** Fixed: `[gdb]` → `[GDB]` to match every other stub log, and `gdb_bind=lan` → `virtualjaguar_gdb_bind=lan`, the actual key a user would search for or set. Both were exactly as described — harder to grep, easier to misread.

Verification: `make TEST_EXPORTS=1 test` exit 0 / zero FAILs, `test_gdbstub_proto` 22 passed, `c89-lint` clean.

Note this PR now targets `release/3.6.0` rather than `develop`, so the option ships in v3.6.0 rather than a later release.

JoeMatt:
Applied the `no-issue` label, because the link check is failing for a reason its own error message hides.

The job log says:

```
gh: Issue exceeds manual reference limit
##[error]PR #725 has no linked issue and no link tag in its body.
```

The body **does** carry both `Refs #652` and `<!-- link-issue: #652 -->`. What actually happened is that GitHub refused `addCloseIssueReferences` because **#652 has hit its cap on manually-linked PRs** — the GDB stub shipped across many PRs. #652 is also already closed, so the link would add nothing.

So this is the documented opt-out being used for a real opt-out case, not a way around a missing tag. Worth noting the check reports a limit error as "no link tag", which will mislead the next person to hit it — that is worth a small fix to `pr-issue-link.yml` (distinguish the API refusal from an absent tag), but not on a release branch.

Issue #646, opened 2026-08-27
[feat] Support RETRO_ENVIRONMENT_SET_SUPPORT_NO_GAME (boot bare console with no content)


JoeMatt:
Today the core requires content: \`retro_load_game(NULL)\` is not supported at all
(\`if (!info) return false;\`).  BigPEmu supports launching the Jaguar core with no
content, booting straight into the console boot ROM (cube logo), which is exactly what
real hardware does with an empty cart slot.

## Ask

Implement \`RETRO_ENVIRONMENT_SET_SUPPORT_NO_GAME\` so the frontend can launch the core
with no content:

- Boot the real console boot ROM with no cartridge inserted (Series K / Model M per the
  existing \`virtualjaguar_bios_type\` option) -- shows the boot animation and then sits,
  matching real hardware.
- Do not crash or dereference \`info\`/\`info->data\`/\`info->path\` anywhere in the
  no-content path (titledb CRC lookup, cart BIOS autodetect, CD detection, memory maps,
  save RAM).
- CD boot (HLE or real CD BIOS) is **out of scope** for this ticket: there is no
  \`RETRO_ENVIRONMENT_SET_DISK_CONTROL_INTERFACE\`/EXT implementation in this core, so a
  frontend has no way to hand the core a disc image after a no-game launch. That is a
  separate, larger feature.

## Non-goals

- Disk-control interface / swapping media into a no-game session.
- Any change to normal cart/CD boot behavior.

PR #731, opened 2026-09-05
fix(boot): no-content boot comes up in the CD BIOS, not an empty cart slot

### LINKED ISSUE #726: [bug] v3.6.0: booting with no content shows the red logo and dies — cannot reach the CD BIOS
Reported by the maintainer hours after v3.6.0 published. **This is the headline feature of the release and it does not work.**

## Symptom

Booting the core with **no content** ("gameless" / boot-core-raw) shows the **red Jaguar logo and dies**, in **both** HLE and real-BIOS modes. From there you cannot reach the CD BIOS to insert a disc.

That combination — #646 (no-content boot) plus #651 (disk control) — was supposed to give: boot bare console → insert an audio CD → Virtual Light Machine. The insert half works (demonstrated live, including over the network); the **no-content entry point** does not.

## Why CI and the unit tests missed it

`test/tools/test_disk_control.c` case 1 asserts only that `bootConfig.strategy` moves off `"none"` onto a CD strategy. **It never renders or inspects a single frame.** A machine that resolves the boot path correctly and then freezes on the logo passes all four assertions. It still does today:

```
./test/tools/test_disk_control ./virtualjaguar_libretro.dylib --disc <myst.cdi> --case 1
```

So any fix must add a **frame-progression** assertion, or this class ships again. (Same shape as #712, where a titledb assertion encoded a fact about the table rather than the behaviour under test.)

## Already established — please don't redo

- **No headless crash.** 600 frames of no-content boot followed by an insert exits 0. Whatever kills it is not visible to the dlopen harness — suspect the RetroArch-side path (AV info / geometry / video presentation with no content).
- **Not an unallocated buffer.** `videoBuffer` is allocated at `libretro.c:6173`, *before* the content branch at ~6249. Ruled out.
- **`none_boot()`** (`src/cd/jagcd_cart.c:102`) zeroes the cart window (`memset(jaguarMainROM, 0, JGD_AUTO_THRESHOLD)`), clears `jaguarROMSize`/`jaguarMainROMCRC32`, sets `jaguarCartInserted = true` to force the boot-ROM vector path, then `JaguarReset()`. **The red logo is the boot ROM running against an empty cart window** — on real hardware that screen sits forever, so "crashed" and "frozen" look identical from outside.

## Next steps, in order

1. **Decide frozen vs crashed** in RetroArch — the log at the moment of failure with `virtualjaguar_crash_detect` enabled. Note crash_detect is blind to anything that never exits `retro_run`, so a hang inside a frame shows nothing.
2. **Verify `JGD_AUTO_THRESHOLD`** against the real flat cart-window allocation in `jaggd.h`. Raised in the #724 review and still unverified: if the memset is short, stale bytes survive where the boot ROM's cart-header check can read them as a cartridge; if long, it overruns. Cheapest concrete lead.
3. **Reproduce headlessly with video** — no-content boot, N frames, assert the framebuffer actually changes. That assertion is simultaneously the missing test and the repro.
4. Only then check that insert-after-launch reaches a *booting game*, which is a different claim from "the strategy re-resolved".

## Scope

v3.6.0 is published and a 3.6.1 has been announced publicly as the fix, so this is committed to.

JoeMatt:
Closes #726

## The report, and what it actually was

#726 said v3.6.0's no-content boot shows the red logo and dies, with no way to reach the CD BIOS.

**First: that report was made against a stale install.** The core in the reporter's RetroArch is `v3.5.1 e6709d2`, and its info file is v3.1.0-era. No-content boot landed in bc5b849, which `git tag --contains` puts in **v3.6.0 only** — so the installed core never had the feature. Confirmed by string-scanning the installed dylib: it has three boot paths (cart, HLE, real BIOS) and no bare-console path at all.

**But the feature was still broken**, for a different reason, and that is what this fixes.

## The real defect

No-content boot resolved to `cd_boot_strategy_none`: the boot ROM running against a zeroed cart window. That is authentic for a Jaguar with no CD unit attached — and a dead end. The CD BIOS is what offers "insert a disc", and it is mapped as a cartridge at `$800000`, so **nothing reachable from the empty-slot screen can ever get to it.** "Launch bare, insert a disc, reach the VLM" — the point of #646 + #651 and the headline of v3.6.0 — was unreachable by construction.

This resolves to the real CD BIOS instead, mirroring `ResolveBootConfig()`'s `CDBOOT_BIOS` arm by hand (it still isn't usable here — it only knows cart/CD content). Neither call can fail for want of a file: `stage_cd_bios()` prefers an external ROM and otherwise copies an embedded image, and `bios_boot()` needs no disc.

Second commit fixes the same path leaving `jaguar_cd_mode` and `jaguarMemTrackInserted` at their cartridge defaults — see the Copilot thread below.

## Measured, not assumed

Two separate runs, both against this branch's core. Stating the provenance of each, since the frame numbers differ:

**1. `test_disk_control` case 5** — 600 headless frames of no-content boot, reading the real framebuffer:

| | frames | non-black | distinct |
|---|---|---|---|
| before | 600 | 564 | **35** (static logo) |
| after | 600 | 549 | **528** (BIOS animating) |

**2. A throwaway capture probe** (not case 5, and not in the diff) using the same `harness_load_no_content` entry point, purely to see what those 528 frames actually *are* — a change-counter alone cannot tell an animating BIOS from churn:

- frame 400 of a 600-frame run → the CD BIOS cube intro
- frame 2200 of a **2400**-frame run → its insert-disc attract screen, which is the screen a disc is inserted from

The 2400-frame run is the only source of the frame-2200 capture; case 5 itself remains 600 frames. Thanks to @kimi for catching that the original wording implied otherwise.

## Also: `supports_no_game` was never flipped

`dist/info` still said `supports_no_game = "false"` after `SET_SUPPORT_NO_GAME` shipped. RetroArch reads the **info database**, not the runtime env call, to decide whether to offer a contentless launch — so the feature wasn't advertised to the frontend at all. One line.

## Two tests asserted the old strategy

Both encoded the bug, which is how it shipped green:

- `no_game_boot_selects_none_strategy` asserted `IS_NONE`. Renamed to `no_game_boot_selects_cd_bios_strategy`; now pins the BIOS plus `cdBiosAvailable`.
- `test_disk_control` case 1 compared strategy names either side of an insert. No-content is now `"bios"` and a real-BIOS disc also resolves to `"bios"`, so that comparison **would have gone vacuous**. It now pairs the name with `isCDGame`, which is what actually moves: false for the bare BIOS on its insert screen, true once a disc is mounted.

## Verification

- Full suite: exit 0, zero failures
- C89 lint clean; build-identity guard OK
- Audio pair by hand: presence RMS 1174.2 (expected ~1175), clipping 0 saturated samples
- Acid gate: `Regressions: 0` (147/150, the 3 FAILs are baselined)
- `test_disk_control` cases 1/3/5 green with a real disc; case 1's changed assertion is load-bearing (it failed before the `libretro.c` change)

The `.gitignore` hunk is unrelated to the boot fix and kept deliberately: the `dsp_halt_backtrace` binary is produced by a tool whose source is already tracked, and without the ignore it gets staged by `git add -A`. One line, and it stops a 100 KB binary landing in the tree.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -54,6 +54,7 @@ tools/jagcd/jagcd-chd-check
 tools/jagcd/chdman
 tools/jagcd/*.exe
 test/tools/blit_memo_verify
+test/tools/dsp_halt_backtrace
 test/tools/dsp_idle_ab
 test/tools/dsp_idle_probe_falsify
 test/tools/gpu_idle_ab
--- a/dist/info/virtualjaguar_libretro.info
+++ b/dist/info/virtualjaguar_libretro.info
@@ -14,7 +14,7 @@ systemname = "Jaguar"
 systemid = "atari_jaguar"
 
 # Libretro Features
-supports_no_game = "false"
+supports_no_game = "true"
 database = "Atari - Jaguar"
 savestate = "true"
 savestate_features = "3"
--- a/docs/settings-and-performance-guide.md
+++ b/docs/settings-and-performance-guide.md
@@ -449,6 +449,11 @@ disk-control interface, so a frontend can start it with no content at all and
 hand it a disc afterwards — which is what makes "launch the core, put in a
 music CD, get the Virtual Light Machine" reachable.
 
+Started with no content, the core comes up **in the CD BIOS**, on its
+insert-disc screen, which is the screen you insert *from*. No files are
+required: an external CD BIOS ROM in the system directory is preferred if you
+have one, and otherwise an embedded image is used.
+
 Two things to know about how that behaves:
 
 - **Inserting a disc restarts the console.** The boot strategy is derived from
--- a/libretro.c
+++ b/libretro.c
@@ -6267,6 +6267,21 @@ bool retro_load_game(const struct retro_game_info *info)
        * zero files present. */
       stage_cd_bios();
 
+      /* The CD unit is attached -- that is WHY the BIOS is running -- so
+       * this is CD mode with nothing in the tray, not cartridge mode.
+       * Both flags mirror open_disc_and_resolve_boot(); only the disc is
+       * missing.  Without them a machine sitting in the CD BIOS gets the
+       * options menu inverted against it (update_option_visibility() keys
+       * both gates off jaguar_cd_mode, so CD Boot Mode / CD BIOS Type
+       * would be HIDDEN and the cartridge-BIOS option shown), and the
+       * Memory Track cart -- which plugs into the CD unit and is present
+       * whether or not a disc is -- would stay absent until the first
+       * insert.  Safe with no disc: the three disc-identity accessors
+       * return zero when nothing is mounted (see the savestate chunk),
+       * and the MT save-buffer paths only touch mtMem. */
+      jaguar_cd_mode             = true;
+      jaguarMemTrackInserted     = opt_memory_track;
+
       vjs.useJaguarBIOS          = true;
       bootConfig.isCDGame        = false;
       bootConfig.showBootROM     = true;
--- a/libretro.c
+++ b/libretro.c
@@ -6242,15 +6242,36 @@ bool retro_load_game(const struct retro_game_info *info)
    }
    else if (!info)
    {
-      /* No-content boot (issue #646): ResolveBootConfig() only knows about
-       * cart/CD content, and there is no 68K program anywhere for HLE to
-       * jump into, so the real boot ROM is the only sane strategy --
-       * exactly what real hardware shows with an empty cart slot. */
+      /* No-content boot (issue #646, corrected for #726): come up in the
+       * CD BIOS, not on an empty cart slot.
+       *
+       * cd_boot_strategy_none models a bare console -- the boot ROM running
+       * against a zeroed cart window.  That is authentic for a Jaguar with
+       * no CD unit attached, but it is a dead end for the feature this path
+       * exists to serve: the CD BIOS is what offers "insert a disc", and it
+       * is mapped as a cartridge at $800000, so nothing reachable from the
+       * empty-slot screen can ever get to it.  Booting bare in order to
+       * insert a disc afterwards -- the whole point of #646 plus disk
+       * control (#651) -- was therefore unreachable, which is what #726
+       * reported as "red logo, and you cannot boot into CD".
+       *
+       * ResolveBootConfig() is still not usable here (it only knows about
+       * cart/CD content), so this mirrors its CDBOOT_BIOS arm by hand:
+       * showBootROM is forced on because the real CD BIOS path requires it.
+       *
+       * Neither call below can fail for want of a file.  stage_cd_bios()
+       * prefers an external ROM from the system directory and otherwise
+       * copies an embedded image, and bios_boot() needs no disc -- it
+       * memcpys the staged BIOS, takes the run address from $800404 and
+       * resets -- so the BIOS comes up on its own insert-disc screen with
+       * zero files present. */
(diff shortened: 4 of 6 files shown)


JoeMatt:
@copilot-pull-request-reviewer's finding was correct, and checking it turned up a second consequence of the same root cause. Fixed in 95c8899.

**The finding as stated.** The new path staged and booted the CD BIOS but left `jaguarMemTrackInserted` false. The Memory Track cart plugs into the CD unit and is present whether or not a disc is, so it should not have been waiting on the first insert.

**What that led to.** The real root cause was broader: the path left *both* CD-unit flags at their cartridge defaults, `jaguar_cd_mode` included. That one is worse, because `update_option_visibility()` keys both of its gates off it:

```c
show_cd_options       = (!content_loaded || jaguar_cd_mode);
show_cart_bios_option = (!content_loaded || !jaguar_cd_mode);
```

`content_loaded` is set unconditionally at the end of `retro_load_game` — no-content path included, verified — so it is true when this evaluates. A machine sitting in the CD BIOS therefore had **CD Boot Mode / CD BIOS Type hidden and the cartridge-BIOS option shown**: exactly inverted, on the one screen where the CD options are the ones you want.

Both flags now mirror `open_disc_and_resolve_boot()`; only the disc itself is missing.

**Why it is safe with an empty tray**, since that is the obvious next question: the three disc-identity accessors return zero when nothing is mounted (documented at the savestate chunk, libretro.c:4632), so a state taken on the bare BIOS records 0/0/0 and is still correctly refused against a mounted disc. The Memory Track save-buffer paths only ever touch `mtMem`.

Full suite green (exit 0, zero failures), C89 lint clean.

JoeMatt:
Both Kimi findings are fair. Addressed in the description rather than the code, since neither touches behaviour.

**1. Frame-count discrepancy — you're right, and the provenance was genuinely missing.** The captures did not come from case 5. Case 5 is 600 frames and stays 600 frames; the images came from a separate throwaway probe (not in the diff) that uses the same `harness_load_no_content` entry point, run twice — frame 400 of a 600-frame run for the cube intro, frame 2200 of a **2400**-frame run for the attract screen. I wrote both numbers under one heading as though they shared a run. The description now states which run each capture is from.

Worth saying why the second run exists at all, because it is the same point you're making: 35 → 528 distinct frames is a *change counter*, and a change counter cannot distinguish an animating BIOS from garbage. The captures are there to confirm the 528 are real frames. So the load-bearing claim needed exactly the provenance it lacked.

**2. `.gitignore` hunk — unrelated, agreed.** Keeping it, with the reasoning recorded: `test/tools/dsp_halt_backtrace` is the build product of a tool whose source is already tracked, and with no ignore rule a `git add -A` stages a 100 KB binary. Splitting a one-line ignore into its own PR costs more than the blame noise it saves. Flagging it as a deliberate call rather than an oversight.

On the two you checked and didn't flag, both caveats are correctly stated:

- `cdBiosAvailable = true` does depend on the embedded fallback being real. It is today — `stage_cd_bios()` falls through to `jaguarCDBootROM` / `jaguarDevCDBootROM` unconditionally — and you're right that making the embedded image optional at build time would turn both this assignment and the test into a lie. Not hypothetical enough to guard now, but a real coupling.
- The disc-identity accessors zeroing on an empty tray is stated at libretro.c:4632 ("zero when no disc is mounted") and is what makes the savestate identity chunk safe here: a state taken on the bare BIOS records 0/0/0 and is still correctly refused against a mounted disc.

PR #732, opened 2026-09-05
release: v3.6.1

### LINKED ISSUE #726: [bug] v3.6.0: booting with no content shows the red logo and dies — cannot reach the CD BIOS

JoeMatt:
Refs #726

Patch release for one thing: **booting the core with no content now lands you in the CD BIOS**, which is the screen you insert a disc from.

v3.6.0 shipped no-content boot (#646) and disk control (#651) so you could launch bare, insert a disc, and reach the VLM with an audio CD. That combination did not work — no-content boot resolved to a bare console (boot ROM against a zeroed cart window), and the CD BIOS is mapped *as a cartridge* at `$800000`, so nothing reachable from the empty-slot screen could get to it. The headline feature was unreachable by construction.

## Contents

- **#731** — no-content boot comes up in the CD BIOS; same path now reports CD mode with an empty tray rather than cartridge mode (the options menu was inverted against it, and Memory Track was absent until the first insert); `supports_no_game` flipped in `dist/info`.
- **#727 / #728** — already on `develop`: the frame-progression test that redirected #726, and the acid baseline ratchet.
- **#729** — the `pr-issue-link` error-message fix, back-merged from `master`.

## Version bump

| | |
|---|---|
| `CORE_BASE_VERSION` | `v3.6.1` |
| `display_version` | `v3.6.1` |
| `check-info-version.sh` | OK |

## Pre-tag gates

Per `docs/release-process.md`, all run against this branch:

- `make test` — exit 0, zero failures
- C89 lint clean
- **Audio pair by hand** (the gate `make test` silently skips without the private ROMs): presence RMS **1174.2** against an expected ~1175; clipping **0** saturated samples
- **Acid baseline** — `Regressions: 0` (147/150; the 3 FAILs are the baselined ones)

## Note for anyone tracking #726

The original report was made against a **v3.5.1** core — a version predating the feature entirely. The stale install is worth ruling out separately from this fix; the release notes say so.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -54,6 +54,7 @@ tools/jagcd/jagcd-chd-check
 tools/jagcd/chdman
 tools/jagcd/*.exe
 test/tools/blit_memo_verify
+test/tools/dsp_halt_backtrace
 test/tools/dsp_idle_ab
 test/tools/dsp_idle_probe_falsify
 test/tools/gpu_idle_ab
--- a/Makefile
+++ b/Makefile
@@ -2061,6 +2061,15 @@ test: test/test_dram_timing test/test_cheat test/test_event_queue test/test_jlin
 	@# REM SESSION markers and CDI headers declare numSessions=2 -- so it
 	@# would pass against a data disc for the wrong reason.  Ledgered as a
 	@# skip rather than written to look like coverage.
+	@# Case 5 (no-content boot renders AND progresses, #726) takes NO disc
+	@# and must NOT sit inside the corpus guard below.  It is the
+	@# regression cover for the bug that shipped in v3.6.0, so it has to
+	@# run on public CI and on contributor machines, which are exactly the
+	@# checkouts with no test/roms/private.  An earlier revision nested it
+	@# in the `if [ -n "$$disc" ]` block, where it silently ran nowhere but
+	@# a corpus machine.
+	./test/tools/test_disk_control ./$(TARGET) --case 5 --quiet
+
 	@bash scripts/test-skip.sh record "Disk control audio-disc insert (#651)" \
 		"no one-session (Red Book) disc in the private corpus"
 	@disc=$$(find -L test/roms/private -iname '*.cdi' -o -iname '*.cue' 2>/dev/null | head -1); \
@@ -2068,7 +2077,6 @@ test: test/test_dram_timing test/test_cheat test/test_event_queue test/test_jlin
 		rc=0; \
 		./test/tools/test_disk_control ./$(TARGET) --disc "$$disc" --case 1 --quiet || rc=1; \
 		./test/tools/test_disk_control ./$(TARGET) --disc "$$disc" --case 3 --quiet || rc=1; \
-		./test/tools/test_disk_control ./$(TARGET) --case 5 --quiet || rc=1; \
 		discb=$$(find -L test/roms/private -iname '*.cdi' -o -iname '*.cue' 2>/dev/null | sed -n 2p); \
 		if [ -n "$$discb" ]; then \
 			if ./test/tools/test_disk_control ./$(TARGET) --disc "$$disc" --disc-b "$$discb" --case 4 --quiet; then :; \
--- a/Makefile
+++ b/Makefile
@@ -183,7 +183,7 @@ TARGET_NAME := virtualjaguar
 # Single source-of-truth for the human-readable version string.
 # Bumped by .github/workflows/version-bump.yml (greps this line).
 # Composed into CORE_VERSION in src/core/version.h, generated below.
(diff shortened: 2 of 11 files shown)


JoeMatt:
Both @copilot-pull-request-reviewer findings are real. Filed as #733 rather than fixed here, deliberately — release branches are bug-fix-only until tagged, and neither of these is a bug in the release.

**Unaligned `uint32_t` reads.** Correct in principle: `vid_cb()`'s guard checks that `pitch` is long enough (`pitch < w * 4`) but not that it is a multiple of 4, so a non-aligned pitch would produce UB rather than the loud failure the guard's own comment promises.

Not holding the release for it, for reasons that are worth stating rather than assumed:

- It is **test-only** code — `test/tools/test_disk_control.c`, not the shipped core. No user-facing binary is affected.
- For the core's XRGB8888 output `pitch` is always `w * 4`, and case 5 passes today on linux-x64, linux-arm64, linux-i686 and macos-arm64. Nothing in the build matrix trips it.
- It is **unchanged from #727**, where it was introduced and already shipped to `develop`. It is not a regression in v3.6.1, so tagging does not make anything worse.

The reason it is still worth fixing is the one the finding implies: the guard exists specifically to catch a pixel-format change (the comment names RGB565), and a non-multiple-of-4 pitch is one of the ways such a change would arrive. As written the guard would wave that through into UB instead of failing loudly. That is a latent hole in an intentional safety check, which is exactly the kind of thing that deserves its own issue rather than a rushed commit on a release branch.

**`--disc` usage text** — same call, same issue.

Everything else is green: 69/70 checks pass, the one non-success is `release: SKIPPED`, which only runs on tag push. The two earlier failures are resolved — `version_fallback.h` carried a third copy of the version string that my local `check-info-version.sh` does not cover (fixed in a170814), and the Windows `jagcd tools` job was an `Add-MpPreference` Defender-service flake that passed on re-run.

Merging and tagging.

